PR #141, opened 2022-09-18
feat: One scan per folder [ROAD-1126]

asaf92:
### Description

Product scanners avoid running several scans at once for the same paths. For OSS & IAC the previous scans are cancelled and restarted, and for Code the new scan will wait for the previous scan to finish

### DIFF
--- a/infrastructure/code/code.go
+++ b/infrastructure/code/code.go
@@ -34,7 +34,7 @@ type ScanMetrics struct {
 
 type ScanStatus struct {
 	// finished channel is closed once the scan has finished
-	finished chan struct{}
+	finished chan bool
 
 	// isRunning is true when the scan is either running or waiting to run, and changed to false when it's done
 	isRunning bool
@@ -45,7 +45,7 @@ type ScanStatus struct {
 
 func NewScanStatus() *ScanStatus {
 	return &ScanStatus{
-		finished:  make(chan struct{}),
+		finished:  make(chan bool),
 		isRunning: false,
 		isPending: false,
 	}
--- a/infrastructure/code/code.go
+++ b/infrastructure/code/code.go
@@ -96,6 +96,7 @@ func (sc *Scanner) Scan(ctx context.Context, _ string, folderPath string) []snyk
 	previousScanStatus, wasFound := sc.runningScans[folderPath]
 	if wasFound && previousScanStatus.isRunning {
 		if previousScanStatus.isPending {
+			sc.scanStatusMutex.Unlock()
 			return []snyk.Issue{}
 		}
 
--- a/infrastructure/code/code.go
+++ b/infrastructure/code/code.go
@@ -32,13 +32,34 @@ type ScanMetrics struct {
 	lastScanFileCount         int
 }
 
+type ScanStatus struct {
+	// finished channel is closed once the scan has finished
+	finished chan struct{}
+
+	// isRunning is true when the scan is either running or waiting to run, and changed to false when it's done
+	isRunning bool
+
+	// isPending is true when the scan is currently waiting for a previous scan to finish
+	isPending bool
+}
+
+func NewScanStatus() *ScanStatus {
+	return &ScanStatus{
+		finished:  make(chan struct{}),
+		isRunning: false,
+		isPending: false,
+	}
+}
+
 type Scanner struct {
-	BundleUploader *BundleUploader
-	SnykApiClient  snyk_api.SnykApiClient
-	errorReporter  error_reporting.ErrorReporter
-	analytics      ux2.Analytics
-	ignorePatterns []string
-	mutex          sync.Mutex
+	BundleUploader  *BundleUploader
+	SnykApiClient   snyk_api.SnykApiClient
+	errorReporter   error_reporting.ErrorReporter
+	analytics       ux2.Analytics
+	ignorePatterns  []string
+	mutex           sync.Mutex
+	scanStatusMutex sync.Mutex
+	runningScans    map[string]*ScanStatus
 }
 
 func New(bundleUploader *BundleUploader, apiClient snyk_api.SnykApiClient, reporter error_reporting.ErrorReporter, analytics ux2.Analytics) *Scanner {
@@ -64,6 +85,41 @@ func (sc *Scanner) SupportedCommands() []snyk.CommandName {
 }
 
 func (sc *Scanner) Scan(ctx context.Context, _ string, folderPath string) []snyk.Issue {
+	// When starting a scan for a folderPath that's already scanned, the new scan will wait for the previous scan
+	// to finish before starting.
+	// When there's already a scan waiting, the function returns immediately with empty results.
+	waitForPreviousScan := false
+	scanStatus := NewScanStatus()
+	scanStatus.isRunning = true
+	sc.scanStatusMutex.Lock()
+	previousScanStatus, wasFound := sc.runningScans[folderPath]
+	if wasFound && previousScanStatus.isRunning {
+		if previousScanStatus.isPending {
+			return []snyk.Issue{}
+		}
+
+		waitForPreviousScan = true
+		scanStatus.isPending = true
+	}
+
+	sc.runningScans[folderPath] = scanStatus
+	sc.scanStatusMutex.Unlock()
+	if waitForPreviousScan {
+		<-previousScanStatus.finished // Block here until previous scan is finished
+
+		// Setting isPending = false allows for future scans to wait for the current
+		// scan to finish, instead of returning immediately
+		sc.scanStatusMutex.Lock()
+		scanStatus.isPending = false
+		sc.scanStatusMutex.Unlock()
+	}
+	defer func() {
+		sc.scanStatusMutex.Lock()
+		scanStatus.isRunning = false
+		close(scanStatus.finished)
+		sc.scanStatusMutex.Unlock()
+	}()
+
 	startTime := time.Now()
 	span := sc.BundleUploader.instrumentor.StartSpan(ctx, "code.ScanWorkspace")
 	defer sc.BundleUploader.instrumentor.Finish(span)
@@ -78,7 +134,8 @@ func (sc *Scanner) Scan(ctx context.Context, _ string, folderPath string) []snyk
 	}
 
 	metrics := sc.newMetrics(len(files), startTime)
-	return sc.UploadAndAnalyze(span.Context(), files, folderPath, metrics)
+	results := sc.UploadAndAnalyze(span.Context(), files, folderPath, metrics)
+	return results
 }
 
 func (sc *Scanner) files(folderPath string) (filePaths []string, err error) {
--- a/infrastructure/code/code.go
+++ b/infrastructure/code/code.go
@@ -68,6 +68,7 @@ func New(bundleUploader *BundleUploader, apiClient snyk_api.SnykApiClient, repor
 		SnykApiClient:  apiClient,
 		errorReporter:  reporter,
 		analytics:      analytics,
+		runningScans:   map[string]*ScanStatus{},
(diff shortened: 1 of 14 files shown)


infrastructure/code/code.go
```diff
@@ -32,13 +32,34 @@ type ScanMetrics struct {
...
-	analytics      ux2.Analytics
-	ignorePatterns []string
-	mutex          sync.Mutex
+	BundleUploader  *BundleUploader
+	SnykApiClient   snyk_api.SnykApiClient
+	errorReporter   error_reporting.ErrorReporter
+	analytics       ux2.Analytics
+	ignorePatterns  []string
+	mutex           sync.Mutex
+	scanStatusMutex sync.Mutex
```
bastiandoetsch, line 61: Is it really necessary to have a second mutex here?

infrastructure/code/code.go
```diff
@@ -64,6 +86,41 @@ func (sc *Scanner) SupportedCommands() []snyk.CommandName {
 }
 
 func (sc *Scanner) Scan(ctx context.Context, _ string, folderPath string) []snyk.Issue {
+	// When starting a scan for a folderPath that's already scanned, the new scan will wait for the previous scan
+	// to finish before starting.
+	// When there's already a scan waiting, the function returns immediately with empty results.
+	waitForPreviousScan := false
+	scanStatus := NewScanStatus()
+	scanStatus.isRunning = true
+	sc.scanStatusMutex.Lock()
```
bastiandoetsch, line 95: I may be missing something, but there should be a `defer sc.scanStatusMutex.Unlock()` so the function is never exited without releasing the lock.

infrastructure/code/code.go
```diff
@@ -64,6 +86,41 @@ func (sc *Scanner) SupportedCommands() []snyk.CommandName {
...
+	}
+
+	sc.runningScans[folderPath] = scanStatus
+	sc.scanStatusMutex.Unlock()
+	if waitForPreviousScan {
+		<-previousScanStatus.finished // Block here until previous scan is finished
+
+		// Setting isPending = false allows for future scans to wait for the current
+		// scan to finish, instead of returning immediately
+		sc.scanStatusMutex.Lock()
```
bastiandoetsch, line 113: Does that work, when you've already locked it above?

infrastructure/code/code.go
```diff
@@ -64,6 +86,41 @@ func (sc *Scanner) SupportedCommands() []snyk.CommandName {
...
+	// to finish before starting.
+	// When there's already a scan waiting, the function returns immediately with empty results.
+	waitForPreviousScan := false
+	scanStatus := NewScanStatus()
+	scanStatus.isRunning = true
+	sc.scanStatusMutex.Lock()
+	previousScanStatus, wasFound := sc.runningScans[folderPath]
+	if wasFound && previousScanStatus.isRunning {
+		if previousScanStatus.isPending {
+			return []snyk.Issue{}
```
bastiandoetsch, line 99: Here you're missing an unlock, I think.

infrastructure/code/code.go
```diff
@@ -64,6 +86,41 @@ func (sc *Scanner) SupportedCommands() []snyk.CommandName {
...
+	}
+	defer func() {
+		sc.scanStatusMutex.Lock()
+		scanStatus.isRunning = false
+		close(scanStatus.finished)
+		sc.scanStatusMutex.Unlock()
+	}()
+
 	startTime := time.Now()
 	span := sc.BundleUploader.instrumentor.StartSpan(ctx, "code.ScanWorkspace")
```
bastiandoetsch, line 125: shouldn't this be the first thing done? Else the milliseconds spend before are ignored, which is not the semantics of the performance measurement.

infrastructure/iac/iac.go
```diff
@@ -73,33 +78,64 @@ func (iac *Scanner) SupportedCommands() []snyk.CommandName {
 func (iac *Scanner) Scan(ctx context.Context, path string, _ string) (issues []snyk.Issue) {
 	if ctx.Err() != nil {
 		log.Info().Msg("Cancelling IAC scan - IAC scanner received cancellation signal")
-		return []snyk.Issue{}
+		return
```
bastiandoetsch, line 81: syntactically, i'd prefer `return issues`

infrastructure/iac/iac.go
```diff
@@ -73,33 +78,64 @@ func (iac *Scanner) SupportedCommands() []snyk.CommandName {
...
 	p.BeginUnquantifiableLength("Scanning for Snyk IaC issues", path)
-	defer p.End("Snyk Iac Scan completed.")
 
 	var workspacePath string
 	if uri.IsDirectory(documentURI) {
 		workspacePath = uri.PathFromUri(documentURI)
 	} else {
 		workspacePath = filepath.Dir(uri.PathFromUri(documentURI))
 	}
+	iac.mutex.Lock()
```
bastiandoetsch, line 100: If the previous scan is running, how can the new scan relock the mutex here and cancel it?

bastiandoetsch: Hm... I see, the runningScans map will do that.

infrastructure/iac/iac.go
```diff
@@ -117,12 +153,16 @@ func (iac *Scanner) doScan(ctx context.Context, documentURI sglsp.DocumentURI, w
 	s := iac.instrumentor.StartSpan(ctx, method)
 	defer iac.instrumentor.Finish(s)
 
-	iac.mutex.Lock()
-	defer iac.mutex.Unlock()
+	// iac.mutex.Lock()
```
bastiandoetsch, line 156: This can be removed instead of commented.

infrastructure/oss/scanner.go
```diff
@@ -72,11 +74,16 @@ var supportedFiles = map[string]bool{
...
+// a counter for scans, used for logging
+var scanCount = 1
+
 type Scanner struct {
 	instrumentor  performance.Instrumentor
 	errorReporter error_reporting.ErrorReporter
 	analytics     ux2.Analytics
 	cli           cli.Executor
+	mutex         *sync.Mutex
+	runningScans  map[string]*scans.ScanProgress
```
bastiandoetsch, line 86: Would it make sense to use an AtomicMap here to automatically synchronize access?

infrastructure/oss/scanner.go
```diff
@@ -169,35 +207,33 @@ func (oss *Scanner) unmarshallAndRetrieveAnalysis(res []byte, documentURI sglsp.
...
-			return
+			return nil
 		}
-		issues = oss.retrieveIssues(scanResult, targetFileUri, fileContent)
-
-		oss.trackResult(true)
+		issues = append(issues, oss.retrieveIssues(scanResult, targetFileUri, fileContent)...)
 	}
+
+	oss.trackResult(true)
```
bastiandoetsch, line 215: nice catch

internal/scans/scan_progress.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func NewScanProgress() *ScanProgress {
+	return &ScanProgress{
+		cancel: make(chan struct{}),
+		done:   make(chan struct{}),
+	}
+}
+
+func (rs *ScanProgress) GetDoneChannel() <-chan struct{}   { return rs.done }
```
bastiandoetsch, line 28: why not a typed channel?

asaf92: I don't have any data to pass so it's essentially a signal-only channel.

I can switch to `chan bool` and write `true` if it's less confusing/ugly

bastiandoetsch: yeah, I'd prefer `bool` - easier on the eyes and to understand I think.

infrastructure/oss/scanner.go
asaf92: I don't think so

We're only accessing this map inside locked sections, and there are more actions being done inside these locked sections that are unrelated to the map (for example checking the `ScanProgress` value

infrastructure/code/code.go
asaf92: The 1st mutex is locking a section inside `files` on a section that updates the `ignorePatterns` field, the 2nd one is for synchronizing new scan requests

infrastructure/code/code.go
asaf92: I'm not sure if we can use defer here because the mutex is unlocked on line 108, and there's a deferred function that locks and unlocks the mutes on lines 118-123.

I'll add an unlock before returning + tests

infrastructure/code/code.go
asaf92: The problem is that on line 109 we block execution until the previous scan finished, so that's probably not a time we want to measure

infrastructure/iac/iac.go
```diff
@@ -73,33 +78,64 @@ func (iac *Scanner) SupportedCommands() []snyk.CommandName {
...
+			return
 		}
 	}
+
+	issues = iac.retrieveIssues(ctx, scanResults, issues, workspacePath, err)
+	iac.mutex.Lock()
+	log.Debug().Msgf("Scan %v is done", i)
+	newScan.SetDone()
+	iac.mutex.Unlock()
+	p.End("Snyk Iac Scan completed.")
```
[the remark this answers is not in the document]
asaf92: I think I wanted to `p.End()` with a different message when the scan is cancelled, but I'll using `defer` is better here

infrastructure/iac/iac.go
asaf92: I'll uncomment these because of the IAC restriction (+ added comment)

infrastructure/iac/iac.go
```diff
@@ -90,16 +98,46 @@ func (iac *Scanner) Scan(ctx context.Context, path string, _ string) (issues []s
...
 	}
+
+	issues = iac.retrieveIssues(ctx, scanResults, issues, workspacePath, err)
+	return issues
+}
+
+func (iac *Scanner) retrieveIssues(ctx context.Context, scanResults []iacScanResult, issues []snyk.Issue, workspacePath string, err error) []snyk.Issue {
 	if len(scanResults) > 0 {
 		for _, s := range scanResults {
+			if ctx.Err() != nil {
```
bastiandoetsch, line 137: I'm wondering if we should really cancel here, if we have results - the parsing only takes milliseconds, so we could just return the result and be done ;).

infrastructure/oss/scanner.go
```diff
@@ -134,33 +144,61 @@ func (oss *Scanner) Scan(ctx context.Context, path string, _ string) (issues []s
...
-	if done {
-		return
+	scanResults, err := oss.unmarshallOssJson(res)
+	if err != nil {
+		oss.errorReporter.CaptureError(err)
+		return nil
 	}
 
 	for _, scanResult := range scanResults {
+		if ctx.Err() != nil {
```
bastiandoetsch, line 198: Same here: cancelling here may be of very limited value, as the conversion is done in ms.

internal/scans/scan_progress.go
```diff
@@ -0,0 +1,100 @@
...
+const timeout = 5 * time.Second
+
+// ScanProgress is a type that's used for tracking current running scans, listen to their progress events
+// and invoke cancellations or "done" signals. This allows throttling or cancelling previous scans instead of
+// having many scans running at once for the same files.
+//
+// The correct usage would be to create a NewScanProgress and call "go Listen()" to
+// listen for cancel/done signals in the background.
+// ScanProgress operations should be done in locked sections.
+type ScanProgress struct {
```
bastiandoetsch, line 20: Not for today, but would be nice to integrate a progress tracker here, as we have some overlap in the functionality :)

bastiandoetsch (approved):
Very good job.

PR #145, opened 2022-09-28
feat: add cli concurrency limit [ROAD-1155]

j-luong:
### Description

_This PR should limit the concurrent CLI processes by using a semaphore. It also sets a 15 mins deadline on running CLI processes to guard against the CLI hanging._

### Checklist

- [x] Linted

### DIFF
--- a/infrastructure/cli/cli.go
+++ b/infrastructure/cli/cli.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"os"
 	"os/exec"
-	"runtime"
 	"strings"
 	"sync"
 	"time"
--- a/infrastructure/cli/cli.go
+++ b/infrastructure/cli/cli.go
@@ -4,8 +4,10 @@ import (
 	"context"
 	"os"
 	"os/exec"
+	"runtime"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/rs/zerolog/log"
 	sglsp "github.com/sourcegraph/go-lsp"
@@ -33,15 +35,24 @@ type SnykCli struct {
 	authenticator snyk.AuthenticationService
 	errorReporter error_reporting.ErrorReporter
 	analytics     ux.Analytics
+	semaphore     chan int
 }
+
+const cliTimeout = 15 * time.Minute
 
 var Mutex = &sync.Mutex{}
 
 func NewExecutor(authenticator snyk.AuthenticationService, errorReporter error_reporting.ErrorReporter, analytics ux.Analytics) Executor {
+	concurrencyLimit := runtime.NumCPU()
+	if concurrencyLimit < 1 {
+		concurrencyLimit = 1
+	}
+
 	return &SnykCli{
 		authenticator,
 		errorReporter,
 		analytics,
+		make(chan int, concurrencyLimit),
 	}
 }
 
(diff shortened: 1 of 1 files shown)


infrastructure/cli/cli.go
```diff
@@ -33,15 +35,24 @@ type SnykCli struct {
...
 	analytics     ux.Analytics
+	semaphore     chan int
 }
 
+const cliTimeout = 15 * time.Minute
+
 var Mutex = &sync.Mutex{}
 
 func NewExecutor(authenticator snyk.AuthenticationService, errorReporter error_reporting.ErrorReporter, analytics ux.Analytics) Executor {
+	concurrencyLimit := runtime.NumCPU()
```
michelkaporin, line 46: How do we limit to an only single spawned CLI **per product** in this solution?

michelkaporin:
Why we're still spawning multiple CLI processes for a workspace folder, where we can spawn a single one for all folders (e.g. in case of Open Source by using `snyk test path/to/folder1 path/to/folder2`)?

asaf92:
@michelkaporin #141 ensures 1 CLI process per workspace folder. This PR is enforcing a global limit on the amount of CLI processes. If there's 1 CLI process per workspace folder but there are 500 workspace folders, we want to avoid spawning 500 CLI processes at once.

Stacking scan requests and launching a single CLI is a possible optimization, but beyond the scope of this ticket for now unless we start seeing that it's necessary

bastiandoetsch:
I agree, having all scans passed to one CLI process would need a radical refactor in the open source scanner, e.g. running it agnostic of folders and actually providing a batch scan facility in the platform & interfaces.

infrastructure/cli/cli.go
bastiandoetsch: Yes, runtime.NumCPU() should not be used. One CLI process can easily take 5 cores on my machine, spawning several java executables and doing analysis itself. So we should limit concurrency to maybe a quarter of the cores at most.

michelkaporin:
Does it mean that if I have 10 workspace folders and 1 CPU, all my folders will be scanned one after another with this change? Don't think it's a much desired behaviour, but also works.

I thought we wanted rather to spawn a single process and run several scans as part a single process within this ticket. Sorry, was away during its decomp.

bastiandoetsch:
No, it was one option, but it is not necessarily the better option, as the CLI doesn't stream results. Meaning, you won't get any results before all folders are scanned - when you have 50 projects, you may want to work while the scan(s) are still running, so having everything in one scan does not improve the latency until you see results, it just saves some overhead regarding CLI spinup.

infrastructure/cli/cli.go
```diff
@@ -36,23 +36,20 @@ type SnykCli struct {
...
-		concurrencyLimit = 1
-	}
+	concurrencyLimit := 2
 
 	return &SnykCli{
 		authenticator,
 		errorReporter,
 		analytics,
 		make(chan int, concurrencyLimit),
+		90 * time.Minute, // TODO: add preference to make this configurable
```
michelkaporin, line 51: Let's have a ticket on this, please link it here as ROAD-xxx

michelkaporin (approved):

PR #148, opened 2022-10-11
feat: handle invalid user tokens [ROAD-1059]

asaf92:
### Description

Sends a message when the user is not authenticated.

### Checklist

- [ ] Linted

### DIFF
--- a/domain/snyk/scanner.go
+++ b/domain/snyk/scanner.go
@@ -69,7 +69,12 @@ func (sc *DelegatingConcurrentScanner) Scan(
 	done := make(chan bool)
 	defer close(done)
 
-	sc.initializer.Init()
+	err := sc.initializer.Init()
+	if err != nil {
+		log.Error().Err(err).Msg("Scan initialization error, cancelling scan")
+		return
+	}
+
 	tokenChangeChannel := config.CurrentConfig().TokenChangesChannel()
 	ctx, cancelFunc := context.WithCancel(ctx)
 	defer cancelFunc()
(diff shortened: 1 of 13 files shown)


infrastructure/cli/auth/initializer.go
```diff
@@ -45,28 +44,43 @@ func NewInitializer(authenticator snyk.AuthenticationService, errorReporter erro
...
 
 	notification.Send(sglsp.ShowMessageParams{Type: sglsp.Info, Message: "Authenticating to Snyk. This could open a browser window."})
 
-	token, err := i.authenticator.Provider().Authenticate(context.Background())
+	token, err := authenticator.Provider().Authenticate(context.Background())
 	if token == "" || err != nil {
+		if err != nil {
+			err = &AuthenticationFailedError{}
+		}
 		log.Error().Err(err).Msg("Failed to authenticate. Terminating server.")
```
asaf92, line 70: Change the log message

domain/snyk/scanner.go
```diff
@@ -69,7 +69,12 @@ func (sc *DelegatingConcurrentScanner) Scan(
 	done := make(chan bool)
 	defer close(done)
 
-	sc.initializer.Init()
+	err := sc.initializer.Init()
+	if err != nil {
+		log.Error().Err(err).Msg("Scan initialization error, cancelling scan")
```
bastiandoetsch, line 74: would logging to Sentry make sense here?

bastiandoetsch (commented):
looks good, we will merge later as talked about.

domain/snyk/scanner.go
asaf92: The initializers report to Sentry so I don't think there's a need

PR #149, opened 2022-10-12
fix: CLI path reset during initial download [ROAD-1183]

michelkaporin:
### Description

If server is started first time and CLI is not found, IDE can issue `WorkspaceDidChangeConfiguration` notification. At this point CLI path won't be set, resulting in [UpdateSettings](https://github.com/snyk/snyk-ls/blob/692b48a2a0c941df409e3ca8a2cea524e84eede6/application/server/configuration.go#L69) reseting CLI path in `CurrentConfig` during download.

When download finishes Installer would report empty path to the initializer, resulting in `Could not find, nor install Snyk CLI` error.

To solve this, we'll assume CLI path is default, whenever client (IDE) reports us an empty string for cliPath in `WorkspaceDidChangeConfiguration`.

### Checklist

- [x] Linted

### Screenshots / GIFs
image

### DIFF
--- a/application/config/config.go
+++ b/application/config/config.go
@@ -68,6 +68,12 @@ type CliSettings struct {
 	cliPathAccessMutex   sync.Mutex
 }
 
+func NewCliSettings() *CliSettings {
+	settings := &CliSettings{}
+	settings.SetPath("")
+	return settings
+}
+
 func (c *CliSettings) Installed() bool {
 	c.cliPathAccessMutex.Lock()
 	defer c.cliPathAccessMutex.Unlock()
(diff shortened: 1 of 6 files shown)


application/config/config.go
```diff
@@ -152,9 +155,8 @@ func IsDevelopment() bool {
 // New creates a configuration object with default values
 func New() *Config {
 	c := &Config{}
-	c.cliSettings = &CliSettings{
-		cliPath: filepath.Join(c.DefaultBinaryInstallPath(), filename.ExecutableName),
-	}
+	c.cliSettings = &CliSettings{}
+	c.cliSettings.SetPath("")
```
asaf92, line 159: Maybe better to have a `New()` function for CLI settings with all the default values

asaf92 (approved):

asaf92:
Fix the tests and we can merge

PR #155, opened 2022-10-17
fix: empty integration reported to CLI in auth provider [ROAD-1049]

michelkaporin:
### Description

CLI crafts authentication url with wrong UTM parameters when LS is launched via VS Code (see `CLI_V1_PLUGIN`):
`https://app.snyk.io/login/cli?token=4bc3ddde-1209-4fef-b0e3-4a64fedc325c&utm_medium=cli&utm_source=cli&utm_campaign=CLI_V1_PLUGIN&utm_campaign_content=1.1032.0&os=darwin&docker=false`
vs. Eclipse:
`https://app.snyk.io/login/cli?token=4e29b656-90cb-42e5-90af-ce3512c19d59&utm_medium=cli&utm_source=cli&utm_campaign=ECLIPSE&utm_campaign_content=2.0.0.202209271622&os=darwin&docker=false`

### Checklist

- [ ] Linted

### DIFF
--- a/infrastructure/cli/environment.go
+++ b/infrastructure/cli/environment.go
@@ -2,10 +2,23 @@ package cli
 
 import "github.com/snyk/snyk-ls/application/config"
 
+const (
+	OrganizationEnvVar                  = "SNYK_CFG_ORG"
+	ApiEnvVar                           = "SNYK_API"
+	TokenEnvVar                         = "SNYK_TOKEN"
+	DisableAnalyticsEnvVar              = "SNYK_CFG_DISABLE_ANALYTICS"
+	IntegrationNameEnvVarKey            = "SNYK_INTEGRATION_NAME"
+	IntegrationVersionEnvVarKey         = "SNYK_INTEGRATION_VERSION"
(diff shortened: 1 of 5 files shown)


infrastructure/cli/environment.go
```diff
@@ -0,0 +1,33 @@
+package cli
+
+import "github.com/snyk/snyk-ls/application/config"
+
+// Returns the input array with additional variables used in the CLI run in the form of "key=value".
+// Since we append, our values are overwriting existing env variables (because exec.Cmd.Env chooses the last value
+// in case of key duplications).
+func AppendCliEnvironmentVariables(currentEnv []string) (updatedEnv []string) {
```
j-luong, line 8: are/should this fn be tested for all cases?

michelkaporin: Fair point! I'll add tests around the func 👍

michelkaporin: In fact there were few tests already, I've moved them over. Also remembered that the reason why we had two different places for environment passing was because token shouldn't be passed in authentication provider. Corrected this.

PR #167, opened 2022-10-20
feat: auto-auth disabled notification only when token is invalid [ROAD-1059]

asaf92:
### Description
In case auto-auth is disabled, only send a notification when the token provided is invalid and not when the user hasn't authenticated yet.

### DIFF
--- a/infrastructure/cli/auth/initializer.go
+++ b/infrastructure/cli/auth/initializer.go
@@ -70,7 +70,7 @@ func (i *Initializer) Init() error {
 		return errors.New(errorMessage)
 	}
 
-	notification.Send(sglsp.ShowMessageParams{Type: sglsp.Info, Message: "Authenticating to Snyk. This could open a browser window."})
+	notification.SendShowMessage(sglsp.Info, "Authenticating to Snyk. This could open a browser window.")
 
 	token, err := authenticator.Provider().Authenticate(context.Background())
 	if token == "" || err != nil {
--- a/application/server/lsp/message_types.go
+++ b/application/server/lsp/message_types.go
@@ -161,6 +161,8 @@ type InitializeParams struct {
 	WorkspaceFolders []WorkspaceFolder `json:"workspaceFolders,omitempty"`
 }
 
+type InitializedParams struct{}
+
 type ServerCapabilities struct {
 	TextDocumentSync                 *sglsp.TextDocumentSyncOptionsOrKind   `json:"textDocumentSync,omitempty"`
 	HoverProvider                    bool                                   `json:"hoverProvider,omitempty"`
--- a/application/server/server.go
+++ b/application/server/server.go
@@ -78,6 +78,7 @@ func Start() {
 
 func initHandlers(srv *jrpc2.Server, handlers *handler.Map) {
 	(*handlers)["initialize"] = InitializeHandler(srv)
+	(*handlers)["initialized"] = InitializedHandler(srv)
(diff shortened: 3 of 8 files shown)


infrastructure/cli/auth/initializer.go
```diff
@@ -59,8 +59,10 @@ func (i *Initializer) Init() error {
 		return nil
 	}
 	if !currentConfig.AutomaticAuthentication() {
-		notification.Send(sglsp.ShowMessageParams{Message: "Please authenticate your Snyk user in order to start scanning"})
-		log.Info().Msg("Skipping authentication - automatic authentication is disabled")
+		if currentConfig.Token() != "" { // Only send notification when the token is invalid
+			notification.SendError(&AuthenticationFailedError{})
```
michelkaporin, line 63: This will print `You can reset the token to re-authenticate automatically` in notification. That is wrong, since it's non-automatic auth case.

infrastructure/cli/auth/initializer.go
```diff
@@ -59,8 +59,10 @@ func (i *Initializer) Init() error {
...
+		}
+		log.Info().Msg("Skipping scan - user is not authenticated and automatic authentication is disabled")
 		return errors.New(errorMessage)
```
michelkaporin, line 66: Why is it an error, if automatic authentication is disabled?

michelkaporin (changes requested):

infrastructure/cli/auth/initializer.go
asaf92: Because the user doesn't have a valid token, so we cannot start a scan.
When the initializers return an error, the scan is skipped.

michelkaporin: Semantically it's not an error. It's just a normal case, when user is not authenticated. I would return `nil` here.

michelkaporin (approved):
lgtm, with a tiny comment

domain/snyk/scanner_test.go
```diff
@@ -104,7 +104,7 @@ func Test_userNotAuthenticated_ScanSkipped(t *testing.T) {
 		productScanner,
 	)
 	config.CurrentConfig().SetToken("")
-	isAuthenticated := config.CurrentConfig().Authenticated()
+	isAuthenticated := config.CurrentConfig().NonEmptyToken()
```
bastiandoetsch, line 107: I guess that variable name should be renamed, too :)

infrastructure/cli/auth/initializer.go
```diff
@@ -54,13 +54,19 @@ func (i *Initializer) Init() error {
...
 	isAuthenticated, _ := authenticator.IsAuthenticated()
-	if currentConfig.Authenticated() && isAuthenticated {
+	if currentConfig.NonEmptyToken() && isAuthenticated {
 		log.Info().Msg("Skipping authentication - user is already authenticated")
 		return nil
 	}
 	if !currentConfig.AutomaticAuthentication() {
-		notification.Send(sglsp.ShowMessageParams{Message: "Please authenticate your Snyk user in order to start scanning"})
-		log.Info().Msg("Skipping authentication - automatic authentication is disabled")
+		if currentConfig.Token() != "" { // Only send notification when the token is invalid
```
bastiandoetsch, line 62: why not `currentContif.NonEmptyToken()`?

bastiandoetsch (approved):

PR #172, opened 2022-10-25
feat: scan after server initialization

michelkaporin:
### Description

At the moment when server is responding to initialization request, it runs workspace scan. This results in notifications (e.g. `SnykIsAvailableCli`) to be sent during initialization.

Unfortunately, VS Code language client `v8.0.0-next.2` is not ready to react to notifications during initialization process, since it doesn't see server in the readiness state (see [here](https://github.com/microsoft/vscode-languageserver-node/blob/cdf4d6fdaefe329ce417621cf0f8b14e0b9bb39d/client/src/common/client.ts#L3307) when server is ready to accept notifications).

The sibling PR for VS Code, fixing the readiness state is https://github.com/snyk/vscode-extension/pull/298

### Checklist

- [ ] Linted

### DIFF
--- a/application/server/lsp/message_types.go
+++ b/application/server/lsp/message_types.go
@@ -161,6 +161,8 @@ type InitializeParams struct {
 	WorkspaceFolders []WorkspaceFolder `json:"workspaceFolders,omitempty"`
 }
 
+type InitializedParams struct{}
+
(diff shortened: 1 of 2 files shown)


bastiandoetsch (approved):
LGTM. Cleaner than before.

PR #173, opened 2022-10-27
fix: Ignore IAC errors about malformed json files [ROAD-1206]

asaf92:
### Description

- Ignore the errors about malformed JSON files in IAC scan when there are no IAC files to be scanned.
- Adapts to new CLI output when no issues were found
- Fixes download test, avoids reporting progress to many times to save some bandwith

### Checklist


j-luong:
It wasn't clear to me, but does this account for malformed IAC files? Or does it ignore any that match the error codes?

michelkaporin:
@j-luong it ignores results that are related to possibly malformed IaC files, that are marked with the error codes by IaC team. I.e. we don't want to error when IaC scan with CLI errors with these error codes.

j-luong:
Ok, I would consider putting that caveat in the README or other appropriate documentation, something like `IAC scans will ignore malformed IAC files` as it may be a potential source of confusion.

michelkaporin:
@j-luong I like being explicit to our customers about it.
@francescomari @YairZ101 is this documented somewhere within the IaC docs, such that customers are aware of such behaviour?

YairZ101:
> is this documented somewhere within the IaC docs, such that customers are aware of such behaviour?

No, the CLI doesn't ignore malformed IaC config files, we report them as errors so we never needed to document such behavior.

michelkaporin:
@YairZ101 right. This appears to be the desired behaviour in IDE plugins since introduction of error codes though. Does web ui also report them as errors? If not, is this documented somewhere?

Generally, what would be the right place for mentioning this?

YairZ101:
> Does web ui also report them as errors? If not, is this documented somewhere?

yes.

> Generally, what would be the right place for mentioning this?

In my opinion, because this is a specific implementation detail for the IDE then it should be documented somewhere in the IDE docs.

michelkaporin:
@YairZ101 thanks!
@j-luong I'll update Eclipse & IntelliJ docs after the code lands on `main` 👍

michelkaporin (approved):
Reviewed by @asaf92 ✅

PR #180, opened 2022-11-06
[ROAD-1227] fix: CLI path not provided error

asaf92:
### Description

Fixed an error where the `$/snyk.isAvailableCli` message is sent with an empty CLI path.

_(Note that the message is sent on line 117, so no further changes needed)_

### DIFF
--- a/application/server/server_test.go
+++ b/application/server/server_test.go
@@ -241,8 +241,8 @@ func Test_TextDocumentCodeLenses_shouldReturnCodeLenses(t *testing.T) {
 			ActivateSnykIac:             "false",
 			Organization:                "fancy org",
 			Token:                       "xxx",
-			ManageBinariesAutomatically: "false",
-			CliPath:                     "dummy",
+			ManageBinariesAutomatically: "true",
+			CliPath:                     "",
 		},
 	}
 	_, err := loc.Client.Call(ctx, "initialize", clientParams)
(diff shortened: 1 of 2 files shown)


michelkaporin:
@asaf92 what is the reason why we've sent this message before, when not installed?

asaf92:
> @asaf92 what is the reason why we've sent this message before, when not installed?

Probably by mistake. It was introduced in [this commit ](https://github.com/snyk/snyk-ls/commit/6a219d23a97660b4a7e179128a1c0b51e3a3e680) and the message is sent before checking for errors.

michelkaporin (approved):
LGTM 👍 Please merge it only once you've tested it in Eclipse / VS Code.

infrastructure/cli/initializer.go
```diff
@@ -46,13 +47,15 @@ func NewInitializer(errorReporter error_reporting.ErrorReporter, installer insta
 }
 
 func (i *Initializer) Init() error {
+	const errorMessage = "could not download CLI" // Returning an error notifies that the scan cannot continue
```
michelkaporin, line 50: "Could not initialize Snyk Language Server dependencies" may be more appropriate here, as we don't try to download when managed binaries are disabled.

michelkaporin (changes requested):

application/server/server_test.go
```diff
@@ -261,7 +261,7 @@ func Test_TextDocumentCodeLenses_shouldReturnCodeLenses(t *testing.T) {
 			path := uri.PathFromUri(didOpenParams.TextDocument.URI)
 			return workspace.Get().GetFolderContaining(path).DocumentDiagnosticsFromCache(path) != nil
 		},
-		5*time.Second,
+		50*time.Second,
```
michelkaporin, line 264: Why such a drastic change to the wait time?

asaf92: Because it might actually download the CLI

michelkaporin: And it wasn't previously? 👀 

Let's mark this as smoke test maybe then, since we have this category for long-running tests.

michelkaporin (approved):

PR #183, opened 2022-11-13
feat: Silence IaC error [ROAD-1142]

asaf92:
### Description

Ignore the error about not finding IAC files

### DIFF
--- a/application/server/server.go
+++ b/application/server/server.go
@@ -101,7 +101,15 @@ func initHandlers(srv *jrpc2.Server, handlers *handler.Map) {
 // when files are deleted
 func WorkspaceWillDeleteFilesHandler() jrpc2.Handler {
 	return handler.New(func(ctx context.Context, params lsp.DeleteFilesParams) (interface{}, error) {
-		log.Info().Msg("Handling file deletions")
+		ws := workspace.Get()
+		for _, file := range params.Files {
+			path := uri.PathFromUri(file.Uri)
+
+			// Instead of branching whether it's a file or a folder, we'll attempt to remove both and the redundant case
+			// will be a no-op
+			ws.DeleteFolder(path)
+			ws.DeleteFile(path)
+		}
 		return nil, nil
 	})
 }
@@ -360,7 +368,7 @@ func TextDocumentDidSaveHandler() jrpc2.Handler {
 		// todo can we push cache management down?
 		f := workspace.Get().GetFolderContaining(filePath)
 		if f != nil {
-			f.ClearDiagnosticsCache(filePath)
+			f.ClearDiagnosticsFromFile(filePath)
 			di.HoverService().DeleteHover(params.TextDocument.URI)
 			go f.ScanFile(bgCtx, filePath)
 		} else {
--- a/application/server/server.go
+++ b/application/server/server.go
@@ -107,7 +107,7 @@ func WorkspaceWillDeleteFilesHandler() jrpc2.Handler {
 
 			// Instead of branching whether it's a file or a folder, we'll attempt to remove both and the redundant case
 			// will be a no-op
-			ws.DeleteFolder(path)
+			ws.RemoveFolder(path)
 			ws.DeleteFile(path)
 		}
 		return nil, nil
--- a/application/server/server.go
+++ b/application/server/server.go
@@ -91,9 +91,19 @@ func initHandlers(srv *jrpc2.Server, handlers *handler.Map) {
 	(*handlers)["shutdown"] = Shutdown()
 	(*handlers)["exit"] = Exit(srv)
 	(*handlers)["workspace/didChangeWorkspaceFolders"] = WorkspaceDidChangeWorkspaceFoldersHandler()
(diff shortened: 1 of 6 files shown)


application/server/server.go
```diff
@@ -91,11 +91,29 @@ func initHandlers(srv *jrpc2.Server, handlers *handler.Map) {
...
 	(*handlers)["exit"] = Exit(srv)
 	(*handlers)["workspace/didChangeWorkspaceFolders"] = WorkspaceDidChangeWorkspaceFoldersHandler()
+	(*handlers)["workspace/willDeleteFiles"] = WorkspaceWillDeleteFilesHandler()
 	(*handlers)["workspace/didChangeConfiguration"] = WorkspaceDidChangeConfiguration(srv)
 	(*handlers)["window/workDoneProgress/cancel"] = WindowWorkDoneProgressCancelHandler()
 	(*handlers)["workspace/executeCommand"] = ExecuteCommandHandler(srv)
 }
 
+// WorkspaceWillDeleteFilesHandler handles the workspace/willDeleteFiles message that's raised by the client
+// when files are deleted
```
michelkaporin, line 101: Why not "workspace/didDeleteFiles", since we are interested in the end-result?

domain/ide/workspace/workspace.go
```diff
@@ -62,10 +62,21 @@ func Set(w *Workspace) {
 	instance = w
 }
 
-func (w *Workspace) DeleteFolder(folder string) {
+func (w *Workspace) DeleteFolder(folderPath string) {
 	w.mutex.Lock()
 	defer w.mutex.Unlock()
-	delete(w.folders, folder)
+	folder := w.GetFolderContaining(folderPath)
+	folder.ClearDiagnosticsFromPathRecursively(folderPath)
```
michelkaporin, line 69: How about we do null-check folder here, similarly to the method below?

domain/ide/workspace/workspace.go
```diff
@@ -62,10 +62,21 @@ func Set(w *Workspace) {
...
+	delete(w.folders, folderPath)
+}
+
+func (w *Workspace) DeleteFile(filePath string) {
+	w.mutex.Lock()
+	defer w.mutex.Unlock()
+	folder := w.GetFolderContaining(filePath)
+	if folder != nil {
+		folder.ClearDiagnosticsFromFile(filePath)
```
michelkaporin, line 78: This method resets folder scan status. Should we reset it when a file was deleted? What is the implication of that? 

Also `server.go` resets scan status for a folder (though marked as no-op).

domain/ide/workspace/folder.go
```diff
@@ -108,12 +109,33 @@ func (f *Folder) Contains(path string) bool {
...
+		Diagnostics: []lsp.Diagnostic{},
+	})
 	f.ClearScannedStatus()
 }
 
+func (f *Folder) ClearDiagnosticsFromPathRecursively(removedPath string) {
+	f.documentDiagnosticCache.Range(func(key interface{}, value interface{}) bool {
+		filePath := key.(string)
+		if strings.Contains(filePath, removedPath) {
+			f.documentDiagnosticCache.Delete(filePath)
```
michelkaporin, line 128: How about we reuse `ClearDiagnosticsFromFile` above?

There's also `ClearDiagnostics` method below in this file that does similar stuff.

application/server/server.go
asaf92: I don't think there's a reason to prefer one over the other, but if we remove the diagnostics after the file was removed than there might be a moment in time where the diagnostics are still showing in the IDE even though the files are no longer there, so I preferred this message.

domain/ide/workspace/workspace.go
asaf92: Clearing scanned status is preserving existing behavior. Not sure about the implications either but I didn't want to change it.

domain/ide/workspace/folder.go
asaf92: Done

domain/ide/workspace/workspace.go
asaf92: Done

domain/ide/workspace/workspace.go
asaf92: Actually, it looks like the scanned status is only used in tests 🤣

application/server/server.go
michelkaporin: Sounds fair. Does LSP4E support `workspace/willDeleteFiles`?

domain/ide/workspace/workspace.go
michelkaporin: Oh, okay :)

application/server/server.go
asaf92: It doesn't, but Eclipse removes the errors automatically

michelkaporin: If by errors you mean reported diagnostics, then 👌

michelkaporin (approved):
LGTM 🙆‍♂️

PR #184, opened 2022-11-15
feat: severity setting [ROAD-1143]

j-luong:
### Description

_This PR should implement filtering based on severity._

### Checklist

- [x] Linted

### DIFF
--- a/application/config/config.go
+++ b/application/config/config.go
@@ -32,10 +32,11 @@ import (
 	"github.com/denisbrodbeck/machineid"
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
-	"github.com/sourcegraph/go-lsp"
+	sglsp "github.com/sourcegraph/go-lsp"
 	"github.com/subosito/gotenv"
 	"github.com/xtgo/uuid"
 
+	"github.com/snyk/snyk-ls/application/server/lsp"
 	"github.com/snyk/snyk-ls/infrastructure/cli/filename"
 	"github.com/snyk/snyk-ls/internal/concurrency"
 	"github.com/snyk/snyk-ls/internal/product"
@@ -139,7 +140,7 @@ type Config struct {
 	snykCodeApiUrl               string
 	token                        string
 	deviceId                     string
-	clientCapabilities           lsp.ClientCapabilities
+	clientCapabilities           sglsp.ClientCapabilities
 	m                            sync.Mutex
 	path                         string
 	defaultDirs                  []string
@@ -147,6 +148,7 @@ type Config struct {
 	integrationVersion           string
 	automaticAuthentication      bool
 	tokenChangeChannels          []chan string
+	filterSeverity               lsp.SeverityFilter
 	trustedFolders               []string
 	trustedFoldersFeatureEnabled bool
 }
@@ -192,6 +194,7 @@ func New() *Config {
 	c.clientSettingsFromEnv()
 	c.deviceId = c.determineDeviceId()
 	c.addDefaults()
+	c.setDefaultSeverityFilter()
 	return c
 }
 
@@ -287,6 +290,7 @@ func (c *Config) SnykCodeApi() string                    { return c.snykCodeApiU
 func (c *Config) SnykCodeAnalysisTimeout() time.Duration { return c.snykCodeAnalysisTimeout }
 func (c *Config) IntegrationName() string                { return c.integrationName }
 func (c *Config) IntegrationVersion() string             { return c.integrationVersion }
+func (c *Config) FilterSeverity() lsp.SeverityFilter     { return c.filterSeverity }
 func (c *Config) Token() string {
 	c.m.Lock()
 	defer c.m.Unlock()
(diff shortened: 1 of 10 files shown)


application/config/config.go
```diff
@@ -146,6 +146,10 @@ type Config struct {
 	integrationVersion          string
 	automaticAuthentication     bool
 	tokenChangeChannels         []chan string
+	filterCriticalSeverity      concurrency.AtomicBool
```
michelkaporin, line 149: can we have a complex object in settings for this? That'd nicely group some of the settings :)

j-luong: done

application/config/config.go
```diff
@@ -188,6 +190,7 @@ func New() *Config {
 	c.clientSettingsFromEnv()
 	c.deviceId = c.determineDeviceId()
 	c.addDefaults()
+	c.setDefaultSeverityFilter()
```
bastiandoetsch, line 193: why not fold this into `addDefaults()`?

application/server/configuration.go
```diff
@@ -223,3 +224,19 @@ func updateProductEnablement(settings lsp.Settings) {
...
+	emptySeverityFilter := lsp.SeverityFilter{}
+	defaultSeverityFilter := lsp.SeverityFilter{
+		Critical: true,
+		High:     true,
+		Medium:   true,
+		Low:      true,
+	}
+
+	if s == emptySeverityFilter {
+		config.CurrentConfig().SetSeverityFilter(defaultSeverityFilter)
```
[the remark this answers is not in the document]
bastiandoetsch: Had a look at the acceptance criteria, unfortunately this scenario is missing.

domain/ide/workspace/folder.go
```diff
@@ -134,20 +135,53 @@ func (f *Folder) DocumentDiagnosticsFromCache(file string) []snyk.Issue {
...
+
 		f.documentDiagnosticCache.Put(issue.AffectedFilePath, cachedIssues)
-		issuesByFile[issue.AffectedFilePath] = cachedIssues.([]snyk.Issue)
+	}
+
+	// update issues by file
+	filteredIssues := []snyk.Issue{}
+	severityFilters := config.CurrentConfig().FilterSeverity()
+	// prevent panic: interface conversion: interface {} is nil, not []snyk.Issue
+	if cachedIssues == nil {
```
bastiandoetsch, line 161: So, in which constellation can this be nil? If there is more than one issue to be processed, the array is created in line 147. If `len(issues)` is 0, why are we even processing? Shouldn't we just early end the processing and return

domain/ide/workspace/folder_test.go
```diff
@@ -165,6 +167,57 @@ func TestProcessResults_whenSamePathsAndDuplicateIssues_DeDuplicates(t *testing.
...
+		{ID: "id3", AffectedFilePath: "path1", Severity: snyk.Medium},
+		{ID: "id4", AffectedFilePath: "path1", Severity: snyk.Low},
+		{ID: "id5", AffectedFilePath: "path1", Severity: snyk.Critical},
+	})
+
+	mtx := &sync.Mutex{}
+	var diagnostics []lsp.Diagnostic
+
+	defer notification.DisposeListener()
+	notification.CreateListener(func(event interface{}) {
```
bastiandoetsch, line 197: why not use the jsonRPCRecorder?

application/server/configuration.go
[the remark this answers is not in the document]
j-luong: It's to guard against the second thing; when no severity setting is passed/set. This was the case in some of the tests; but I think in reality, the client should always pass the initial severity settings when the LS client is created.

Actually, we can add the `if s == lsp.SeverityFilter{} { return }` in the `config.SetSeverityFilter()` instead, which would be cleaner.

domain/ide/workspace/folder.go
j-luong: It looks like this was a remnant from earlier in the development, just tried it without this check and it's working fine 👍 
If `len(issues) == 0` then we'd skip the filtering and process the diagnostics/hovers as we did previously.

domain/ide/workspace/folder_test.go
j-luong: I'm not familiar enough with it tbh and just followed the example set in the test below this one :)

domain/ide/workspace/folder.go
```diff
@@ -142,20 +142,54 @@ func (f *Folder) DocumentDiagnosticsFromCache(file string) []snyk.Issue {
 }
 
 func (f *Folder) processResults(issues []snyk.Issue) {
-	var issuesByFile = map[string][]snyk.Issue{}
+	if len(issues) == 0 {
+		f.documentDiagnosticCache.ClearAll()
+		return
+	}
```
asaf92, line 148: This is a bug. 
This function can be called from a scan of a sub-folder/specific file, but the cache holds issues for all files/folders. and the results get cleared

j-luong: 👍 this will be removed once #192 is merged, which will make this line redundant

asaf92 (approved):

PR #192, opened 2022-11-30
[ROAD-1235] fix: remove diagnostics when product disabled

asaf92:
### Description

Disabling a product in settings now removes the previously published diagnostics of that product

### DIFF
--- a/internal/product/product.go
+++ b/internal/product/product.go
@@ -1,0 +1,10 @@
+package product
+
+type Product string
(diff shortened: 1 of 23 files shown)


internal/product/product.go
```diff
@@ -0,0 +1,10 @@
+package product
+
+type Product string
+type ProductAttributes map[string]interface{}
```
asaf92, line 4: Check if possible to remove

PR #193, opened 2022-12-06
feat: filter issues by severity during runtime [ROAD-1143]

michelkaporin:
### Description

- Enables client to filter issues by severity during runtime.
- Centralises filter instantiation to simplify future refactors, if needed.

### Checklist

- [x] Linted

### Screenshots / GIFs
screen

### DIFF
--- a/domain/ide/workspace/workspace.go
+++ b/domain/ide/workspace/workspace.go
@@ -108,7 +108,7 @@ func (w *Workspace) GetFolderContaining(path string) (folder *Folder) {
 }
 
 func (w *Workspace) Folders() (folder []*Folder) {
-	folders := make([]*Folder, len(w.folders))
+	folders := make([]*Folder, 0, len(w.folders))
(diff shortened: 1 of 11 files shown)


domain/ide/workspace/workspace.go
```diff
@@ -107,6 +107,15 @@ func (w *Workspace) GetFolderContaining(path string) (folder *Folder) {
 	return nil
 }
 
+func (w *Workspace) Folders() (folder []*Folder) {
+	folders := make([]*Folder, 0)
```
asaf92, line 111: Probably not critical, but this can be optimized because we know the exact length of the array.

You can `make([]*Folder, len(w.folders))`. This is faster because append keeps resizing/recreating arrays.

michelkaporin: Nice catch, will update :)

asaf92 (approved):

PR #195, opened 2022-12-08
feat: set code quality/security in issueType field [ROAD-1161]

asaf92:
### Description

Allow users to filter between Snyk Code Quality and Snyk Code Security issues


domain/snyk/issues.go
```diff
@@ -44,7 +44,9 @@ type Issue struct {
 	AffectedFilePath string
 	// Product is the Snyk product, e.g. Snyk Open Source
 	Product product.Product // todo: can we avoid it, if it's part of a scanner interface already?
-	// References deliver additional information
+	// FilterableIssueType is the type of issue out of the user-facing categories for issue types.
+	FilterableIssueType product.FilterableIssueType
```
bastiandoetsch, line 48: Why differentiate between issue type and filterable issue type? can't all issue types be filtered?

asaf92: I removed this field and added a function that returns a FilterableIssueType. 

The reason there are two types is because the other "IssueType" has stuff like "License/Container/etc" which are not filterable

bastiandoetsch (approved):

PR #196, opened 2022-12-08
feat: add Amplitude SDK [ROAD-1179]

michelkaporin:
### Description

- Wires Golang Amplitude SDK with [Ampli](https://www.docs.developers.amplitude.com/data/ampli/cli/) [Ampli Wrapper](https://www.docs.developers.amplitude.com/data/sdks/go/ampli/) generated types. 
- Adds tracking plan verification step as part of the release pipeline. I didn't put it on purpose into the build workflow to be cautious of our CI time. This is also on par with how we did it in IDE plugins.
- Fixes IDE property for VS Code integration.

### Checklist

- [x] Linted


.github/workflows/release.yaml
```diff
@@ -107,6 +112,11 @@ jobs:
 
           make clean test
 
+      - name: Verify analytics
```
bastiandoetsch, line 115: shouldn't that run even before the tests, so we don't lose time?

infrastructure/amplitude/client.go
```diff
@@ -0,0 +1,191 @@
...
+
+import (
+	"strings"
+
+	"github.com/amplitude/analytics-go/amplitude"
+	"github.com/amplitude/analytics-go/amplitude/plugins/destination"
+	"github.com/pkg/errors"
+	"github.com/rs/zerolog/log"
+
+	// segment "github.com/segmentio/analytics-go"
```
bastiandoetsch, line 27: This can probably go away

.github/workflows/release.yaml
michelkaporin: Put it after, since it's less likely to fail than tests (thus not increasing CI time), but it works the other way too 😄 Can shift it up.

infrastructure/amplitude/segment_plugin.go
```diff
@@ -0,0 +1,90 @@
+package amplitude
```
bastiandoetsch, line 1: This seems to be missing tests?

michelkaporin: `client_test.go` encompasses tests for this file, didn't want to introduce another set of same tests here.

bastiandoetsch: Doesn't client_test use fakesegmentclient?

michelkaporin: Correct, with the Amplitude client [wiring the Segment plugin](https://github.com/snyk/snyk-ls/blob/2ed76537e9ba1fe39da665e5ec2386ce924f97a1/infrastructure/amplitude/client.go#L56). I don't want to test Amplitude internals here, but if you have some specific cases you can think of that would break, please let me know.

michelkaporin: [This test](https://github.com/snyk/snyk-ls/blob/2ed76537e9ba1fe39da665e5ec2386ce924f97a1/infrastructure/amplitude/client_test.go#L68) exactly tests if an event gets enqueues with a user id. It uses a real Amplitude client with Segment plugin and mock Segment API.

Segment plugin is just a dumb plugin wrapping Segment API.

bastiandoetsch (approved):

PR #197, opened 2022-12-09
chore: update copyright notice and cleanup warnings

bastiandoetsch:
### Description

Update to copyright notice to use copyright symbol, cleaned up most warnings.

### Checklist

- [x] Linted

### Screenshots / GIFs


application/config/client_settings_test.go
```diff
@@ -29,11 +29,11 @@ func TestGetEnabledProducts_DefaultValues(t *testing.T) {
...
-	os.Unsetenv(ActivateSnykOssKey)
-	os.Unsetenv(ActivateSnykCodeKey)
-	os.Unsetenv(ActivateSnykIacKey)
-	os.Unsetenv(ActivateSnykContainerKey)
-	os.Unsetenv(ActivateSnykAdvisorKey)
+	_ = os.Unsetenv(ActivateSnykOssKey)
+	_ = os.Unsetenv(ActivateSnykCodeKey)
+	_ = os.Unsetenv(ActivateSnykIacKey)
+	_ = os.Unsetenv(ActivateSnykContainerKey)
+	_ = os.Unsetenv(ActivateSnykAdvisorKey)
```
asaf92, line 36: Why do we do this? Can't we use `t.Setenv(key, "")`?

bastiandoetsch: Empty string is not the same as no variable. I think that's why.

infrastructure/code/bundle_uploader_test.go
```diff
@@ -32,7 +32,9 @@ import (
 func Test_Bundler_Upload(t *testing.T) {
 	temporaryDir := setup(t)
 	t.Cleanup(func() {
-		defer os.RemoveAll(temporaryDir)
+		defer func(path string) {
+			_ = os.RemoveAll(path)
+		}(temporaryDir)
```
asaf92, line 37:
```suggestion
		_ = os.RemoveAll(temporaryDir)
```

infrastructure/code/encoding/encoder.go
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022 Snyk Ltd.
+ * © 2022 Snyk Limited All rights reserved.
```
asaf92, line 2: What's with the `Â`?

asaf92 (commented):
There's some special character that got pulled in (Â). Also, can we make all the error-ignoring stuff 1 line instead of 3?

application/config/client_settings_test.go
asaf92: I tried locally and the test passed.

I think the only difference between an empty env-var and an unset one is that `os.LookupEnv("EMPTY_VAR_KEY")` will return a 2nd output value that's a boolean specifying that the env-var exists, and this is only used in `Config.loadFile`. So basically if you have an empty env var it would not be overwritten by the value in the config file.

I think it's better if we don't differentiate between an empty env-var and a missing one.

I don't think that we have to address it in this PR though

infrastructure/code/encoding/encoder.go
asaf92: Strange, I no longer see the special character 😮

bastiandoetsch:
> There's some special character that got pulled in (Â). Also, can we make all the error-ignoring stuff 1 line instead of 3?

This is extremely weird. If you tell github to show the file, the Â character is not there. On my computer, I can't see it either - it seems to be a problem with the github diff view.

PR #201, opened 2022-12-19
fix: tests with fake cli installer

bastiandoetsch:
### Description

tests needed some clean up, as some were expecting, or were affected by results from previously executed tests

Also added some debug logging

### Checklist

- [x] Linted

### Screenshots / GIFs

### DIFF
--- a/internal/testutil/test_setup.go
+++ b/internal/testutil/test_setup.go
@@ -21,6 +21,8 @@ import (
 	"runtime"
 	"testing"
 
+	"github.com/rs/zerolog/log"
+
 	"github.com/snyk/snyk-ls/application/config"
 	"github.com/snyk/snyk-ls/internal/notification"
 	"github.com/snyk/snyk-ls/internal/progress"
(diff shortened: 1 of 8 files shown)


internal/testutil/test_setup.go
```diff
@@ -42,13 +44,29 @@ func SmokeTest(t *testing.T) {
...
 	CLIDownloadLockFileCleanUp(t)
 	notification.DisposeListener()
-	t.Cleanup(func() { notification.DisposeListener() })
+	t.Cleanup(func() {
+		notification.DisposeListener()
+		cleanupFakeCliFile(c)
+	})
+}
+
+func cleanupFakeCliFile(c *config.Config) {
```
PeterSchafer, line 58: Suggestion: How about moving this logic to FakeInstaller, so that creating the fake file and deleting it are close by.

PeterSchafer: @bastiandoetsch and I decided to postpone this and unblock this PR.

bastiandoetsch: Good idea, to move it. But, as discussed, let's investigate tomorrow, if we can move FakeInstaller instead.

PeterSchafer (approved):

PR #204, opened 2022-12-25
[ROAD-1216] Allow auto-scans disabling

asaf92:
### Description

Added a settings option to disable auto-scans


application/server/lsp/message_types.go
```diff
@@ -302,6 +302,7 @@ type Settings struct {
 	OsArch                      string         `json:"osArch,omitempty"`
 	RuntimeVersion              string         `json:"runtimeVersion,omitempty"`
 	RuntimeName                 string         `json:"runtimeName,omitempty"`
+	ScanningMode                string         `json:"scanningMode,omitempty"`
```
bastiandoetsch, line 305: why is this named differently than the config setting in `configuration`?

application/server/server.go
```diff
@@ -168,10 +168,11 @@ func WorkspaceDidChangeWorkspaceFoldersHandler(srv *jrpc2.Server) jrpc2.Handler
 		// The context provided by the JSON-RPC server is cancelled once a new message is being processed,
 		// so we don't want to propagate it to functions that start background operations
 		bgCtx := context.Background()
+		logger := log.With().Str("method", "WorkspaceDidChangeWorkspaceFoldersHandler").Logger()
```
bastiandoetsch, line 171: interesting

application/server/server.go
```diff
@@ -380,19 +394,24 @@ func TextDocumentDidSaveHandler() jrpc2.Handler {
...
-		log.Info().Str("method", method).Interface("params", params).Msg("RECEIVING")
+		logger.Info().Interface("params", params).Msg("Receiving")
 		filePath := uri.PathFromUri(params.TextDocument.URI)
 
 		// todo can we push cache management down?
 		f := workspace.Get().GetFolderContaining(filePath)
-		if f != nil {
+		autoScanEnabled := config.CurrentConfig().IsAutoScanEnabled()
+		if f != nil && autoScanEnabled {
 			f.ClearDiagnosticsFromFile(filePath)
```
bastiandoetsch, line 406: Question: should we clear diagnostics on save even if not auto-scanning? Jetbrains plugins invalidate the results/caches after a save.

domain/ide/workspace/workspace.go
```diff
@@ -124,16 +124,20 @@ func (w *Workspace) ScanWorkspace(ctx context.Context) {
 	}
 }
 
-func (w *Workspace) AddAndRemoveFoldersAndTriggerScan(ctx context.Context, params lsp.DidChangeWorkspaceFoldersParams) {
+// ChangeWorkspaceFolders clears the "Removed" folders, adds the "New" folders,
+// and starts an automatic scan if auto-scans are enabled.
+func (w *Workspace) ChangeWorkspaceFolders(ctx context.Context, params lsp.DidChangeWorkspaceFoldersParams) {
```
bastiandoetsch, line 129: The old name was more explicit and easier to understand, IMHO.

bastiandoetsch:
AFAICT, this PR is missing tests for:
- analytics (scan mode)
- configuration update
- respecting auto-scan setting on save, open, startup

application/server/lsp/message_types.go
asaf92: This is a string, the config setting is a boolean specific to whether we do auto-scans or not.

It's easier to get a simple true/false about auto scans rather then comparing strings to "manual" everywhere.

application/server/server.go
asaf92: I think it would be very annoying if deleting a comment in some code file would immediately clear all the diagnostics for the workspace.

Notice that usually there's only one `Folder` for the entire workspace

domain/ide/workspace/workspace.go
asaf92: But now it's `AddAndRemoveFoldersAndTriggerScanIfAutoScanIsEnabled`

bastiandoetsch: It doesn't need to cover all conditions in the name ;). Right now, e.g. trust is not in there either.

application/server/server.go
bastiandoetsch: In Eclipse, there is a lot of folders usually in the workspace. But I agree that this has been causing issues every once in a while and I much prefer not to invalidate results.

application/server/configuration.go
```diff
@@ -162,6 +168,16 @@ func updateDeviceInformation(settings lsp.Settings) {
 	}
 }
 
+func updateAutoScan(settings lsp.Settings) {
+	// Auto scan true by default unless the AutoScan value in the settings is not missing & false
+	autoScan := true
+	if settings.ScanningMode == "manual" {
```
bastiandoetsch, line 174: should this be a constant maybe?

infrastructure/code/fake_snyk_code_api_service.go
```diff
@@ -81,16 +80,13 @@ var (
...
-func FakeDiagnosticPath(t *testing.T) (filePath string, path string) {
+func TempWorkdirWithVulnerabilities(t *testing.T) (filePath string, path string) {
 	FakeSnykCodeApiServiceMutex.Lock()
 	defer FakeSnykCodeApiServiceMutex.Unlock()
 
-	temp, err := os.MkdirTemp(xdg.DataHome, "fakeDiagnosticTempDir")
-	if err != nil {
-		t.Fatal(err, "couldn't create tempdir")
-	}
+	temp := t.TempDir()
```
bastiandoetsch, line 87: This sometimes breaks on macOS...but nevermind, I can live with that.

internal/testutil/test_setup.go
```diff
@@ -60,7 +62,10 @@ func cleanupFakeCliFile(c *config.Config) {
 	}
 	if stat.Size() < 1000 {
 		// this is a fake CLI, removing it
-		_ = os.Remove(c.CliSettings().Path())
+		err = os.Remove(c.CliSettings().Path())
+		if err != nil {
+			log.Warn().Err(err).Msg("Failed to remove fake CLI")
```
bastiandoetsch, line 67: I wouldn't add this - Had that previously and yielded in a lot of error messages.

bastiandoetsch (approved):

internal/testutil/test_setup.go
asaf92: The condition was flipped (`==` instead of `!=`). This commit fixes it

PR #212, opened 2023-01-13
fix: Use API /user/me to check if user is authenticated [ROAD-1236]

j-luong:
### Description

_This PR should change how we check a user is authenticated by using an API endpoint instead of CLI auth. This should stop auth checks from writing to the config store.._

### Checklist

- [x] Linted


michelkaporin (approved):
🚀

PR #213, opened 2023-01-23
fix: panic when connection fails [HEAD-80]

asaf92:
### Description

Fixes a panic that occurs when the connection returns an error


michelkaporin:
@asaf92 is `response.StatusCode` equal to `nil` when connection returns an error?

asaf92:
@michelkaporin 

> // On error, any Response can be ignored. A non-nil Response with a
> // non-nil error only occurs when CheckRedirect fails, and even then
> // the returned Response.Body is already closed.

PR #215, opened 2023-01-27
fix: 🐛 correct hover / position overlap calculation [HEAD-53]

bastiandoetsch:
### Description

Previously, hovers were not correctly calculated, when 

- the hover was multi-line and 
- the last char position on the last line was < cursor position

The change switches usage from sourcegraph ranges to our Snyk domain ranges, as that type already has working helper functions to test for overlap. Also, it removes duplication of range/overlap logic.

### Checklist

- [x] Linted

### Screenshots / GIFs

image

image


PeterSchafer (approved):
LGTM

PR #234, opened 2023-02-15
feat(code/autofix): add preliminary backend support for autofix [ML-62]

agronskiy:
### Description

- Added the structures modelling autofix requests and converters
- Added the backend service parts 
- Added simple tests
- Checked that integrations work

### Checklist

- [x] Linted
- [ ] ~~README.md updated, if user-facing~~
- [ ] ~~License file updated, if new 3rd-party dependency is introduced~~

### Screenshots / GIFs

### DIFF
--- a/application/config/config.go
+++ b/application/config/config.go
@@ -185,6 +185,10 @@ func IsDevelopment() bool {
 	return parseBool
 }
 
+func wippGetEnvSnykToken() string {
+	return os.Getenv("SNYK_TOKEN")
+}
+
 // New creates a configuration object with default values
 func New() *Config {
 	c := &Config{}
@@ -292,15 +296,22 @@ func (c *Config) Format() string { return c.format }
 func (c *Config) CLIDownloadLockFileName() string {
 	return filepath.Join(c.cliSettings.DefaultBinaryInstallPath(), "snyk-cli-download.lock")
 }
-func (c *Config) IsErrorReportingEnabled() bool          { return c.isErrorReportingEnabled.Get() }
-func (c *Config) IsSnykOssEnabled() bool                 { return c.isSnykOssEnabled.Get() }
-func (c *Config) IsSnykCodeEnabled() bool                { return c.isSnykCodeEnabled.Get() }
-func (c *Config) IsSnykIacEnabled() bool                 { return c.isSnykIacEnabled.Get() }
-func (c *Config) IsSnykContainerEnabled() bool           { return c.isSnykContainerEnabled.Get() }
-func (c *Config) IsSnykAdvisorEnabled() bool             { return c.isSnykAdvisorEnabled.Get() }
-func (c *Config) LogPath() string                        { return c.logPath }
-func (c *Config) SnykApi() string                        { return c.snykApiUrl }
-func (c *Config) SnykCodeApi() string                    { return c.snykCodeApiUrl }
+func (c *Config) IsErrorReportingEnabled() bool { return c.isErrorReportingEnabled.Get() }
+func (c *Config) IsSnykOssEnabled() bool        { return c.isSnykOssEnabled.Get() }
+func (c *Config) IsSnykCodeEnabled() bool {
+	// return c.isSnykCodeEnabled.Get()
+	return true // WIPP
+}
+func (c *Config) IsSnykIacEnabled() bool       { return c.isSnykIacEnabled.Get() }
+func (c *Config) IsSnykContainerEnabled() bool { return c.isSnykContainerEnabled.Get() }
+func (c *Config) IsSnykAdvisorEnabled() bool   { return c.isSnykAdvisorEnabled.Get() }
+func (c *Config) LogPath() string              { return c.logPath }
+func (c *Config) SnykApi() string              { return c.snykApiUrl }
+func (c *Config) SnykCodeApi() string {
+	return c.snykCodeApiUrl
+	// WIPP hardcoded local
+	// return "http://localhost:11981"
+}                                                        // WIPP TODO hardocode
 func (c *Config) SnykCodeAnalysisTimeout() time.Duration { return c.snykCodeAnalysisTimeout }
 func (c *Config) IntegrationName() string                { return c.integrationName }
 func (c *Config) IntegrationVersion() string             { return c.integrationVersion }
@@ -308,7 +319,8 @@ func (c *Config) FilterSeverity() lsp.SeverityFilter     { return c.filterSeveri
 func (c *Config) Token() string {
 	c.m.Lock()
 	defer c.m.Unlock()
-	return c.token
+	// WIPP hardcoded``
+	return wippGetEnvSnykToken()
 }
 
 // TokenChangesChannel returns a channel that will be written into once the token has changed.
@@ -431,6 +443,8 @@ func (c *Config) ConfigureLogging(level string) {
 		logLevel = envLevel
 	}
 	zerolog.SetGlobalLevel(logLevel)
+	// WIPP
+	zerolog.SetGlobalLevel(zerolog.DebugLevel)
 	zerolog.TimeFieldFormat = time.RFC3339
 
 	if c.logPath != "" {
@@ -636,7 +650,7 @@ func (c *Config) GetDisplayableIssueTypes() map[product.FilterableIssueType]bool
 }
 
 func (c *Config) IsSnykCodeSecurityEnabled() bool {
-	return c.activateSnykCodeSecurity
+	return true // WIPP
 }
 
 func (c *Config) EnableSnykCodeSecurity(activate bool) {
@@ -644,7 +658,7 @@ func (c *Config) EnableSnykCodeSecurity(activate bool) {
 }
 
 func (c *Config) IsSnykCodeQualityEnabled() bool {
-	return c.activateSnykCodeQuality
+	return true // WIPP
 }
 
 func (c *Config) EnableSnykCodeQuality(activate bool) {
--- a/application/config/config.go
+++ b/application/config/config.go
@@ -186,10 +186,6 @@ func IsDevelopment() bool {
 	return parseBool
 }
 
-func wippGetEnvSnykToken() string {
-	return os.Getenv("SNYK_TOKEN")
-}
-
 // New creates a configuration object with default values
 func New() *Config {
 	c := &Config{}
@@ -298,20 +294,16 @@ func (c *Config) Format() string { return c.format }
 func (c *Config) CLIDownloadLockFileName() string {
 	return filepath.Join(c.cliSettings.DefaultBinaryInstallPath(), "snyk-cli-download.lock")
 }
-func (c *Config) IsErrorReportingEnabled() bool { return c.isErrorReportingEnabled.Get() }
-func (c *Config) IsSnykOssEnabled() bool        { return c.isSnykOssEnabled.Get() }
-func (c *Config) IsSnykCodeEnabled() bool       { return c.isSnykCodeEnabled.Get() }
-func (c *Config) IsSnykIacEnabled() bool        { return c.isSnykIacEnabled.Get() }
-func (c *Config) IsSnykContainerEnabled() bool  { return c.isSnykContainerEnabled.Get() }
-func (c *Config) IsSnykAdvisorEnabled() bool    { return c.isSnykAdvisorEnabled.Get() }
-func (c *Config) IsSnykAutofixEnabled() bool    { return c.isSnykAutofixEnabled.Get() }
-func (c *Config) LogPath() string               { return c.logPath }
-func (c *Config) SnykApi() string               { return c.snykApiUrl }
-func (c *Config) SnykCodeApi() string {
-	return c.snykCodeApiUrl
-	// WIPP hardcoded local
-	// return "http://localhost:11981"
-}                                                        // WIPP TODO hardocode
+func (c *Config) IsErrorReportingEnabled() bool          { return c.isErrorReportingEnabled.Get() }
+func (c *Config) IsSnykOssEnabled() bool                 { return c.isSnykOssEnabled.Get() }
+func (c *Config) IsSnykCodeEnabled() bool                { return c.isSnykCodeEnabled.Get() }
+func (c *Config) IsSnykIacEnabled() bool                 { return c.isSnykIacEnabled.Get() }
+func (c *Config) IsSnykContainerEnabled() bool           { return c.isSnykContainerEnabled.Get() }
+func (c *Config) IsSnykAdvisorEnabled() bool             { return c.isSnykAdvisorEnabled.Get() }
+func (c *Config) IsSnykAutofixEnabled() bool             { return c.isSnykAutofixEnabled.Get() }
+func (c *Config) LogPath() string                        { return c.logPath }
+func (c *Config) SnykApi() string                        { return c.snykApiUrl }
+func (c *Config) SnykCodeApi() string                    { return c.snykCodeApiUrl }
 func (c *Config) SnykCodeAnalysisTimeout() time.Duration { return c.snykCodeAnalysisTimeout }
 func (c *Config) IntegrationName() string                { return c.integrationName }
 func (c *Config) IntegrationVersion() string             { return c.integrationVersion }
@@ -444,8 +436,6 @@ func (c *Config) ConfigureLogging(level string) {
 		logLevel = envLevel
 	}
 	zerolog.SetGlobalLevel(logLevel)
-	// WIPP
-	zerolog.SetGlobalLevel(zerolog.DebugLevel)
 	zerolog.TimeFieldFormat = time.RFC3339
 
 	if c.logPath != "" {
--- a/application/config/config.go
+++ b/application/config/config.go
@@ -44,15 +44,17 @@ import (
 )
 
 const (
-	deeproxyApiUrlKey     = "DEEPROXY_API_URL"
-	FormatHtml            = "html"
-	FormatMd              = "md"
-	snykCodeTimeoutKey    = "SNYK_CODE_TIMEOUT" // timeout as duration (number + unit), e.g. 10m
-	defaultSnykApiUrl     = "https://snyk.io/api"
-	defaultDeeproxyApiUrl = "https://deeproxy.snyk.io"
-	pathListSeparator     = string(os.PathListSeparator)
-	windows               = "windows"
-	govDomain             = "snykgov.io"
+	deeproxyApiUrlKey       = "DEEPROXY_API_URL"
+	FormatHtml              = "html"
+	FormatMd                = "md"
+	snykCodeTimeoutKey      = "SNYK_CODE_TIMEOUT"    // timeout as duration (number + unit), e.g. 10m
+	autofixEnabledEnvVarKey = "SNYK_AUTOFIX_ENABLED" // "true" or "false", see `defaultAutofixEnabled`
+	defaultAutofixEnabled   = false
+	defaultSnykApiUrl       = "https://snyk.io/api"
+	defaultDeeproxyApiUrl   = "https://deeproxy.snyk.io"
+	pathListSeparator       = string(os.PathListSeparator)
+	windows                 = "windows"
+	govDomain               = "snykgov.io"
 )
 
 var (
@@ -210,7 +212,7 @@ func New() *Config {
 	c.deviceId = c.determineDeviceId()
 	c.addDefaults()
 	c.filterSeverity = lsp.DefaultSeverityFilter()
-	c.isSnykAutofixEnabled.Set(false)
+	c.isSnykAutofixEnabled.Set(getAutofixEnabledFromEnvOrDefault())
 	return c
 }
 
@@ -466,6 +468,8 @@ func (c *Config) SetConfigFile(configFile string) { c.configFile = configFile }
 func getCodeApiUrlFromCustomEndpoint(endpoint string) (string, error) {
 	// Code API endpoint can be set via env variable for debugging using local API instance
 	deeproxyEnvVarUrl := strings.Trim(os.Getenv(deeproxyApiUrlKey), "/")
+	// WIPP
+	log.Logger.Info().Str("deeproxyEnvVarUrl", os.Getenv(deeproxyApiUrlKey))
 	if deeproxyEnvVarUrl != "" {
 		return deeproxyEnvVarUrl, nil
 	}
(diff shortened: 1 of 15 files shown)


infrastructure/snyk_api/fake_api_client.go
```diff
@@ -41,9 +41,7 @@ func (f *FakeApiClient) addCall(params []any, op string) {
 	}
 	calls := f.Calls[op]
 	var opParams []any
-	for p := range params {
-		opParams = append(opParams, params[p])
-	}
+	opParams = append(opParams, params...)
```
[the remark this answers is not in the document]
agronskiy: 😅

domain/snyk/command.go
```diff
@@ -25,6 +25,7 @@ const (
 	NavigateToRangeCommand       = "snyk.navigateToRange"
 	WorkspaceScanCommand         = "snyk.workspace.scan"
 	WorkspaceFolderScanCommand   = "snyk.workspaceFolder.scan"
+	AutofixRunCommand            = "snyk.autofix.run"
```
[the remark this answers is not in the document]
agronskiy: 🦅 👁️ ! Removed!

application/config/config.go
```diff
@@ -161,6 +161,7 @@ type Config struct {
 	runtimeName                  string
 	runtimeVersion               string
 	automaticScanning            bool
+	isSnykAutofixEnabled         concurrency.AtomicBool
```
bastiandoetsch, line 164: Most likely, we should have a capability to have scanners/infrastructure implementations provide their own configuration and merge it automatically to the top level.

infrastructure/code/bundle.go
```diff
@@ -148,3 +155,61 @@ func (b *Bundle) getShardKey(rootPath string, authToken string) string {
...
+	// Polling function just calls the endpoint and registers result, signalling `done` to the
+	// channel.
+	pollFunc := func() {
+		fixSuggestions, fixStatus, err := b.SnykCode.RunAutofix(s.Context(), autofixOptions)
+		if err != nil {
+			log.Error().
+				Err(err).Str("method", method).Str("requestId", b.requestId).
+				Str("stage", "requesting autofix").Msg("error requesting autofix")
+			close(done)
+		} else if fixStatus.message == "COMPLETE" {
```
bastiandoetsch, line 194: this could use `compleStatus` I guess.

application/config/config.go
agronskiy: Yeah that sounds like a neat idea in a longer term indeed. I guess not only autofix is the candidate to go, but the whole bunch of settings e.g. from Code.

infrastructure/code/bundle.go
agronskiy: ✅

agronskiy:
Calling autofix works on top of the resolves now.

application/config/config.go
agronskiy: That would be outside of the scope of this PR, right?

michelkaporin: My point is that it shouldn't be configurable in `config.go`, instead within Code scanner itself, if you want to have it under environment variable.

michelkaporin: Application config should be agnostic to scanner-level configuration, unless it's being exposed as a key to be configured on LS startup / or via user settings in IDE.

agronskiy: Ok, I moved the settings into `code` package as a singletone too, and added tests. Commit: [fe494b0be564f88cbfc03954cac2c0ebf894df66 ](https://github.com/snyk/snyk-ls/pull/234/commits/fe494b0be564f88cbfc03954cac2c0ebf894df66)

Rejected options:
- moving config as a property of the scanner was less preferrable because the codeaction creation happens at the level of bundle, which has no reverse access to the scanner that created this bundle

Questions about tests:
- Currently, I kept the code for resetting settings defaults in `infrastructure/code` tests only in the respective places (where I test them(, but I would generally like some way to do resetting uniformly. I could import `code` from `testutils` and add `resetCodeSettings` in to the `t.Cleanup` in `testutils.UnitTest(t)` and Co. but that's subject to your decision -- because it makes top-level package aware of the bottom-level ones. Not sure you want that :D

infrastructure/code/backend_service.go
```diff
@@ -349,3 +351,110 @@ func (s *SnykCodeHTTPClient) checkResponseCode(r *http.Response) error {
...
+	request := AutofixRequest{
+		Key: AutofixRequestKey{
+			Type:     "file",
+			Hash:     options.bundleHash,
+			FilePath: options.filePath,
+			RuleId:   ruleIdSplit[1],
+			LineNum:  options.issue.Range.Start.Line + 1,
+		},
+		AutofixContext: AutofixContext{
+			Initiatior: "IDE",
```
michelkaporin, line 445: The context here is the same as the `AnalysisContext` for `AnalysisRequest` (see `analysisRequestBody`). We can extract this to a func.

infrastructure/code/backend_service.go
```diff
@@ -349,3 +351,110 @@ func (s *SnykCodeHTTPClient) checkResponseCode(r *http.Response) error {
...
+	return suggestions, AutofixStatus{message: response.Status}, nil
+}
+
+func (s *SnykCodeHTTPClient) autofixRequestBody(options *AutofixOptions) ([]byte, error) {
+	orgName := "unknown"
+	if config.CurrentConfig().GetOrganization() != "" {
+		orgName = config.CurrentConfig().GetOrganization()
+	}
+
+	issueData, ok := options.issue.AdditionalData.(IssueData)
```
michelkaporin, line 424: You should be careful to depend on anything from `snyk.Issue`. Since it's a domain object, any change to it will impact the autofix behaviour. For instance, `AdditionalData` was meant as a view model for UI so far only. If we decide to change our UI, ruleId dependent logic may break here by mistake. 

As a solution I'd suggest to have own mapping between `snyk.Issue` and the Code issue model that stores its own representation that's needed for autofix. E.g. for ruleId, line number, as they come from `options.issue`.

It's more or less ok for now, but you/we should watch out for regressions here.

infrastructure/code/bundle.go
```diff
@@ -149,3 +157,68 @@ func (b *Bundle) getShardKey(rootPath string, authToken string) string {
 
 	return ""
 }
+
+// returns the deferred code action CodeAction which calls autofix.
+func (b *Bundle) createDeferredAutofixCodeAction(ctx context.Context, issue snyk.Issue) *snyk.CodeAction {
+
+	// WIPP: check that all the captures are safe to go.
```
michelkaporin, line 164: Still WIP?

infrastructure/code/settings.go
```diff
@@ -17,11 +17,68 @@ package code
...
 
+const (
+	// String "true" or "false", see `defaultAutofixEnabled`
+	autofixEnabledEnvVarKey = "SNYK_AUTOFIX_ENABLED"
+	// By default this is disabled
+	defaultAutofixEnabled = false
+)
+
+var (
+	codeSettingsSingletone      *codeSettings
```
michelkaporin, line 37: Sorry for being grammar nazi - `Singleton`. Singletone sounds either Italian or French, not sure yet 😄

michelkaporin, line 37: Any reason to have this on a "global" package level? Why not to make it part of the code scanner singleton?

infrastructure/code/types.go
```diff
@@ -59,3 +59,46 @@ type MarkerPosition struct {
...
+
+type autofixResponseSuggestion = string
+
+type AutofixRequestKey struct {
+	Type     string `json:"type"`
+	Hash     string `json:"hash"`
+	Shard    string `json:"shard"`
+	FilePath string `json:"filePath"`
+	RuleId   string `json:"ruleId"`
+	LineNum  int    `json:"lineNum"` // 1-based
```
michelkaporin, line 78: Isn't it a problem/confusing that it's 1-based on backend, if e.g. analysis results are 0-based on the Code Analysis backend?

infrastructure/code/convert.go
```diff
@@ -33,6 +34,10 @@ import (
 	"github.com/snyk/snyk-ls/internal/util"
 )
 
+const (
+	AutofixCodeActionTitle = "Run LSP Autofix (I'm feeling lucky!)"
```
michelkaporin, line 38: This isn't used any more, right?

michelkaporin (changes requested):

application/config/config.go
```diff
@@ -455,6 +455,8 @@ func (c *Config) SetConfigFile(configFile string) { c.configFile = configFile }
 func getCodeApiUrlFromCustomEndpoint(endpoint string) (string, error) {
 	// Code API endpoint can be set via env variable for debugging using local API instance
 	deeproxyEnvVarUrl := strings.Trim(os.Getenv(deeproxyApiUrlKey), "/")
+	// WIPP
+	log.Logger.Info().Str("deeproxyEnvVarUrl", os.Getenv(deeproxyApiUrlKey))
```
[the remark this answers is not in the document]
agronskiy: ✅

domain/snyk/autofix_suggestion.go
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package snyk
+
+// AutofixSuggestion models a fix returned by autofix service
+type AutofixSuggestion struct {
```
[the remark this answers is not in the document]
agronskiy: Very good catch, thanks, moved there

infrastructure/code/backend_service.go
```diff
@@ -349,3 +351,110 @@ func (s *SnykCodeHTTPClient) checkResponseCode(r *http.Response) error {
...
+	requestBody, err := s.autofixRequestBody(&options)
+	if err != nil {
+		log.Err(err).Str("method", method).Str("requestBody", string(requestBody)).Msg("error creating request body")
+		return nil, AutofixStatus{}, err
+	}
+
+	responseBody, err := s.doCall(span.Context(), "POST", "/autofix/suggestions", requestBody)
+	failed := AutofixStatus{message: "FAILED"}
+	if err != nil {
+		log.Err(err).Str("method", method).Str("responseBody", string(responseBody)).Msg("error response from analysis")
```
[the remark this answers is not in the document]
agronskiy: ✅ 🤦

infrastructure/code/types.go
agronskiy: I see your point, actually I was hesitating back and forth if to implement your comment. I confirm that the VScode and LSP are using 0-based ones. I double checked this
> analysis results are 0-based on the Code Analysis backend
and found that it is still 1-based in our API https://github.com/snyk/snyk-ls/blob/7667009db2846f270ab1b4fac2278133fc408836/infrastructure/code/convert.go#L283

This seems to follow the SARIF standard (https://docs.oasis-open.org/sarif/sarif/v2.1.0/sarif-v2.1.0.html#_Ref493492556) which says
> The line number of the first line in a text artifact SHALL be 1

So here I vote to keep conistent with the backend API, I'll add links to comments and make convertation more prominent.

infrastructure/code/settings.go
agronskiy: ✅

infrastructure/code/bundle.go
agronskiy: ✅

infrastructure/code/backend_service.go
agronskiy: https://github.com/snyk/snyk-ls/pull/234#discussion_r1131187779 
> moving config as a property of the scanner was less preferrable because the codeaction creation happens at the level of bundle, which has no reverse access to the scanner that created this bundle

maybe I'm wrong here, you mention that scanner is the singleton but I AFAIU accesing it goes via `snyk` package, then going to its product scanners, filtering them to find the code one etc -- just to arrive back to the code.

As opposed to that chain, keeping it just a config singleton at a package level allows to access it now (and in future) without overhead. Do I oversimplify or miss smth?

infrastructure/code/settings.go
agronskiy: https://github.com/snyk/snyk-ls/pull/234#discussion_r1131187779 
> moving config as a property of the scanner was less preferrable because the codeaction creation happens at the level of bundle, which has no reverse access to the scanner that created this bundle

maybe I'm wrong here, you mention that scanner is the singleton but I AFAIU accesing it goes via `snyk` package, then going to its product scanners, filtering them to find the code one etc -- just to arrive back to the code.

As opposed to that chain, keeping it just a config singleton at a package level allows to access it now (and in future) without overhead. Do I oversimplify or miss smth?

infrastructure/code/backend_service.go
agronskiy: ✅

infrastructure/code/backend_service.go
agronskiy: I agree with the direction, although detaching the information stored in `Issue` from sounds like not a fully polished solution -- because it opens path to maintaining this map in sync with which issues are not dead (went out of scope) etc. 

You anyways already started with `any` there, so what I would do is (similar to scanners) to have an array of `[]any` in it, and each backend is then free to add its own struct:
```go
import "fmt"

type Issue struct {
	additionalData []any
}

type codeData struct {
	myData string
}

type ossData struct {
	myData int
}

func main() {
	issue := Issue{
		additionalData: []any{},
	}

	// code part creating issue
	issue.additionalData = append(issue.additionalData, codeData{myData: "code"})
	// oss part
	issue.additionalData = append(issue.additionalData, ossData{myData: 42})

	// part consuming the issue
	for _, data := range issue.additionalData {
		codeData, ok := data.(codeData)
		if !ok {
			continue
		}
		fmt.Print("Code data ", codeData.myData)
	}
}
```

All in all, this seems more conformant to the story of having product-specific things on top level (like scanners).

infrastructure/code/bundle.go
```diff
@@ -115,10 +115,18 @@ func (b *Bundle) retrieveAnalysis(ctx context.Context) ([]snyk.Issue, error) {
...
 		}
 
-		if status.message == "COMPLETE" {
+		if status.message == completeStatus {
 			logger.Trace().Str("requestId", b.requestId).
 				Msg("sending diagnostics...")
 			p.End("Analysis complete.")
+
+			if getCodeSettings().isAutofixEnabled.Get() {
+				// TODO(alex.gronskiy): logging
```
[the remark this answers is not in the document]
agronskiy: ✅

infrastructure/code/types.go
agronskiy: ✅

infrastructure/code/convert.go
agronskiy: ✅

infrastructure/code/types.go
bastiandoetsch: It's a mess: two standards with different bases. Sigh.

infrastructure/code/backend_service.go
```diff
@@ -349,3 +345,102 @@ func (s *SnykCodeHTTPClient) checkResponseCode(r *http.Response) error {
...
+	method := "code.RunAutofix"
+	span := s.instrumentor.StartSpan(ctx, method)
+	defer s.instrumentor.Finish(span)
+
+	requestId, err := performance2.GetTraceId(span.Context())
+	if err != nil {
+		log.Err(err).Str("method", method).Msg("Failed to obtain request id. " + err.Error())
+		return nil, AutofixStatus{}, err
+	}
+	log.Debug().Str("method", method).Str("requestId", requestId).Msg("API: Retrieving analysis for bundle")
```
bastiandoetsch, line 369: maybe the comment should mention autofix :)

infrastructure/code/backend_service.go
```diff
@@ -349,3 +345,102 @@ func (s *SnykCodeHTTPClient) checkResponseCode(r *http.Response) error {
...
+	err = json.Unmarshal(responseBody, &response)
+	if err != nil {
+		log.Err(err).Str("method", method).Str("responseBody", string(responseBody)).Msg("error unmarshalling")
+		return nil, failed, err
+	}
+
+	log.Debug().Str("method", method).Str("requestId", requestId).Msgf("Status: %s", response.Status)
+
+	if response.Status == failed.message {
+		log.Err(err).Str("method", method).Str("responseStatus", response.Status).Msg("analysis failed")
```
bastiandoetsch, line 395: just go over the comments once more - will be helpful in error analysis. Not blocking approval though.

bastiandoetsch (approved):

infrastructure/code/autofix_suggestion.go
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package code
+
+import (
+	"github.com/snyk/snyk-ls/domain/snyk"
+)
+
+// AutofixSuggestion models a fix returned by autofix service
+type AutofixSuggestion struct {
```
michelkaporin, line 24: Should this belong to `types.go`?

infrastructure/code/types.go
michelkaporin: Nice find, @agronskiy 👍

michelkaporin (commented):
Left few more comms, can't wait it to be merged ❤️ 

How about we update CODEOWNERS file here as well? Or would you rather trust us in making changes to this without your approvals for now? Perhaps if you extract autofix logic to completely separate files, that'd be easier to mark ownership. Hammerhead could still continue owning the basic Code API wrapper implementation.

cc @snyk/hammerhead @snyk/zenith

infrastructure/code/fake_snyk_code_api_service.go
```diff
@@ -101,15 +106,21 @@ func TempWorkdirWithVulnerabilities(t *testing.T) (filePath string, path string)
...
+	ExtendedBundleCount          int
+	AnalysisDuration             time.Duration
+	FailOnCreateBundle           bool
+	currentConcurrentScans       int
+	maxConcurrentScans           int
+	AutofixDuration              time.Duration
+	currentConcurrentAutofixRuns int
+	maxConcurrentAutofixRuns     int
+
+	// Test constants
```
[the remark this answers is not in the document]
agronskiy: ✅

infrastructure/code/sarif_types.go
```diff
@@ -183,23 +183,36 @@ type AnalysisRequestKey struct {
...
-	Initiator string             `json:"initiator"`
-	Flow      string             `json:"flow,omitempty"`
-	Org       AnalysisContextOrg `json:"org,omitempty"`
+type codeRequestContext struct {
+	Initiator string                `json:"initiator"`
+	Flow      string                `json:"flow,omitempty"`
+	Org       codeRequestContextOrg `json:"org,omitempty"`
+}
+
+func newCodeRequestContext(orgName string) codeRequestContext {
```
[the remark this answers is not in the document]
agronskiy: Moved to sarif_utils.go

infrastructure/code/autofix_suggestion.go
agronskiy: ✅

infrastructure/code/backend_service.go
agronskiy: ✅

infrastructure/code/settings.go
```diff
@@ -17,11 +17,95 @@ package code
...
+func extractAutofixLanguagesFromEnvOrDefault(m *concurrency.AtomicMap) {
+	env, ok := os.LookupEnv(autofixLanguagesEnvVarKey)
+	var languagesString string
+	if !ok {
+		languagesString = defaultAutofixLanguages
+	} else {
+		languagesString = env
+	}
+
+	languagesStringRegexp := regexp.MustCompile("[a-z]+(,[a-z])*")
```
bastiandoetsch, line 98: shouldn't this be `.js, .java` etc? In this case, I think the regexp must be changed.

infrastructure/code/settings.go
```diff
@@ -17,11 +17,95 @@ package code
...
+	}
+
+	parseBool, err := strconv.ParseBool(env)
+	if err != nil {
+		return defaultAutofixEnabled
+	}
+	return parseBool
+}
+
+func extractAutofixLanguagesFromEnvOrDefault(m *concurrency.AtomicMap) {
```
bastiandoetsch, line 89: This is very C-style that you pass in a pointer to a map and then manipulate it. Would be better to return a map with the languages.

infrastructure/code/settings.go
agronskiy: @vraychev has an opened PR on sast side which adds additional payload under `/filters` and renders my env var effort unneeded -- I deleted it.

Other than that no, in this case the languages are indeed languages (as in`javascritp`) -- what is stored by the `RunAnalysis` in the `snyk.Issue.AdditionalData` is language model. There's one to many mapping from `language` to `file extenstion`, e.g. engine considers `.ts` as `javascript` for the purposes of analysis. Same for autofix.

infrastructure/code/settings.go
```diff
@@ -17,11 +17,68 @@ package code
...
+const (
+	// String "true" or "false", see `defaultAutofixEnabled`
+	autofixEnabledEnvVarKey = "SNYK_AUTOFIX_ENABLED"
+	// By default this is disabled
+	defaultAutofixEnabled = false
+)
+
+var (
+	codeSettingsSingleton      *codeSettings
+	codeSettingsSingletonMutex = &sync.Mutex{}
```
bastiandoetsch, line 38: not a blocker, but maybe better to put into the `codeSettings` type.

infrastructure/code/settings.go
```diff
@@ -17,11 +17,68 @@ package code
...
+// Create new code settings
+func newCodeSettings() *codeSettings {
+	s := &codeSettings{}
+	s.isAutofixEnabled.Set(getAutofixEnabledFromEnvOrDefault())
+	return s
+}
+
+// Gets the codeSettings singletone, lazily constructing it on the fly at the first call
+func getCodeSettings() *codeSettings {
+	if codeSettingsSingleton == nil {
```
bastiandoetsch, line 54: you probably should use the mutex to avoid races here

infrastructure/code/settings.go
```diff
@@ -17,11 +17,68 @@ package code
...
+func resetCodeSettings() {
+	codeSettingsSingletonMutex.Lock()
+	defer codeSettingsSingletonMutex.Unlock()
+	codeSettingsSingleton = newCodeSettings()
+}
+
+// Attempts to read the `autofixEnabledEnvVarKey` env variable or sets the
+// bool to default
+func getAutofixEnabledFromEnvOrDefault() bool {
+	env := os.Getenv(autofixEnabledEnvVarKey)
```
bastiandoetsch, line 70: you probably should use the mutex to avoid races here

michelkaporin (approved):

PR #236, opened 2023-02-17
fix: gitignore regression

bastiandoetsch:
### Description

Using relative paths in gitignore logic led to ignore `.`. Until we find the right fix, this uses absolute paths now.

### Checklist

- [x] Linted

### Screenshots / GIFs

### DIFF
--- a/infrastructure/code/code_test.go
+++ b/infrastructure/code/code_test.go
@@ -276,7 +276,7 @@ func Test_IgnoresWithNegationInSnykCode(t *testing.T) {
 
 	calls := codeClientMock.GetAllCalls("extendBundleWithSource")
 	assert.Len(t, calls, 1)
-	assert.Contains(t, scanner.ignorePatterns, "!**/temp")
+	assert.Contains(t, scanner.ignorePatterns, "!"+filepath.ToSlash(repobase+"/**/temp"))
 }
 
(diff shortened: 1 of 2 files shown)


infrastructure/code/code_test.go
```diff
@@ -394,16 +394,14 @@ func Test_LoadIgnorePatternsAndCountFiles_RelativePathIgnores(t *testing.T) {
...
+	expectedSubDirPattern := filepath.ToSlash(filepath.Join(subDir, "**/*"))
 
 	sc := Scanner{}
-	_, err = sc.loadIgnorePatternsAndCountFiles(tempDir)
+	_, err := sc.loadIgnorePatternsAndCountFiles(tempDir)
 
 	assert.NoError(t, err)
 	assert.Contains(t, sc.ignorePatterns, expectedSubDirPattern)
-	assert.Len(t, sc.ignorePatterns, len(getDefaultIgnorePatterns())+1)
+	assert.Len(t, sc.ignorePatterns, len(getDefaultIgnorePatterns())+2)
```
michelkaporin, line 404: Why is it +2 now?

bastiandoetsch: Because the parsing into globs can return more than one glob - and it now does in this case.

michelkaporin (approved):

PR #253, opened 2023-03-15
fix: panics on unrelated code actions [HEAD-205]

asaf92:
### Description

Fixed panics when attempting to resolve non-deferred code actions


bastiandoetsch (approved):
Ship it :)

PR #254, opened 2023-03-16
fix: IaC scan message issues not unique [HEAD-70]

michelkaporin:
### Description

### Checklist

- [ ] Linted

### Screenshots / GIFs

### DIFF
--- a/application/server/notification/scan_notifier.go
+++ b/application/server/notification/scan_notifier.go
@@ -96,7 +96,7 @@ func (n *scanNotifier) appendIacIssues(scanIssues []lsp.ScanIssue, folderPath st
 		}
 
 		scanIssues = append(scanIssues, lsp.ScanIssue{
-			Id:       issue.ID,
+			Id:       additionalData.Key,
 			Title:    additionalData.Title,
 			Severity: issue.Severity.String(),
 			FilePath: issue.AffectedFilePath,
@@ -158,7 +158,7 @@ func (n *scanNotifier) appendCodeIssues(scanIssues []lsp.ScanIssue, folderPath s
 		}
 
 		scanIssues = append(scanIssues, lsp.ScanIssue{
-			Id:       issue.ID,
+			Id:       additionalData.Key,
 			Title:    issue.Message,
 			Severity: issue.Severity.String(),
 			FilePath: issue.AffectedFilePath,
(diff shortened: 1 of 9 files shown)


infrastructure/code/convert.go
```diff
@@ -366,7 +367,7 @@ func (s *SarifResponse) toIssues() (issues []snyk.Issue) {
 	return issues
 }
 
-func getIssueId(ruleId string, path string, startLine int, endLine int, startCol int, endCol int) string {
+func getIssueKey(ruleId string, path string, startLine int, endLine int, startCol int, endCol int) string {
 	// deepcode ignore InsecureHash: The hash isn't used for security purposes.
 	id := md5.Sum([]byte(ruleId + path + strconv.Itoa(startLine) + strconv.Itoa(endLine) + strconv.Itoa(startCol) + strconv.Itoa(endCol)))
```
bastiandoetsch, line 372: Using md5 is still a smell :). I don't see any reason to not use a secure hash like sha-256, as performance is comparable.

infrastructure/iac/iac.go
```diff
@@ -408,3 +414,9 @@ func (iac *Scanner) toIssueSeverity(snykSeverity string) snyk.Severity {
 	}
 	return severity
 }
+
+func getIssueKey(affectedFilePath string, issue iacIssue) string {
+	// deepcode ignore InsecureHash: The hash isn't used for security purposes.
+	id := md5.Sum([]byte(affectedFilePath + strconv.Itoa(issue.LineNumber) + issue.PublicID))
```
bastiandoetsch, line 420: Please don't use md5. Just use sha-256. These old hash algorithms are just not needed, and the comment would be completely unnecessary.

application/server/notification/scan_notifier.go
```diff
@@ -158,7 +158,7 @@ func (n *scanNotifier) appendCodeIssues(scanIssues []lsp.ScanIssue, folderPath s
 		}
 
 		scanIssues = append(scanIssues, lsp.ScanIssue{
-			Id:       issue.ID,
+			Id:       additionalData.Key,
```
asaf92, line 161: Why are we changing code issues as well?

bastiandoetsch (approved):
:)

application/server/notification/scan_notifier.go
michelkaporin: See the PR description:
> We've assumed that it's not surfaced in UI and changed in [here](https://github.com/snyk/snyk-ls/pull/216/files#diff-c487190192b10556a66dd6213400d3084215c69eb2c4f547285fe7b4ab61c0f4), however these are reported in problem views as issue IDs (see screenshot attached).

PR #256, opened 2023-03-22
chore: add initial oauth authentication wrapper to call GAF

bastiandoetsch:
### Description

Not used yet, but adds the functionality to call GAF via a language server auth provider.

### Checklist

- [x] Linted

### Screenshots / GIFs


application/config/config.go
```diff
@@ -339,6 +344,9 @@ func (c *Config) UpdateApiEndpoints(snykApiUrl string) bool {
 		}
 
 		c.setSnykCodeApi(snykCodeApiUrl)
+		if strings.Contains(snykCodeApiUrl, "fedramp") {
```
PeterSchafer, line 347: probably better snykgov.io

application/config/config_test.go
```diff
@@ -168,6 +169,13 @@ func TestSnykCodeApi(t *testing.T) {
 		codeApiEndpoint, _ := getCodeApiUrlFromCustomEndpoint("")
 		assert.Equal(t, customDeeproxyUrl, codeApiEndpoint)
 	})
+
+	t.Run("fedramp substring endpoint enables oauth authentication", func(t *testing.T) {
+		endpoint := "https://app.fedramp,snyk.io/api/v1"
```
PeterSchafer, line 174: snykgov.io and typo

application/di/init.go
```diff
@@ -105,6 +109,7 @@ func initInfrastructure() {
 		},
 	)
 
+	engine = app.CreateAppEngine()
```
PeterSchafer, line 112: init missing and maybe re-use common method here and in TestInit()

application/server/configuration_test.go
```diff
@@ -420,4 +420,16 @@ func Test_InitializeSettings(t *testing.T) {
 
 		assert.Equal(t, true, c.IsSnykCodeQualityEnabled())
 	})
+
+	t.Run("authenticationMethod is passed", func(t *testing.T) {
+		c := config.New()
+		config.SetCurrentConfig(c)
+		di.TestInit(t)
+		authenticationMethod := lsp.OAuthAuthentication
+
```
PeterSchafer, line 429: please check the pre-condition

infrastructure/oauth/oauth_provider.go
```diff
@@ -0,0 +1,74 @@
...
+
+func (p *oAuthProvider) ClearAuthentication(ctx context.Context) error {
+	p.config.Set(auth.CONFIG_KEY_OAUTH_TOKEN, "")
+	p.config.Set(configuration.AUTHENTICATION_TOKEN, "")
+	p.config.Set(configuration.AUTHENTICATION_BEARER_TOKEN, "")
+	return nil
+}
+
+func (p *oAuthProvider) AuthURL(ctx context.Context) string {
+	// TODO: this is a hack, we should be able to get the auth url from the authenticator
```
PeterSchafer, line 59: return p.auythURL

infrastructure/oauth/oauth_provider.go
```diff
@@ -0,0 +1,74 @@
...
+	return tokenUrl
+}
+
+func (p *oAuthProvider) AuthenticateToken(token string) error {
+	var oauthToken oauth2.Token
+	err := json.Unmarshal([]byte(token), &oauthToken)
+	if err != nil {
+		return err
+	}
+	if oauthToken.Expiry.Before(time.Now()) {
```
PeterSchafer, line 70: Please use https://pkg.go.dev/golang.org/x/oauth2#Token.Valid

infrastructure/oauth/oauth_provider_test.go
```diff
@@ -0,0 +1,171 @@
...
+	config.Set(configuration.WEB_APP_URL, "https://app.fake.snyk.io")
+	authenticator := NewFakeOauthAuthenticator(time.Now().Add(10*time.Second), true, config).(*fakeOauthAuthenticator)
+	provider := NewOAuthProvider(config, authenticator)
+
+	url := provider.AuthURL(context.Background())
+
+	assert.NotEmpty(t, url)
+	_, err := url2.Parse(url)
+	assert.NoError(t, err)
+	assert.Equal(t, url, "https://id.fake.snyk.io/oauth2/default/v1/token")
```
PeterSchafer, line 170: too specific, maybe just check that the url is parseable

application/config/config.go
```diff
@@ -201,11 +203,12 @@ func New() *Config {
 	c.snykCodeAnalysisTimeout = snykCodeAnalysisTimeoutFromEnv()
 	c.token = ""
 	c.trustedFoldersFeatureEnabled = true
+	c.automaticScanning = true
```
PeterSchafer, line 206: how is `automaticScanning` related?

infrastructure/oss/maven_range_finder.go
```diff
@@ -30,13 +29,11 @@ type mavenRangeFinder struct {
 }
 
 func (m *mavenRangeFinder) find(issue ossIssue) snyk.Range {
-	searchPackage, version := introducingPackageAndVersion(issue)
-	log.Trace().Interface("issue", issue).Str("searchPackage", searchPackage).Str("searchVersion", version)
```
PeterSchafer, line 34: why was this removed? noise?

infrastructure/oss/oss.go
```diff
@@ -238,10 +238,8 @@ func (oss *Scanner) unmarshallAndRetrieveAnalysis(ctx context.Context, res []byt
 		targetFileUri := uri.PathToUri(targetFilePath)
 		fileContent, err := os.ReadFile(targetFilePath)
 		if err != nil {
-			log.Err(err).Str("method", "unmarshallAndRetrieveAnalysis").
-				Msgf("Error while reading the file %v, err: %v", targetFile, err)
-			oss.errorReporter.CaptureErrorAndReportAsIssue(path, err)
-			return nil
+			// don't fail the scan if we can't read the file. No annotations with ranges, though.
+			fileContent = []byte{}
```
PeterSchafer, line 242: these changes seem unrelated. What is the reason for them?

application/config/config.go
bastiandoetsch: it's not. was there before.

infrastructure/oss/oss.go
bastiandoetsch: True. LS was erroring when not having a defined targetFile, e.g. when scanning unmanaged. this should make it a bit more resilient - instead of erroring out, it now just skips range finding.

PeterSchafer: do we have any tests that could cover this change?

bastiandoetsch: No yet, but I can write a test

application/server/configuration.go
```diff
@@ -127,6 +132,29 @@ func writeSettings(settings lsp.Settings, initialize bool) {
...
+}
+
+func updateAuthenticationMethod(settings lsp.Settings) {
+	c := config.CurrentConfig()
+	c.SetAuthenticationMethod(settings.AuthenticationMethod)
+	if config.CurrentConfig().GetAuthenticationMethod() == lsp.OAuthAuthentication {
+		engine := di.Engine()
+		c := engine.GetConfiguration()
+		c.Set(configuration.OAUTH_AUTH_ENABLED, true)
+		c.Set(configuration.API_URL, config.CurrentConfig().SnykApi())
```
PeterSchafer, line 145: it might make sense to do this somewhere else, since we need this with and without oauth

bastiandoetsch: yep, will come with config sync

application/di/init.go
```diff
@@ -93,27 +97,39 @@ func initDomain() {
...
+				"/usr/java",
+				"/opt",
+				"/Library",
+			})
+	}
 
+	initializeWorkflowEngine()
 	errorReporter = sentry2.NewSentryErrorReporter()
-	installer = install.NewInstaller(errorReporter)
+	installer = install.NewInstaller(errorReporter, engine.GetNetworkAccess().GetUnauthorizedHttpClient)
```
PeterSchafer, line 121: Just FYI, I think it would it would have also been fine to use the GetHttpClient() here.

internal/httpclient/httpclient.go
```diff
@@ -17,42 +17,18 @@
...
-		log.Err(err).Str("method", method).Send()
-	}
-	if proxy != nil {
-		proxySplit := strings.Split(proxy.String(), "@")
-		proxyLogString := proxySplit[0]
-		if len(proxySplit) > 1 {
-			proxyLogString = "xxx@" + proxySplit[1]
-		}
-		log.Info().Str("method", method).Str("proxy", proxyLogString).Msg("created http client with proxy support")
+func NewHTTPClient(access networking.NetworkAccess) *http.Client {
```
PeterSchafer, line 25: Question: these functions appear to be used only once. Shall we maybe just get rid of them?

bastiandoetsch: let's do that in the next PR though. this one is big enough.

PR #258, opened 2023-03-23
fix: URL encoding [HEAD-211][HEAD-212]

asaf92:
### Description

Fix URL Encoding in `limitToFiles` for Code API calls


infrastructure/code/code.go
```diff
@@ -488,14 +479,29 @@ func (sc *Scanner) createBundle(
 		if !(len(fileContent) > 0 && len(fileContent) <= maxFileSize) {
 			continue
 		}
-		file := getFileFrom(filePath, fileContent)
-		bundleFiles[filePath] = file
-		fileHashes[filePath] = file.Hash
+		bundleFile := getFileFrom(filePath, fileContent)
+		bundleFiles[filePath] = bundleFile
+		fileHashes[filePath] = bundleFile.Hash
+		fileName := filepath.Base(filePath)
```
bastiandoetsch, line 485: this is most likely not correct - it should be the path from folder root - e.g. if you have several `index.js` it wouldn't work, if I understand correctly.

infrastructure/code/bundle.go
```diff
@@ -36,6 +36,7 @@ type Bundle struct {
 	SnykCode      SnykCodeClient
 	BundleHash    string
 	UploadBatches []*UploadBatch
+	Files         map[string]BundleFile
```
bastiandoetsch, line 39: why do we need to have a copy of the files here?

infrastructure/code/code_test.go
```diff
@@ -75,11 +75,16 @@ func TestCreateBundle(t *testing.T) {
...
+				"testRequestId",
+				dir,
+				[]string{file},
+				map[string]bool{})
 			if err != nil {
 				t.Fatal(err)
 			}
-			assert.Len(t, missingFiles, 1, "bundle should have 1 missing files")
+			bundleFiles := bundle.Files
+			assert.Len(t, bundleFiles, 1, "bundle should have 1 bundle files")
```
bastiandoetsch, line 87: missing files != bundle files

infrastructure/code/code_test.go
```diff
@@ -92,11 +97,16 @@ func TestCreateBundle(t *testing.T) {
...
+				"testRequestId",
+				dir,
+				[]string{file},
+				map[string]bool{})
 			if err != nil {
 				t.Fatal(err)
 			}
-			assert.Len(t, missingFiles, 0, "bundle should not have missing files")
+			bundleFiles := bundle.Files
+			assert.Len(t, bundleFiles, 0, "bundle should not have bundle files")
```
bastiandoetsch, line 109: same here - missing files is what it is about - not bundle files

infrastructure/code/code_test.go
asaf92: It's supposed to be "bundle files". The missing files are coming from a component that is not under test here (the snyk code client is mocked). It always tested for bundle files, nothing was changed logic-wise

PR #260, opened 2023-03-26
fix: send Snyk Code config files [HEAD-213]

asaf92:
### Description

Fixed Snyk Code not sending config files to the backend

### Checklist


infrastructure/code/code_test.go
```diff
@@ -163,6 +163,31 @@ func TestCreateBundle(t *testing.T) {
 			assert.Len(t, snykCodeMock.GetAllCalls(CreateBundleOperation), 0, "bundle shouldn't have called createBundle")
 		},
 	)
+
+	t.Run("includes config files", func(t *testing.T) {
```
bastiandoetsch, line 167: Would be good to have an integration test that includes `.gitignore` as config file to make sure that the ignore doesn't mess with the upload of config files.

michelkaporin (approved):

PR #265, opened 2023-03-28
feat(autofix): support extensions endpoint [ML-185]

agronskiy:
Counterpart of the PR on the sast side which adds the filtering and prohibits the non-supported autofix extensions to have codeactions.

Side-effects:
- added `EncodePath(...)` as a followup to #267 

@bastiandoetsch, @michelkaporin, @asaf92 would you like to pick up from there and add pact testing etc? (not an expert)

### DIFF
--- a/infrastructure/code/backend_service.go
+++ b/infrastructure/code/backend_service.go
@@ -78,8 +78,9 @@ type extendBundleRequest struct {
 }
 
 type filtersResponse struct {
-	ConfigFiles []string `json:"configFiles" pact:"min=1"`
-	Extensions  []string `json:"extensions" pact:"min=1"`
+	ConfigFiles       []string `json:"configFiles" pact:"min=1"`
+	Extensions        []string `json:"extensions" pact:"min=1"`
+	AutofixExtensions []string `json:"autofixExtensions" pact:"min=1"`
 }
 
 func NewHTTPRepository(
@@ -90,7 +91,12 @@ func NewHTTPRepository(
 	return &SnykCodeHTTPClient{client, instrumentor, errorReporter}
 }
 
-func (s *SnykCodeHTTPClient) GetFilters(ctx context.Context) (configFiles []string, extensions []string, err error) {
+func (s *SnykCodeHTTPClient) GetFilters(ctx context.Context) (
+	configFiles []string,
+	extensions []string,
+	autofixExtensions []string,
+	err error,
+) {
 	method := "code.GetFilters"
 	log.Debug().Str("method", method).Msg("API: Getting file extension filters")
 
@@ -99,16 +105,16 @@ func (s *SnykCodeHTTPClient) GetFilters(ctx context.Context) (configFiles []stri
 
 	responseBody, err := s.doCall(span.Context(), "GET", "/filters", nil)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, nil, err
 	}
 
 	var filters filtersResponse
 	err = json.Unmarshal(responseBody, &filters)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, nil, err
 	}
 	log.Debug().Str("method", method).Msg("API: Finished getting filters")
-	return filters.ConfigFiles, filters.Extensions, nil
+	return filters.ConfigFiles, filters.Extensions, filters.AutofixExtensions, nil
 }
 
 func (s *SnykCodeHTTPClient) CreateBundle(
--- a/infrastructure/code/backend_service.go
+++ b/infrastructure/code/backend_service.go
@@ -92,9 +92,7 @@ func NewHTTPRepository(
 }
 
 func (s *SnykCodeHTTPClient) GetFilters(ctx context.Context) (
-	configFiles []string,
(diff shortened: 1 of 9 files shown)


infrastructure/code/backend_service.go
```diff
@@ -90,7 +91,12 @@ func NewHTTPRepository(
 	return &SnykCodeHTTPClient{client, instrumentor, errorReporter}
 }
 
-func (s *SnykCodeHTTPClient) GetFilters(ctx context.Context) (configFiles []string, extensions []string, err error) {
+func (s *SnykCodeHTTPClient) GetFilters(ctx context.Context) (
+	configFiles []string,
```
michelkaporin, line 95: You can replace multiple return types with a single `filterResponse` type to have cleaner interface.

infrastructure/code/bundle.go
```diff
@@ -94,6 +95,27 @@ func getIssueLangAndRuleId(issue snyk.Issue) (string, string, bool) {
 	return ruleIdSplit[0], ruleIdSplit[1], true
 }
 
+// isAutofixSupportedForExtension shall call `GetFilters` if necessary and cache the results
+// in the `codeSettings` singleton.
+func (b *Bundle) isAutofixSupportedForExtension(ctx context.Context, file string) bool {
+	if getCodeSettings().autofixExtensions == nil {
+		// query
+		_, _, autofixExts, err := b.SnykCode.GetFilters(ctx)
```
michelkaporin, line 103: We call and cache `GetFilters` in [BundleUploader](https://github.com/snyk/snyk-ls/blob/6d2a941c53e0effd4d5cd1c9e7f43143c506000f/infrastructure/code/bundle_uploader.go#L119) already. Can we save a network call and unify caching between there and here?

We can introduce something like "Filters" type shared for the whole scanner.

agronskiy: I considered that, but 
- the autofix supported extensions logically do not belong to the bundle uploading. I understand why `supportedExtension` are needed to bundle uploader at the bundle, but the autofix supported extensions come into play later so they are called at most once at the codeaction creation time (and if the autofix is enabled)
- to address your comment I could have made autofix caching at the call site by the `bundleUploader.isSupporte` -- but I would cosider that logically much less sound (why does `isSupported` do anything for autofix caching?)
- the remaining argument about saving a call: I don't think that one literally one call (which is cached on the side of deeproxy) while there are orders of magnitude more calls when e.g. polling a long analysis results is something to worry about

infrastructure/code/bundle.go
```diff
@@ -194,11 +221,12 @@ func (b *Bundle) createDeferredAutofixCodeAction(ctx context.Context, issue snyk
 				Msg("error converting to relative file path")
 			return nil
 		}
+		encodedRelativePath := EncodePath(relativePath)
```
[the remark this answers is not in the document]
agronskiy: 🤗

infrastructure/code/backend_service.go
agronskiy: ✅ Done, very nice point! I used you suggestion to replace it with `filterResponse` -- let me know if you would like to have a mirroring struct not in the `http` world.

infrastructure/code/bundle.go
michelkaporin: Agree 👍 There's no harm in issuing this additional call. It can be little confusing why this state isn't shared already, if it's the same endpoint architecturally. But it's really a minor thing and pragmatically not worth of investment here, just wanted to flag :)

infrastructure/code/backend_service.go
michelkaporin: No need for a mirroring struct, we won't win from it, let's keep it simple.

infrastructure/code/bundle.go
```diff
@@ -94,6 +95,28 @@ func getIssueLangAndRuleId(issue snyk.Issue) (string, string, bool) {
 	return ruleIdSplit[0], ruleIdSplit[1], true
 }
 
+// isAutofixSupportedForExtension shall call `GetFilters` if necessary and cache the results
+// in the `codeSettings` singleton.
+func (b *Bundle) isAutofixSupportedForExtension(ctx context.Context, file string) bool {
```
bastiandoetsch, line 100: Non-blocking question: Is bundle really the right place to have this function? Would it maybe be better located as callable function in a different code-related type?

bastiandoetsch (approved):
Looks good to me, @michelkaporin any objections to merge?

michelkaporin (approved):
Lovely! Thank you, @agronskiy 🙌

infrastructure/code/bundle.go
agronskiy: Like the autofix-related things gathered into some receiver? That would be good, naturally poses a bit of challenge to me since I need bunch of things like `SnykCodeHttpClient` whose lifecycle I can't precisely reason about, and also access to some of `bundle`s information such as `rootPath` -- and bundle's even less clear regarding their lifecycle...

But I'm more than happy to have a session about the best direction of making further changes, to not mess this part of the codebase.

PR #271, opened 2023-03-31
fix: ignore with several work dirs [HEAD-234]

Avishagp:
### Description

This PR fixes ignore files not being respected in different work directories. 

### Checklist

- [X] Tests added and all succeed

### DIFF
--- a/infrastructure/code/code_test.go
+++ b/infrastructure/code/code_test.go
@@ -364,42 +364,36 @@ func Test_GetWorkspaceFiles_SkipIgnoredDirs(t *testing.T) {
 func Test_FilesRespectIgnorePatterns(t *testing.T) {
 	tempDir := t.TempDir()
 	pathA := filepath.Join(tempDir, "A")
-	err := os.Mkdir(pathA, 0755)
-	assert.Nil(t, err)
 	pathB := filepath.Join(tempDir, "B")
-	err = os.Mkdir(pathB, 0755)
-	assert.Nil(t, err)
+
+	assert.NoError(t, os.Mkdir(pathA, 0755))
+	assert.NoError(t, os.Mkdir(pathB, 0755))
 
 	// Create files
 	fooAPath := filepath.Join(pathA, "foo.go")
-	err = os.WriteFile(fooAPath, []byte(""), 0644)
-	assert.Nil(t, err)
 	barAPath := filepath.Join(pathA, "bar.go")
-	err = os.WriteFile(barAPath, []byte(""), 0644)
-	assert.Nil(t, err)
 	fooBPath := filepath.Join(pathB, "foo.go")
-	err = os.WriteFile(fooBPath, []byte(""), 0644)
-	assert.Nil(t, err)
 	barBPath := filepath.Join(pathB, "bar.go")
-	err = os.WriteFile(barBPath, []byte(""), 0644)
-	assert.Nil(t, err)
+	assert.NoError(t, os.WriteFile(fooAPath, []byte(""), 0644))
+	assert.NoError(t, os.WriteFile(barAPath, []byte(""), 0644))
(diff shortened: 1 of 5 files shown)


infrastructure/code/code_test.go
```diff
@@ -361,6 +361,49 @@ func Test_GetWorkspaceFiles_SkipIgnoredDirs(t *testing.T) {
 	assert.NotContains(t, walkedFiles, ignoredFileInDir)
 }
 
+func Test_FilesRespectIgnorePatterns(t *testing.T) {
+	tempDir := t.TempDir()
+	pathA := filepath.Join(tempDir, "A")
+	err := os.Mkdir(pathA, 0755)
+	assert.Nil(t, err)
```
bastiandoetsch, line 368: you can also use `assert.NoError`

bastiandoetsch (approved):
Looks good to me, and the test may be long, but I don't see how it can be made more concise without losing understandability.

infrastructure/code/ignores_test.go
```diff
@@ -64,7 +64,6 @@ func Test_IgnoresWithNegationInSnykCode(t *testing.T) {
 
 	calls := codeClientMock.GetAllCalls("extendBundleWithSource")
 	assert.Len(t, calls, 1)
-	assert.Contains(t, scanner.ignorePatterns, "!"+filepath.ToSlash(repobase+"/**/temp"))
```
michelkaporin, line 67: The test doesn't make sense without this assertion.

infrastructure/code/code_test.go
```diff
@@ -361,6 +361,49 @@ func Test_GetWorkspaceFiles_SkipIgnoredDirs(t *testing.T) {
...
+	pathB := filepath.Join(tempDir, "B")
+	err = os.Mkdir(pathB, 0755)
+	assert.Nil(t, err)
+
+	// Create files
+	fooAPath := filepath.Join(pathA, "foo.go")
```
michelkaporin, line 374: You can group this in 
```
fooAPath := filepath.Join(pathA, "foo.go")
barAPath := filepath.Join(pathA, "bar.go")
...
assert.NoError(t, os.WriteFile(fooAPath, []byte(""), 0644))
assert.NoError(t, os.WriteFile(barAPath, []byte(""), 0644))
...
```

michelkaporin (changes requested):

infrastructure/code/code_test.go
Avishagp: Nice! thank you

infrastructure/code/ignores_test.go
Avishagp: @asaf92

infrastructure/code/code_test.go
Avishagp: Nice, thank you!

infrastructure/code/ignores_test.go
asaf92: You're right, I'll refactor the test (and the one below it)

asaf92: I changed this and other ignore tests to be more generic, and test the call to the mock when scanning rather than the results of the private filtering functions

infrastructure/code/fake_snyk_code_api_service.go
```diff
@@ -211,6 +211,7 @@ func (f *FakeSnykCodeClient) ExtendBundle(
 	f.HasExtendedBundle = true
 	f.TotalBundleCount++
 	f.ExtendedBundleCount++
+	f.ExtendBundleFiles = files
```
asaf92, line 214: Maybe deep copy instead of assignment

michelkaporin (approved):

PR #273, opened 2023-04-03
fix: enable oauth feature flag in cli calls when using oauth2

bastiandoetsch:
### Description

Sets additional environment variable to enable oauth when CLI is spawned. Also adds debug output and a "token-refresh" API call before scans are executed.

### Checklist

- [ ] Linted

### Screenshots / GIFs

### DIFF
--- a/infrastructure/cli/cli.go
+++ b/infrastructure/cli/cli.go
@@ -18,7 +18,6 @@ package cli
 
 import (
 	"context"
-	"io"
 	"os"
 	"os/exec"
 	"strings"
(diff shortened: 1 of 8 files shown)


infrastructure/cli/cli.go
```diff
@@ -93,6 +94,17 @@ func (c SnykCli) doExecute(ctx context.Context, cmd []string, workingDir string,
 		if firstAttempt && shouldRetry {
 			output, err = c.doExecute(ctx, cmd, workingDir, false)
 		}
+	} else {
+		// ignore explicitly the errors as the following lines are only for debug output
```
PeterSchafer, line 98: question: this is an obsolete comment, isn't it?

PeterSchafer (approved):

PR #275, opened 2023-04-04
fix: UI reporting success when product errors

michelkaporin:
### Description

- Snyk Code didn't propagate error from GetFilters, resulting in UI showing success when API error happens
- UI was reporting success when IaC scan would finish, but Code is still running

### Checklist

- [x] Linted

### Screenshots / GIFs

### DIFF
--- a/infrastructure/code/bundle.go
+++ b/infrastructure/code/bundle.go
@@ -97,14 +97,14 @@ func getIssueLangAndRuleId(issue snyk.Issue) (string, string, bool) {
 
 // isAutofixSupportedForExtension shall call `GetFilters` if necessary and cache the results
 // in the `codeSettings` singleton.
-func (b *Bundle) isAutofixSupportedForExtension(ctx context.Context, file string) bool {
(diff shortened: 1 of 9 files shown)


infrastructure/code/bundle.go
```diff
@@ -165,7 +165,12 @@ func (b *Bundle) retrieveAnalysis(ctx context.Context) ([]snyk.Issue, error) {
 				for i := range issues {
 					// We only allow the issues whose file extensions are supported by the
 					// backend.
-					if !b.isAutofixSupportedForExtension(ctx, issues[i].AffectedFilePath) {
+					supported, err := b.isAutofixSupportedForExtension(ctx, issues[i].AffectedFilePath)
+					if err != nil {
+						return []snyk.Issue{}, err
+					}
```
asaf92, line 171: Why not log the error and continue? If we can't tell whether autofix is supported for some extension, does it mean we want to discard all the results?

michelkaporin: No, we error only if there's an error in retrieval this information. We won't be able progress here, since it's not possible to parse anything later.

asaf92 (approved):

PR #278, opened 2023-04-04
feat: implement token refresh command [HEAD-238]

bastiandoetsch:
### Description

### Checklist

- [x] Linted

### Screenshots / GIFs

### DIFF
--- a/application/config/config.go
+++ b/application/config/config.go
@@ -35,6 +35,7 @@ import (
 	"github.com/snyk/go-application-framework/pkg/app"
 	"github.com/snyk/go-application-framework/pkg/auth"
 	"github.com/snyk/go-application-framework/pkg/configuration"
+	localworkflows "github.com/snyk/go-application-framework/pkg/local_workflows"
 	"github.com/snyk/go-application-framework/pkg/workflow"
 	sglsp "github.com/sourcegraph/go-lsp"
 	"github.com/subosito/gotenv"
@@ -213,12 +214,20 @@ func New() *Config {
 	c.deviceId = c.determineDeviceId()
 	c.addDefaults()
 	c.filterSeverity = lsp.DefaultSeverityFilter()
-	c.engine = app.CreateAppEngine()
+	initWorkFlowEngine(c)
 	err := c.engine.Init()
 	if err != nil {
 		log.Warn().Err(err).Msg("Failed to initialize workflow engine")
 	}
 	return c
+}
+
+func initWorkFlowEngine(c *Config) {
+	c.engine = app.CreateAppEngine()
+	err := localworkflows.InitWhoAmIWorkflow(c.engine)
+	if err != nil {
+		log.Err(err).Msg("Failed to initialize WhoAmI workflow")
+	}
 }
 
 func (c *Config) AddBinaryLocationsToPath(searchDirectories []string) {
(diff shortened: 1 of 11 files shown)


domain/ide/command/oauth_refresh_test.go
```diff
@@ -0,0 +1,112 @@
...
+			receivedChan <- true
+		}
+	})
+
+	err := cmd.Execute(context.Background())
+
+	assert.NoErrorf(t, err, "cmd.Execute() error = %v", err)
+	assert.Equal(t, c.Token(), engineConfig.GetString(auth2.CONFIG_KEY_OAUTH_TOKEN))
+	assert.True(t, analytics.Identified)
+	assert.True(t, <-receivedChan)
```
michelkaporin, line 111: This will block test runs and time out in 10 mins only, if the listener won't hit.

domain/ide/command/oauth_refresh.go
```diff
@@ -0,0 +1,75 @@
...
+type oauthRefreshCommand struct {
+	command     snyk.CommandData
+	authService snyk.AuthenticationService
+}
+
+func (cmd *oauthRefreshCommand) Command() snyk.CommandData {
+	return cmd.command
+}
+
+func (cmd *oauthRefreshCommand) Execute(_ context.Context) error {
```
michelkaporin, line 42: How about we start providing some documentation comments over execute func in commands? We have it in readme, but it could be more technical here, answering questions like "Why?".

domain/ide/command/oauth_refresh.go
```diff
@@ -0,0 +1,75 @@
...
+	if c.AuthenticationMethod() != lsp.OAuthAuthentication {
+		log.Debug().Str("method", "oauthRefreshCommand.Execute").Msg("authentication method is token, no refresh needed")
+		return nil
+	}
+
+	oldToken := c.Token()
+
+	log.Debug().Str("method", "oauthRefreshCommand.Execute").Msgf("calling whoami workflow")
+	conf := c.Engine().GetConfiguration()
+	conf.Set("experimental", true)
```
michelkaporin, line 53: Does this modify global config? If yes, this line can lead to side effects when using engine in other places.

application/config/config.go
```diff
@@ -213,14 +214,22 @@ func New() *Config {
...
-	c.engine = app.CreateAppEngine()
+	initWorkFlowEngine(c)
 	err := c.engine.Init()
 	if err != nil {
 		log.Warn().Err(err).Msg("Failed to initialize workflow engine")
 	}
 	return c
 }
 
+func initWorkFlowEngine(c *Config) {
```
michelkaporin, line 225: Should initialisation logic leave in configuration? I suggest we keep this out of config to keep SRP.

bastiandoetsch: In the future this should done with dependency injection and initialization. right now, I don't think we can do it without cyclical deps between di.Init and config.

domain/ide/command/oauth_refresh.go
bastiandoetsch: changed

domain/ide/command/oauth_refresh_test.go
bastiandoetsch: can use assert.Eventually instead

application/config/config.go
michelkaporin: Can we leave a comment for that here please? We can also move this to a separate file for at least some visual separation of concerns.

michelkaporin (approved):

PR #272, opened 2023-04-03
fix: negation rules not respected in Snyk Code [HEAD-236]

asaf92:
### Description

Fixed a problem where negation rules were not respected in the file filtering of Snyk Code scans.

This PR also contains optimization of the file filtering in Snyk Code scans:
- Files are now streamed via a channel rather than waiting for all the filtered files to be collected before calling `createBundle`
- Ignore rules are being checked in parallel

### Checklist

- [x] Tests added and all succeed

PR #280, opened 2023-04-13
fix: code issue type determination logic

michelkaporin:
### Description

Correct implementation : https://github.com/snyk/vscode-extension/blob/main/src/snyk/snykCode/utils/analysisUtils.ts#L292
[Ref discussion](https://snyk.slack.com/archives/C0370TRPF7F/p1681312639073569?thread_ts=1681133733.381349&cid=C0370TRPF7F)

### Checklist

- [ ] Linted

### Screenshots / GIFs


bastiandoetsch:
This uses an experimental feature - should we really rely on functionality in golang/exp?

michelkaporin:
@bastiandoetsch it was already introduced as a dependency, so I assume this was discussed. I'd keep it, just be cautious what do we use from there. Slice manipulation is smth that ideally should be in the standard library anyway.

bastiandoetsch:
No, it wasn't discussed - and I'm wondering it was already there, as the license is added anew. I agree, we can leave it there for now, and if it doesn't work or go away, we can choose an alternative.

bastiandoetsch (approved):

michelkaporin:
@bastiandoetsch the license was mistakenly removed in #272. See my last comment there.

novalex (approved):
Thanks Michel, looks good 👌

PR #297, opened 2023-04-27
feat: remove /user/me call to Snyk API, improve authentication [HEAD-114][HEAD-296]

bastiandoetsch:
### Description

Expired Tokens failed silently. Now invalid tokens are identified, and the user is asked to Re-Authenticate with a message containing a button.

Also, whoami/user/me calls are now handled by the gaf extensible workflow, removing duplication.

### Checklist

- [x] Linted

### Screenshots / GIFs


PeterSchafer (commented):
I now reviewed and tested everything and the only thing I wonder if we should address it here is, that the user/me endpoint is called multiple times within a short time range. It could make sense to cache the result to reduce load but as far as I see, we don't have a caching capability right now. This could also be seen as a follow up task.

domain/ide/command/command_service.go
```diff
@@ -18,15 +18,27 @@ package command
...
+}
+
+func NewService(authService snyk.AuthenticationService, notifier noti.Notifier) snyk.CommandService {
+	return &serviceImpl{
+		authService: authService,
+		notifier:    notifier,
+	}
 }
 
 // ResetService resets the service instance to nil. This causes the next call to
```
asaf92, line 44: Remove `ResetService()`

out.txt
asaf92, line 1: Remove

application/config/config.go
```diff
@@ -417,12 +413,33 @@ func (c *Config) SetSeverityFilter(severityFilter lsp.SeverityFilter) bool {
...
+	// always update the token and auth method in the engine
+	c.token = token
+	c.m.Unlock()
+
+	_, err := c.TokenAsOAuthToken()
+	isOauthToken := err == nil
+	conf := c.engine.GetConfiguration()
+	if !isOauthToken && conf.GetString(configuration.AUTHENTICATION_TOKEN) != token {
+		log.Info().Msg("Token is not an OAuth token, setting token authentication in GAF")
+		conf.Set(configuration.AUTHENTICATION_TOKEN, token)
```
asaf92, line 427: Update `c.authenticationMethod`

michelkaporin (approved):

PeterSchafer (approved):

PR #301, opened 2023-05-04
feat: hide autofix behind feature flag

michelkaporin:
### Description

### Checklist

- [ ] Linted

### Screenshots / GIFs


infrastructure/code/bundle.go
```diff
@@ -200,14 +200,12 @@ func (b *Bundle) addCodeActions(ctx context.Context, issues []snyk.Issue) ([]sny
...
 				supported, err := b.isAutofixSupportedForExtension(ctx, issues[i].AffectedFilePath)
-				if err != nil {
-					return []snyk.Issue{}, err
-				}
-
-				if supported {
-					issues[i].CodeActions = append(issues[i].CodeActions, *b.createDeferredAutofixCodeAction(ctx, issues[i]))
-				} else {
-					log.Debug().Str("method", "addCodeActions").Msg("Autofix is not supported for " + issues[i].AffectedFilePath + " file extension.")
+				if err == nil {
```
bastiandoetsch, line 203: Why did you invert that logic? Checking `err` against `nil` is unintuitive, as it is usually the other way round.

bastiandoetsch (commented):
Looks good, but please have a look at my comment for the code action.

infrastructure/code/bundle.go
```diff
@@ -200,14 +200,12 @@ func (b *Bundle) addCodeActions(ctx context.Context, issues []snyk.Issue) ([]sny
...
+					if supported {
+						issues[i].CodeActions = append(issues[i].CodeActions, *b.createDeferredAutofixCodeAction(ctx, issues[i]))
+					} else {
```
bastiandoetsch, line 206: I'm aware, it is not the scope of the current PR, but it would be beneficial, to add the issue code rule to the code action to display something like `Autofix `javascript/sqlinjection`.

infrastructure/code/bundle.go
michelkaporin: I did invert, since we fail the whole analysis currently, if we fail to verify supported extensions. Refactored the logic a little to be more intuitive, please have a look.

infrastructure/code/bundle.go
michelkaporin: Fair point - updated it but the text might be changed once again later.

infrastructure/code/bundle.go
```diff
@@ -186,42 +183,42 @@ func (b *Bundle) retrieveAnalysis(ctx context.Context) ([]snyk.Issue, error) {
...
+		return
+	}
 
+	for i := range issues {
+		if autoFixEnabled {
+			// We only allow the issues whose file extensions are supported by the
+			// backend.
+			supported, err := b.isAutofixSupportedForExtension(ctx, issues[i].AffectedFilePath)
+			if err != nil {
+				log.Error().Str("method", method).Msg("Failed to check if autofix is supported for extension.")
```
bastiandoetsch, line 205: Would you please add the error to the log with `.Err(err)`? Also, it maybe beneficial to have it captured with error reporter (if not already done in the called method).

infrastructure/code/bundle.go
```diff
@@ -305,7 +302,7 @@ func (b *Bundle) createDeferredAutofixCodeAction(ctx context.Context, issue snyk
 		}
 	}
 
-	action, err := snyk.NewDeferredCodeAction("Attempt to AutoFix the issue (Snyk)", &autofixEditCallback, nil)
+	action, err := snyk.NewDeferredCodeAction("Attempt to fix suggestion "+issue.ID+" (Snyk)", &autofixEditCallback, nil)
```
bastiandoetsch, line 305: NICE!

bastiandoetsch (approved):
Only one very small change left :).

infrastructure/code/bundle.go
michelkaporin: Good catch, added it ✅

PR #303, opened 2023-05-12
feat: add progress reporting for Autofix [ZEN-608] [ZEN-609]

michelkaporin:
### Description

Adds progress and information reporting for Autofix.

### Checklist

- [ ] Linted

### Screenshots / GIFs

autofix

### DIFF
--- a/infrastructure/code/bundle.go
+++ b/infrastructure/code/bundle.go
@@ -26,8 +26,10 @@ import (
 	"time"
 
 	"github.com/rs/zerolog/log"
+	sglsp "github.com/sourcegraph/go-lsp"
 
 	"github.com/snyk/snyk-ls/application/config"
+	"github.com/snyk/snyk-ls/domain/ide/notification"
 	"github.com/snyk/snyk-ls/domain/observability/error_reporting"
 	"github.com/snyk/snyk-ls/domain/observability/performance"
 	"github.com/snyk/snyk-ls/domain/snyk"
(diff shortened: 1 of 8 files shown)


bastiandoetsch (commented):
Hey Michel, could you please add tests for your changes?

michelkaporin:
@bastiandoetsch what do you want to have tested?

PeterSchafer:
In general it looks good to me. I manually tested and reviewed the changes. I was also wondering about tests, for bundle.go for example.

infrastructure/code/bundle.go
```diff
@@ -129,7 +132,7 @@ func (b *Bundle) retrieveAnalysis(ctx context.Context) ([]snyk.Issue, error) {
 	}
 
 	p := progress.NewTracker(false)
```
PeterSchafer, line 134: Question: Do we want to use b.progress here now as well?

michelkaporin: I tested and this doesn't work unfortunately, as ProgressTracker is meant to be transient.
I'd love if you can make it more service-oriented. Currently LS panics when it's reused:
`panic: Called end progress twice. This breaks LSP in Eclipse fix me now and avoid headaches later`.

In fact, I have to revert the last commit, since it panics with the same message, when autofix requested twice.

PeterSchafer (approved):

PR #309, opened 2023-05-24
fix: setting log level via env-var freezes the server [HEAD-317]

asaf92:
### Description

Setting a log level via an env-var would lead to LS freezing. This happens because this leads to printing a log message to stdout, which is the communication channel between client and server.

### DIFF
--- a/application/config/config.go
+++ b/application/config/config.go
@@ -476,14 +476,10 @@ func (c *Config) ConfigureLogging(server lsp.Server) {
 	// env var overrides flag
 	envLogLevel := os.Getenv("SNYK_LOG_LEVEL")
 	if envLogLevel != "" {
-		fmt.Println("Setting log level from environment variable (SNYK_LOG_LEVEL)", envLogLevel)
 		envLevel, err := zerolog.ParseLevel(envLogLevel)
-		if err != nil {
-			fmt.Println("Can't set log level from env. Setting to default (=info)")
-			// fallback to flag
-			envLevel = logLevel
+		if err == nil {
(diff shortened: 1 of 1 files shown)


application/config/config.go
```diff
@@ -476,14 +476,10 @@ func (c *Config) ConfigureLogging(server lsp.Server) {
 	// env var overrides flag
 	envLogLevel := os.Getenv("SNYK_LOG_LEVEL")
 	if envLogLevel != "" {
-		fmt.Println("Setting log level from environment variable (SNYK_LOG_LEVEL)", envLogLevel)
```
bastiandoetsch, line 479: why not keep the messages and use `os.Stderr.WriteString` instead? That way we can keep the messages, but don't mess up stdout.

application/config/config.go
```diff
@@ -476,14 +476,10 @@ func (c *Config) ConfigureLogging(server lsp.Server) {
...
 		envLevel, err := zerolog.ParseLevel(envLogLevel)
-		if err != nil {
-			fmt.Println("Can't set log level from env. Setting to default (=info)")
-			// fallback to flag
-			envLevel = logLevel
+		if err == nil {
```
bastiandoetsch, line 480: This is unexpected to check `err == nil`. I'd keep it inverted.

application/config/config.go
asaf92: Done. I'm using `fmt.Fprintln(os.Stderr, ...) ` to keep it shorter

bastiandoetsch (approved):

PR #311, opened 2023-05-30
feat: request user feedback on Autofix [ZEN-635]

michelkaporin:
### Description

Asks for user feedback when successful fix is provided for Snyk Code results.

### Checklist

- [ ] Linted

### Screenshots / GIFs

Example feedback request message:
image

### DIFF
--- a/domain/ide/command/code_fix_feedback.go
+++ b/domain/ide/command/code_fix_feedback.go
@@ -1,0 +1,48 @@
+/*
+ * © 2023 Snyk Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package command
+
+import (
+	"context"
+
+	"github.com/snyk/snyk-ls/domain/snyk"
+)
+
+type SnykCodeHttpClient interface {
+	SubmitAutofixFeedback(ctx context.Context, fixId string, positive bool) error
+}
+
+type codeFixFeedback struct {
+	command   snyk.CommandData
+	apiClient SnykCodeHttpClient
+}
+
+func (cmd *codeFixFeedback) Command() snyk.CommandData {
+	return cmd.command
+}
+
+func (cmd *codeFixFeedback) Execute(ctx context.Context) (any, error) {
+	args := cmd.command.Arguments
+	fixId := args[0].(string)
+	positive := args[1].(bool)
+	err := cmd.apiClient.SubmitAutofixFeedback(ctx, fixId, positive)
+	if err != nil {
+		return nil, err
+	}
+
+	return nil, nil
+}
--- a/infrastructure/code/bundle.go
+++ b/infrastructure/code/bundle.go
@@ -25,15 +25,18 @@ import (
 	"strings"
 	"time"
 
+	pkgerrors "github.com/pkg/errors"
 	"github.com/rs/zerolog/log"
 	sglsp "github.com/sourcegraph/go-lsp"
 
 	"github.com/snyk/snyk-ls/application/config"
+	"github.com/snyk/snyk-ls/domain/ide/command"
 	"github.com/snyk/snyk-ls/domain/ide/notification"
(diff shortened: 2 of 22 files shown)


domain/ide/command/code_fix_feedback.go
```diff
@@ -0,0 +1,48 @@
...
+
+func (cmd *codeFixFeedback) Command() snyk.CommandData {
+	return cmd.command
+}
+
+func (cmd *codeFixFeedback) Execute(ctx context.Context) (any, error) {
+	args := cmd.command.Arguments
+	fixId := args[0].(string)
+	positive := args[1].(bool)
+	err := cmd.apiClient.SubmitAutofixFeedback(ctx, fixId, positive)
```
bastiandoetsch, line 42: If feedback fails, should we really return the error, or just log it and maybe capture it via Sentry?

infrastructure/code/bundle.go
```diff
@@ -317,26 +320,68 @@ func (b *Bundle) autofixFunc(ctx context.Context, issue snyk.Issue) func() *snyk
...
 
+func (b *Bundle) autofixFeedbackActions(fixId string) (*data_structure.OrderedMap[snyk.MessageAction, snyk.Command], error) {
+	createCommandData := func(positive bool) (snyk.Command, error) {
+		commandData := snyk.CommandData{
+			Title:     snyk.CodeSubmitFixFeedback,
+			CommandId: snyk.CodeSubmitFixFeedback,
+			Arguments: []any{fixId, positive},
+		}
+
+		cmd, err := command.CreateFromCommandData(commandData, nil, nil, nil, nil, nil, b.SnykCode)
```
bastiandoetsch, line 363: This is a clean code / style thingy ;). Calling the factory like this assumes whitebox-knowledge of the factory command creation. I'd advise not to assume how the factory works, thus I'd inject all parameters.

domain/ide/command/code_fix_feedback.go
michelkaporin: It'll be logged and captured by [the server handler](https://github.com/snyk/snyk-ls/blob/f0ac692403569e2ccb0a72a54a05940977ea7720/application/server/execute_command.go#L53) already.

infrastructure/code/bundle.go
michelkaporin: It's quite a bad design that Code scanner has to statically depend on all the dependencies of the command factory in the first place. Should a scanner really now about the server? Server is also an application-level dependency. Imagine there's another scanner that wants to reuse the factory. Suddenly it has to depend on Snyk Code Http Client.

Maybe not using a factory would be temporarily better in the current scenario.

P.S. commands are currently in the domain layer, which doesn't make sense for Scanner-specific commands. I believe they should be registered and handled in the infra layer. This could avoid also problems like the above,

domain/ide/command/code_fix_feedback.go
bastiandoetsch: Yes, but should an error be raised to the user? I'd say no.

infrastructure/code/bundle.go
bastiandoetsch: Happy to discuss these solutions. In the longer run, code scanner should be migrated out of the repository and included as dependency via go.mod. This will need some dependency refactoring. Server is not living in the application layer. If you refer directly to the server, this should change - [this](https://github.com/snyk/snyk-ls/blob/9085f62a3a73a5a6e4e9aaa5ba450fdda6acdc6c/internal/lsp/server.go#L41) should be referenced.

domain/ide/command/code_fix_feedback.go
michelkaporin: Yes, the user should be aware that we failed to submit the feedback.

infrastructure/code/bundle.go
michelkaporin: You are asking to inject all the parameters, where one of them is reference to server. How do you see this happening now? There're authentication service, issue service dependencies that need to be passed as well.

bastiandoetsch: I didn't say that. There's different solutions. 

The problem is that the command is created in the infrastructure layer, not only the command data. The flow is:

```
bundle (infrastructure) -> command factory (ide/command) -> ... -> client
```

The intended flow is that commands are created from command data by the command handler using the command factory and then executed. The command handler has all the necessary dependencies to inject without needing to know how the factory works.

```
client -> command handler -> command factory -> command -> execute -> logic in infrastructure
```

So only CommandData should be used, not the Command. Anyhoo, I'm open to a call to discuss further.

patricia-v (approved):

infrastructure/code/bundle.go
michelkaporin: Let's address this in a separate PR. I've started work on it in `feat/idea-autofix-feedback/command-data` branch.

bastiandoetsch (approved):

PR #317, opened 2023-06-01
fix: refresh code lenses after publishing diagnostics

bastiandoetsch:
### Description

After publishing new diagnostics, we need to update code lenses, too.

### Checklist

- [x] Linted

### Screenshots / GIFs

### DIFF
--- a/application/server/notification.go
+++ b/application/server/notification.go
@@ -22,6 +22,7 @@ import (
 	"github.com/rs/zerolog/log"
 	sglsp "github.com/sourcegraph/go-lsp"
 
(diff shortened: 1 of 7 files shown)


application/server/notification.go
```diff
@@ -108,6 +109,7 @@ func registerNotifier(srv lsp.Server) {
 			if len(params.Diagnostics) > 0 {
 				source = params.Diagnostics[0].Source
 			}
+			handleCodelensRefresh(srv)
```
bastiandoetsch, line 112: This is where the code lenses are refreshed after publishing!

PR #334, opened 2023-06-16
feat: introduce vulnerability count service [HEAD-280]

bastiandoetsch:
### Description

Service groups and counts issues and forwards them to a given processor function

### Checklist

- [x] Linted

### Screenshots / GIFs

### DIFF
--- a/infrastructure/oss/vulnerability_count_test.go
+++ b/infrastructure/oss/vulnerability_count_test.go
@@ -1,0 +1,190 @@
+/*
+ * © 2023 Snyk Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


infrastructure/oss/vulnerability_count_test.go
```diff
@@ -0,0 +1,193 @@
...
+			assert.Equal(t, 1, vci.total)
+			assert.Equal(t, 1, vci.severityCounts[snyk.Medium])
+			assert.Equal(t, "id2", vci.mostSevereVulnerabilityId)
+		} else {
+			assert.Fail(t, "Unexpected file path")
+		}
+		return nil
+	}
+
+	err := vc.ProcessVulnerabilityCount(issues, testFunc)
```
PeterSchafer, line 190: Suggestion: Currently these tests would all pass even if the processor would not be called at all. How about setting an expectation of how often a processor is called in a test.

PeterSchafer (approved):

infrastructure/oss/vulnerability_count_test.go
bastiandoetsch: True. I'll add it.

bastiandoetsch: Will add a channel post for each invocation, read from it from the test and increase the count on read, then check the count.

PR #335, opened 2023-06-18
chore: log integration names and versions [HEAD-375]

asaf92:


infrastructure/cli/initializer.go
```diff
@@ -124,6 +137,8 @@ func (i *Initializer) installCli() {
 
 	if cliPath != "" {
 		i.notifier.Send(lsp.SnykIsAvailableCli{CliPath: cliPath})
+		version := i.versionSource.CliVersion()
+		log.Info().Msg("snyk-cli: " + version + " (" + cliPath + ")")
```
PeterSchafer, line 141: Suggestion: Currently the CLI info is only printed after being update/install. It would be helpful if the CLI info would also be part of the logs when not being updated.

application/server/server.go
```diff
@@ -190,10 +191,19 @@ func initNetworkAccessHeaders(n networking.NetworkAccess) {
...
+	if initParams.IntegrationName != "" && initParams.IntegrationVersion != "" {
+		log.Info().Msg("snyk-plugin: " + initParams.IntegrationName + "/" + initParams.IntegrationVersion)
+	}
+}
+
 func initializeHandler(srv *jrpc2.Server, c *config.Config) handler.Func {
 	return handler.New(func(ctx context.Context, params lsp.InitializeParams) (any, error) {
 		method := "initializeHandler"
-		log.Info().Str("method", method).Interface("params", params).Msg("RECEIVING")
+		logIntegration(params) // Ideally this would be logged in the "initialized" handler, but the parameters are not available there
```
PeterSchafer, line 205: Suggestion: Assuming that initializedHandler() comes after initializeHandler(), how about moving logIntegration() down in initializedHandler(), so that the information is logged as one block?

application/di/test_init.go
```diff
@@ -60,15 +60,15 @@ func TestInit(t *testing.T) {
 	authProvider := snyk.NewFakeCliAuthenticationProvider()
 	snykApiClient = &snyk_api.FakeApiClient{CodeEnabled: true}
 	authenticationService = snyk.NewAuthenticationService(authProvider, analytics, errorReporter, notifier)
-	cliInitializer = cli.NewInitializer(errorReporter, installer, notifier)
+	cliInitializer = cli.NewInitializer(errorReporter, installer, notifier, nil)
```
bastiandoetsch, line 63: shouldn't the snykCli assignment be moved above this line then passed  to the initializer instead of nil?

application/server/server.go
```diff
@@ -276,8 +278,19 @@ func initializeHandler(srv *jrpc2.Server, c *config.Config) handler.Func {
 }
 func initializedHandler(srv *jrpc2.Server) handler.Func {
 	return handler.New(func(ctx context.Context, params lsp.InitializedParams) (any, error) {
+		// Logging these messages only after the client has been initialized.
+		// It has been observed that VSCode changes the log messages format after the LS has been initialized.
```
bastiandoetsch, line 282: It's because it's after server initialization, the `logMessage` request can be send for logging purposes (see LSP protocol)

application/server/server.go
```diff
@@ -276,8 +278,19 @@ func initializeHandler(srv *jrpc2.Server, c *config.Config) handler.Func {
...
+		// No reason to log the method name for these messages, because some of these values are empty and the messages
+		// looks weird when including the method name.
+		c := config.CurrentConfig()
+		log.Info().Msg("snyk-ls: " + config.Version + " (" + util.Result(os.Executable()) + ")")
+		log.Info().Msg("platform: " + runtime.GOOS + "/" + runtime.GOARCH)
+		log.Info().Msg("https_proxy: " + os.Getenv("HTTPS_PROXY"))
+		log.Info().Msg("http_proxy: " + os.Getenv("HTTP_PROXY"))
+		log.Info().Msg("no_proxy: " + os.Getenv("NO_PROXY"))
+		log.Info().Msg("IDE: " + c.IntegrationName() + "/" + c.IdeVersion())
```
bastiandoetsch, line 291: Is this correct? Is integration name the IDE in all cases?

bastiandoetsch: Maybe clientInfo.name instead? `"clientInfo":{"name":"Sublime Text LSP","version":"1.24.0"}`

application/server/server.go
asaf92: Thanks, I'll update the comment

application/server/server.go
asaf92: For the mainstream plugins, it's going to be the same.
For unofficial IDE, `setClientInformation` is going to set integration name to the ClientInfo.Name, so it's going to be the same there as well

application/di/test_init.go
asaf92: Fixed

application/server/server.go
```diff
@@ -342,23 +355,29 @@ func addWorkspaceFolders(params lsp.InitializeParams, w *workspace.Workspace) {
...
 func setClientInformation(initParams lsp.InitializeParams) {
 	var integrationName, integrationVersion string
 	if initParams.InitializationOptions.IntegrationName != "" {
 		integrationName = initParams.InitializationOptions.IntegrationName
 		integrationVersion = initParams.InitializationOptions.IntegrationVersion
 	} else if initParams.ClientInfo.Name != "" {
 		integrationName = strings.ToUpper(strings.Replace(initParams.ClientInfo.Name, " ", "_", -1))
-		integrationVersion = initParams.ClientInfo.Version
-	} else if integrationNameEnvVar := os.Getenv("SNYK_INTEGRATION_NAME"); integrationNameEnvVar != "" {
+		integrationVersion = ""
```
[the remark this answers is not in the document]
asaf92: Removed

application/server/server_test.go
```diff
@@ -520,6 +527,7 @@ func Test_initialize_integrationInClientInfo_readFromClientInfo(t *testing.T) {
 	currentConfig := config.CurrentConfig()
 	assert.Equal(t, expectedIntegrationName, currentConfig.IntegrationName())
 	assert.Equal(t, expectedIntegrationVersion, currentConfig.IntegrationVersion())
+	assert.Equal(t, expectedIdeVersion, currentConfig.IdeVersion())
```
PeterSchafer, line 530: suggestion: maybe update the test to include IdeName and use a different value than IntegrationName

PeterSchafer (approved):

PR #336, opened 2023-06-19
feat: provide vulnerability count as inline value on Open Source scan

bastiandoetsch:
### Description

### Checklist

- [x] Linted

### Screenshots / GIFs

### DIFF
--- a/application/server/inline_values.go
+++ b/application/server/inline_values.go
@@ -1,0 +1,51 @@
+/*
+ * © 2023 Snyk Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package server
+
+import (
+	"context"
+
+	"github.com/creachadair/jrpc2"
+	"github.com/creachadair/jrpc2/handler"
+
+	"github.com/snyk/snyk-ls/application/config"
+	"github.com/snyk/snyk-ls/application/di"
+	"github.com/snyk/snyk-ls/domain/ide/converter"
+	"github.com/snyk/snyk-ls/domain/snyk"
+	"github.com/snyk/snyk-ls/internal/lsp"
+	"github.com/snyk/snyk-ls/internal/uri"
+)
+
+func textDocumentInlineValueHandler(c *config.Config) jrpc2.Handler {
+	return handler.New(func(ctx context.Context, params lsp.InlineValueParams) (any, error) {
+		logger := c.Logger().With().Str("method", "textDocumentInlineValueHandler").Logger()
+		documentURI := params.TextDocument.URI
+		logger.Info().Msgf("Request for %s:%s RECEIVED", documentURI, params.Range.String())
+		defer logger.Info().Msgf("Request for %s:%s DONE", documentURI, params.Range.String())
+		if s, ok := di.Scanner().(snyk.InlineValueProvider); ok {
+			filePath := uri.PathFromUri(documentURI)
+			values, err := s.GetInlineValues(filePath, converter.FromRange(params.Range))
+			if err != nil {
+				return nil, err
+			}
+			lspInlineValues := converter.ToInlineValues(values)
+			logger.Debug().Msgf("found %d inline values for %s", len(values), filePath)
+			return lspInlineValues, nil
+		}
+		return nil, nil
+	})
+}
--- a/application/server/notification.go
+++ b/application/server/notification.go
@@ -109,7 +109,6 @@ func registerNotifier(srv lsp.Server) {
 			if len(params.Diagnostics) > 0 {
 				source = params.Diagnostics[0].Source
 			}
-			handleCodelensRefresh(srv)
 			log.Info().
 				Str("method", "registerNotifier").
 				Interface("documentURI", params.URI).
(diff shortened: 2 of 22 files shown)


application/server/inline_values.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/snyk/snyk-ls/internal/uri"
+)
+
+func textDocumentInlineValueHandler(c *config.Config) jrpc2.Handler {
+	return handler.New(func(ctx context.Context, params lsp.InlineValueParams) (any, error) {
+		logger := c.Logger().With().Str("method", "textDocumentInlineValueHandler").Logger()
+		documentURI := params.TextDocument.URI
+		logger.Info().Msgf("Request for %s:%s RECEIVED", documentURI, params.Range.String())
+		defer logger.Info().Msgf("Request for %s:%s DONE", documentURI, params.Range.String())
+		if s, ok := di.Scanner().(snyk.InlineValueSupporter); ok {
```
PeterSchafer, line 39: Question: is it correct that the range is not considered here at all? If so, maybe we don't need to print it as well.

application/server/inline_values.go
```diff
@@ -0,0 +1,51 @@
...
+			filePath := uri.PathFromUri(documentURI)
+			values, err := s.GetInlineValues(filePath)
+			if err != nil {
+				return nil, err
+			}
+			lspInlineValues := converter.ToInlineValues(values)
```
PeterSchafer, line 45: nitpick: it is a bit confusing that the output of `GetInlineValues()` needs to be converted by `ToInlineValues()`. So either of these method names is a bit misleading.

infrastructure/oss/vulnerability_count.go
```diff
@@ -75,27 +85,71 @@ func (v *vulnerabilityCountImpl) ProcessVulnerabilityCount(
...
+		vci.severityCounts[snyk.High],
+		vci.severityCounts[snyk.Medium],
+		vci.severityCounts[snyk.Low],
+		vci.mostSevereVulnerabilityId,
+	)
+	return text
+}
+
+func (oss *Scanner) toInlineValueAndAddToCache(vci *VulnerabilityCountInformation) snyk.InlineValue {
+	values := oss.inlineValues[vci.FilePath]
```
PeterSchafer, line 134: suggestion: it is a bit tricky to read this code as it uses values and value. maybe rename values to cachedValues, to better differentiate.

infrastructure/oss/vulnerability_count.go
```diff
@@ -75,27 +85,71 @@ func (v *vulnerabilityCountImpl) ProcessVulnerabilityCount(
...
+func (oss *Scanner) addVulnerabilityCountsAsInlineValuesToCache(issues []snyk.Issue) {
+	c := config.CurrentConfig()
+	logger := c.Logger().With().Str("method", "addVulnerabilityCountsAsInlineValuesToCache").Logger()
+	vc := NewVulnerabilityCountService(c)
+	counts, err := vc.DetermineVulnerabilityCounts(issues)
+	if err != nil {
+		logger.Err(err).Msg("couldn't get vulnerability counts")
+		oss.errorReporter.CaptureError(err)
+	}
+	for _, r := range counts {
```
PeterSchafer, line 110: suggestion: could we find better variable names for `r` and maybe even `vci`?

application/server/inline_values.go
bastiandoetsch: This may be incorrect, gotta check

application/server/inline_values.go
bastiandoetsch: It's consistent with the other converts from domain objects to lsp objects

infrastructure/oss/inline_value.go
```diff
@@ -0,0 +1,40 @@
...
+
+package oss
+
+import (
+	"github.com/rs/zerolog/log"
+
+	"github.com/snyk/snyk-ls/domain/snyk"
+)
+
+func (oss *Scanner) GetInlineValues(path string, myRange snyk.Range) ([]snyk.InlineValue, error) {
```
PeterSchafer, line 25: nitpick: In this case not super important, but for structures it might be more efficient to use a pointers.

infrastructure/oss/inline_value.go
```diff
@@ -0,0 +1,40 @@
...
+	method := "GetInlineValues"
+	log.Debug().Str("method", method).Str("path", path).Msg("called")
+
+	value := oss.inlineValues[path][myRange.Start.Line]
```
PeterSchafer, line 29: Question: The current assumption here seems to me, that start line and end line are the same. Is this actually correct. In the lsp specification it says
> The document range for which inline values should be computed.

So I would be assuming that the range can span multiple lines and we would need to iterate over the inlineValues here to determine if they are in range or not.

bastiandoetsch: For OSS vulnerability count it's only the line that's relevant though.  That's why in general, we work with the range, but in the concrete implementation not.

PeterSchafer: I understand that the OSS vulnerability count is associated to a single line. But isn't this logic a lookup/search functionality that is triggered by the lsp client and the lsp client doesn't know anything about this, it only known the viewport of a document to show.

infrastructure/oss/inline_value.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+func (oss *Scanner) GetInlineValues(path string, myRange snyk.Range) (result []snyk.InlineValue, err error) {
+	logger := log.With().Str("method", "GetInlineValues").Logger()
+	logger.Debug().Str("path", path).Msg("called")
+
+	inlineValues := oss.inlineValues[path]
+	if inlineValues == nil {
+		logger.Debug().Str("path", path).Msg("no inlineValues found")
+		return nil, nil
```
PeterSchafer, line 32: Suggestion: Please return an empty slice here, otherwise [this](https://github.com/snyk/snyk-ls/pull/336/files#diff-a44b6d442dbb9afbfc2d35aa88806cfb9b0cb6c4a10508cb40e9c7404ec2fad6R83) will have an issue.

PeterSchafer (approved):

application/server/notification.go
```diff
@@ -109,7 +109,6 @@ func registerNotifier(srv lsp.Server) {
 			if len(params.Diagnostics) > 0 {
 				source = params.Diagnostics[0].Source
 			}
-			handleCodelensRefresh(srv)
```
PeterSchafer, line 112: Question: why was this removed?

PeterSchafer: seems to be unnecessary since a test sends it anyways.

PeterSchafer (approved):

PR #337, opened 2023-06-20
fix: orgID as orgName

j-luong:
### Description

_This PR should fix an issue where the orgId is passed as the orgName in `sarif_utils.go`_

### Checklist

- [ ] Linted


infrastructure/code/sarif_utils.go
```diff
@@ -20,18 +20,18 @@ import "github.com/snyk/snyk-ls/application/config"
 
 func newCodeRequestContext() codeRequestContext {
 	unknown := "unknown"
-	orgName := unknown
+	orgId := unknown
 	if config.CurrentConfig().Organization() != "" {
-		orgName = config.CurrentConfig().Organization()
+		orgId = config.CurrentConfig().Organization()
```
bastiandoetsch, line 25: How do you know that this is always the ID? What happens if we put the org name into PublicID?

PeterSchafer: This is a GAF feature, the config always returns the org id.

strassl-snyk (approved):

infrastructure/code/sarif_utils.go
bastiandoetsch: That may be, but the user can overwrite it with a name.

PeterSchafer: Just as a final summary on this thread. @bastiandoetsch and I looked at the code and thereby could resolve his concerns.

PR #342, opened 2023-06-23
fix: Use common function to generate user-agent

PeterSchafer:
### Description
Use application framework to generate a common user-agent header. This includes filling fields consistently.

### DIFF
--- a/application/config/config.go
+++ b/application/config/config.go
@@ -677,14 +677,22 @@ func (c *Config) addDefaults() {
 
 func (c *Config) SetIntegrationName(integrationName string) {
 	c.integrationName = integrationName
+	c.Engine().GetConfiguration().Set(configuration.INTEGRATION_NAME, integrationName)
 }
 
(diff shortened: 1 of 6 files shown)


application/config/config.go
```diff
@@ -677,14 +677,22 @@ func (c *Config) addDefaults() {
...
 
 func (c *Config) SetIntegrationVersion(integrationVersion string) {
 	c.integrationVersion = integrationVersion
+	c.Engine().GetConfiguration().Set(configuration.INTEGRATION_VERSION, integrationVersion)
 }
 
-func (c *Config) SetIdeName(ideName string)       { c.ideName = ideName }
-func (c *Config) SetIdeVersion(ideVersion string) { c.ideVersion = ideVersion }
+func (c *Config) SetIdeName(ideName string) {
+	c.ideName = ideName
```
bastiandoetsch, line 689: Why not get rid of the storage in LS config entirely and only use GAF?

bastiandoetsch (approved):

PR #353, opened 2023-07-28
fix: architecture determination for download script

bastiandoetsch:
### Description

Previously, the arm64 on mac was not detected correctly. Also, downloads failed if no write access to /usr/local/bin was granted. This PR fixes the arm64 detection and writes to the current working directory, if writing to /usr/local/bin fails

### Checklist

- [x] Linted

### Screenshots / GIFs

### DIFF
--- a/getLanguageServer.sh
+++ b/getLanguageServer.sh
@@ -24,7 +24,7 @@ OS=$(uname -s | tr '[:upper:]' '[:lower:]')
 ARCH=$(uname -m | tr '[:upper:]' '[:lower:]')
 if [[ $ARCH == "x86_64" ]]; then
   ARCH="amd64"
-elif [[ $ARCH == "aarch64" ]]; then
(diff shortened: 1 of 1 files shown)


getLanguageServer.sh
```diff
@@ -35,17 +35,33 @@ VERSION=$(curl -sSL --compressed https://static.snyk.io/snyk-ls/$PROTOCOL_VERSIO
 DESTINATION="/usr/local/bin/snyk-ls"
 DOWNLOAD_URL="https://static.snyk.io/snyk-ls/$PROTOCOL_VERSION/snyk-ls_${VERSION}_${OS}_${ARCH}"
 
+set +e
+if [[ -f $DESTINATION ]]; then
```
PeterSchafer, line 39: Question: if the check below realizes that `DESTINATION` is not accessible, this logic is not executed for `$PWD/snyk-ls`. does it make sense to move the condition in line 48 up?

PeterSchafer (approved):

PR #360, opened 2023-08-23
feat: scan html files for inline values

bastiandoetsch:
### Description

provide package scanning for html files to display inline values

### Checklist

- [x] Linted

### Screenshots / GIFs

image


domain/snyk/scanner.go
```diff
@@ -32,6 +32,11 @@ import (
 	"github.com/snyk/snyk-ls/internal/product"
 )
 
+var (
```
Avishagp, line 35: What is this used for?

bastiandoetsch: These enforce that the class implements the `Scanner` and the `InlineValueProvider` interfaces. It's a pattern in go to ensure at compile time.

Avishagp (approved):

PR #362, opened 2023-08-24
feat: disable sentry and amplitude events in fedramp instance [HEAD-134]

bastiandoetsch:
### Description

Sentry and Amplitude should not be contacted when using the fedramp instance. This PR eenforces this.

### Checklist

- [x] Linted

### Screenshots / GIFs


application/config/config.go
```diff
@@ -851,3 +851,17 @@ func (c *Config) IdeVersion() string {
...
 	return c.Engine().GetConfiguration().GetString(configuration.INTEGRATION_ENVIRONMENT)
 }
+
+func (c *Config) IsFedramp() bool {
+	u, err := url.Parse(c.SnykApi())
+	if err != nil {
+		return false
+	}
+	hostnameParts := strings.Split(u.Hostname(), ".")
+	if len(hostnameParts) < 3 {
```
Avishagp, line 861: Is this check fool proof?

bastiandoetsch: Actually yes. A fedramp instance would have at least two `.`. Take for example: `https://app.fedramp.snykgov.io`. But anyhoo, it shouuld be `<4` ;)

application/config/config.go
```diff
@@ -851,3 +851,17 @@ func (c *Config) IdeVersion() string {
...
 	return c.Engine().GetConfiguration().GetString(configuration.INTEGRATION_ENVIRONMENT)
 }
+
+func (c *Config) IsFedramp() bool {
+	u, err := url.Parse(c.SnykApi())
+	if err != nil {
+		return false
+	}
+	hostnameParts := strings.Split(u.Hostname(), ".")
+	if len(hostnameParts) < 4 {
```
Avishagp, line 861: Could we add a comment here to explain this logic?

Avishagp (approved):

PR #364, opened 2023-08-25
chore: prepare LS as CLI Extension

PeterSchafer:
### Description

* Introduce CLI Extension entry point in parallel to the existing entry point in main to remain existing behaviour.
* Introduce execution mode (standalone/extension) to differentiate
  * automatic binary management 
  * CLI Executor

### Checklist

- [x] Linted


application/config/config.go
```diff
@@ -626,6 +629,13 @@ func (c *Config) SetManageBinariesAutomatically(enabled bool) {
 	c.manageBinariesAutomatically.Set(enabled)
 }
 
+func (c *Config) ManageCliBinariesAutomatically() bool {
+	if c.engine.GetConfiguration().GetString(cli_constants.EXECUTION_MODE_KEY) == cli_constants.EXECUTION_MODE_VALUE_EXTENSION {
```
bastiandoetsch, line 633: Why do we need `EXECUTION_MODE_VALUE_EXTENSION? Can't we use `EXECUTION_MODE_VALUE_STANDALONE`? The LS doesn't need to know if there is an extension mode.

infrastructure/cli/cli_constants/constants.go
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cli_constants
+
+const (
+	EXECUTION_MODE_KEY              string = "execution-mode"
+	EXECUTION_MODE_VALUE_EXTENSION  string = "extension"
```
bastiandoetsch, line 21: Could this go away?

infrastructure/cli/cli_extension_executor_test.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/snyk/go-application-framework/pkg/configuration"
+	"github.com/snyk/go-application-framework/pkg/workflow"
+	"github.com/spf13/pflag"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/snyk/snyk-ls/application/config"
+)
+
+func Test_ExecuteLegacyCLI_SUCCESS(t *testing.T) {
+
```
bastiandoetsch, line 33: Is this a unit, integration or smoke test? Could you maybe add the appropriate testutil.IntegTest or such marker?

ls_extension/main.go
```diff
@@ -0,0 +1,117 @@
+/*
```
bastiandoetsch, line 1: could we share code between main.go and this file?

ls_extension/main.go
```diff
@@ -0,0 +1,117 @@
...
+) (output []workflow.Data, err error) {
+	output = []workflow.Data{}
+
+	c := config.CurrentConfig()
+	zerolog.SetGlobalLevel(zerolog.InfoLevel)
+
+	logger := invocation.GetEnhancedLogger()
+	extensionConfig := invocation.GetConfiguration()
+
+	logger.Log().Msgf("Language Server here!")
```
bastiandoetsch, line 79: Maybe just output "Language Server version: %s" or such?

bastiandoetsch: Also, most likely this should go to stderr - is that ensured?

bastiandoetsch (approved):

ls_extension/main.go
PeterSchafer: Makes sense I'll adapt the text. This logger is safe to use though as it is already initialized by the engine before invoking the extension.

ls_extension/main.go
PeterSchafer: add entrypoint helpers, to at least share some snippets.

infrastructure/cli/cli_constants/constants.go
```diff
@@ -0,0 +1,23 @@
+/*
+ * © 2022 Snyk Limited All rights reserved.
```
bastiandoetsch, line 2: This should be 2023 ;)

infrastructure/cli/cli_extension_executor_test.go
```diff
@@ -0,0 +1,92 @@
+/*
+ * © 2022 Snyk Limited All rights reserved.
```
bastiandoetsch, line 2: This, too *duckandrun*

infrastructure/cli/initializer.go
```diff
@@ -66,7 +66,7 @@ func (i *Initializer) Init() error {
 	logger := log.With().Str("method", "cli.Init").Logger()
 	cliInstalled := config.CurrentConfig().CliSettings().Installed()
 	logger.Debug().Str("cliPath", cliPathInConfig()).Msgf("CLI installed: %v", cliInstalled)
-	if !config.CurrentConfig().ManageBinariesAutomatically() {
+	if !config.CurrentConfig().ManageCliBinariesAutomatically() {
```
bastiandoetsch, line 69: This is very nitpicky of me. But it is only one binary. So it should most likely be `ManageCliBinaryAutomatically`. Not a showstopper, though.

ls_extension/main.go
```diff
@@ -0,0 +1,105 @@
+/*
+ * © 2022 Snyk Limited All rights reserved.
```
bastiandoetsch, line 2: same :)

bastiandoetsch (approved):

PR #373, opened 2023-09-07
fix: code api changes

jozsef-armin-hamos:
### Description

### Checklist

- [ ] Linted

### Screenshots / GIFs

### DIFF
--- a/application/config/config.go
+++ b/application/config/config.go
@@ -61,7 +61,7 @@ const (
 	snykCodeTimeoutKey       = "SNYK_CODE_TIMEOUT" // timeout as duration (number + unit), e.g. 10m
 	DefaultSnykApiUrl        = "https://snyk.io/api"
 	DefaultDeeproxyApiUrl    = "https://deeproxy.snyk.io"
-	DefaultSnykCodeApiGovUrl = "https://api.snyk.io"
+	DefaultSnykCodeApiGovUrl = "https://api.snykgov.io"
 	pathListSeparator        = string(os.PathListSeparator)
 	windows                  = "windows"
 )
@@ -375,7 +375,7 @@ func (c *Config) UpdateApiEndpoints(snykApiUrl string) bool {
 		c.snykApiUrl = snykApiUrl
 
 		// Update Code API endpoint
-		snykCodeApiUrl, err := getCodeApiUrlFromCustomEndpoint(snykApiUrl, c.IsFedramp())
+		snykCodeApiUrl, err := c.GetCodeApiUrlFromCustomEndpoint(snykApiUrl)
 		if err != nil {
 			log.Error().Err(err).Msg("Couldn't obtain Snyk Code API url from CLI endpoint.")
 		}
@@ -546,7 +546,7 @@ func (c *Config) DisableLoggingToFile() {
 
 func (c *Config) SetConfigFile(configFile string) { c.configFile = configFile }
 
-func getCodeApiUrlFromCustomEndpoint(endpoint string, isFedramp bool) (string, error) {
+func (c *Config) GetCodeApiUrlFromCustomEndpoint(endpoint string) (string, error) {
 	// Code API endpoint can be set via env variable for debugging using local API instance
 	deeproxyEnvVarUrl := strings.Trim(os.Getenv(deeproxyApiUrlKey), "/")
 	if deeproxyEnvVarUrl != "" {
@@ -554,7 +554,7 @@ func getCodeApiUrlFromCustomEndpoint(endpoint string, isFedramp bool) (string, e
 	}
 
 	if endpoint == "" {
-		if isFedramp {
+		if c.IsFedramp() {
 			return DefaultSnykCodeApiGovUrl, nil
 		} else {
 			return DefaultDeeproxyApiUrl, nil
@@ -570,7 +570,7 @@ func getCodeApiUrlFromCustomEndpoint(endpoint string, isFedramp bool) (string, e
 	m := regexp.MustCompile(`^(ap[pi]\.)?`)
 
 	subdomain := "deeproxy."
-	if isFedramp {
+	if c.IsFedramp() {
 		subdomain = "api."
 	}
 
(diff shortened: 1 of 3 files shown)


florinMiro (approved):

application/config/config.go
```diff
@@ -561,7 +568,13 @@ func getCodeApiUrlFromCustomEndpoint(endpoint string) (string, error) {
 	}
 
 	m := regexp.MustCompile(`^(ap[pi]\.)?`)
-	endpointUrl.Host = m.ReplaceAllString(endpointUrl.Host, "deeproxy.")
+
+	subdomain := "deeproxy."
+	if isFedramp {
+		subdomain = "api."
```
michelkaporin, line 574: That might not work if a user provides a custom endpoint as `https://api.snykgov.io` in their settings (updated via [here](https://github.com/snyk/snyk-ls/blob/adc52f1731a191eef1b398ff3cffe6c1d1973224/application/config/config.go#L376)).

application/config/config_test.go
```diff
@@ -161,43 +161,62 @@ func Test_loadFile(t *testing.T) {
...
 
 func TestSnykCodeApi(t *testing.T) {
+	t.Setenv("DEEPROXY_API_URL", "")
 	t.Run("endpoint not provided", func(t *testing.T) {
 
-		codeApiEndpoint, _ := getCodeApiUrlFromCustomEndpoint("")
+		codeApiEndpoint, _ := getCodeApiUrlFromCustomEndpoint("", false)
 		assert.Equal(t, "https://deeproxy.snyk.io", codeApiEndpoint)
+
+		codeApiEndpoint, _ = getCodeApiUrlFromCustomEndpoint("", true)
```
michelkaporin, line 170: This and test cases below can be simplified using table-driven tests. Using an array of test inputs and expected outputs for testing is more readable and maintainable.

application/config/config_test.go
```diff
@@ -161,43 +161,62 @@ func Test_loadFile(t *testing.T) {
 }
 
 func TestSnykCodeApi(t *testing.T) {
+	t.Setenv("DEEPROXY_API_URL", "")
```
michelkaporin, line 164: Is it really needed? If yes, can we reuse `deeproxyApiUrlKey` const?

application/config/config.go
```diff
@@ -866,6 +884,28 @@ func (c *Config) IsFedramp() bool {
...
+		return c.SnykCodeApi(), nil
+	}
+
+	// api url is fedramp, but code api url hasn't been updated properly
+	if strings.Contains(c.SnykCodeApi(), "deeproxy.") {
+		snykCodeApiUrl, err := getCodeApiUrlFromCustomEndpoint(c.SnykApi(), c.IsFedramp())
+		if err != nil {
+			return "", err
+		}
+		c.SetSnykCodeApi(snykCodeApiUrl)
```
michelkaporin, line 901: Getter functions shouldn't be mutating. As a caller of `Get` I expect it's not doing any side-effect mutation.

application/config/config.go
```diff
@@ -866,6 +884,28 @@ func (c *Config) IsFedramp() bool {
 	host := strings.ToLower(u.Hostname())
 
 	// fedramp instance should have the format https://*.snykgov.io
-	snykgovDomain := strings.HasSuffix(host, ".snykgov.io")
-	return snykgovDomain
+	return isFedrampHost(host)
+}
+
+func (c *Config) GetSnykCodeApi() (string, error) {
```
michelkaporin, line 890: We have `SnykCodeApi` getter already at line :338.

application/config/config.go
```diff
@@ -866,6 +884,28 @@ func (c *Config) IsFedramp() bool {
...
+	if !c.IsFedramp() {
+		return c.SnykCodeApi(), nil
+	}
+
+	// api url is fedramp, but code api url hasn't been updated properly
```
michelkaporin, line 895: When would that be a case? We should expect it to be updated (or do an update) by the time getter is called.

application/config/config.go
```diff
@@ -866,6 +884,28 @@ func (c *Config) IsFedramp() bool {
...
+			return "", err
+		}
+		c.SetSnykCodeApi(snykCodeApiUrl)
+	}
+
+	if c.Organization() == "" {
+		return "", errors.New("Organization must be present in a fedramp environment")
+	}
+
+	path := "/org/" + c.Organization() + "/code"
```
michelkaporin, line 908: Will that work for when both org slug and org id are provided?

PeterSchafer: `c.Organization()` will always return the Org ID, no matter what the user specified. So this is fine!

application/config/config.go
jozsef-armin-hamos: We do, but that is used by the sast_local_engine, and there is no error handling there, so I didn't want to introduce one. I thought the local_engine can use the original SnykCodeApi getter, and the backend_service will use the new one, that can also return an error.

application/config/config.go
```diff
@@ -54,14 +55,15 @@ import (
...
-	DefaultDeeproxyApiUrl = "https://deeproxy.snyk.io"
-	pathListSeparator     = string(os.PathListSeparator)
-	windows               = "windows"
+	deeproxyApiUrlKey        = "DEEPROXY_API_URL"
+	FormatHtml               = "html"
+	FormatMd                 = "md"
+	snykCodeTimeoutKey       = "SNYK_CODE_TIMEOUT" // timeout as duration (number + unit), e.g. 10m
+	DefaultSnykApiUrl        = "https://snyk.io/api"
+	DefaultDeeproxyApiUrl    = "https://deeproxy.snyk.io"
+	DefaultSnykCodeApiGovUrl = "https://api.snyk.io"
```
PeterSchafer, line 64: Question: This seems redundant to `DefaultSnykApiUrl`, do we really need this? Or would it suffice to change the existing variable?

application/config/config.go
michelkaporin: @PeterSchafer Thanks!

Can we document that in Code please? 🙏

PR #374, opened 2023-09-08
fix: shutdown & exit requests/notifications [HEAD-750]

bastiandoetsch:
### Description

Shutdown was stalling, as was exit. This makes the shutdown async and adds a test for shutdown

### Checklist

- [x] Linted

### Screenshots / GIFs

### DIFF
--- a/application/server/server.go
+++ b/application/server/server.go
@@ -18,7 +18,6 @@ package server
 
 import (
 	"context"
-	"fmt"
 	"os"
 	"runtime"
 	"strings"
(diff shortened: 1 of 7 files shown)


application/server/server.go
```diff
@@ -401,28 +398,36 @@ func monitorClientProcess(pid int) time.Duration {
...
-		di.ErrorReporter().FlushErrorReporting()
+		logger := c.Logger().With().Str("method", "Exit").Logger()
+		logger.Info().Msg("ENTERING")
+		logger.Info().Msg("Flushing error reporting...")
+		go di.ErrorReporter().FlushErrorReporting()
+		logger.Info().Msg("Stopping server...")
+		go srv.Stop()
+		time.Sleep(time.Second * 1)
+		logger.Info().Msg("Exiting...")
+		os.Exit(0)
```
PeterSchafer, line 430: Question: Do we really need to call os.Exit() here? Seems quite a strong move to me. It prevents running an Extension to the end which is not good, as it includes clean up tasks ...

bastiandoetsch: I was wondering about that... I'll change it.

PeterSchafer (approved):

PR #380, opened 2023-09-25
fix: display hovers on windows [HEAD-420]

bastiandoetsch:
### Description

the colon in `C:\...` was url-encoded in the URI representation by VSCode, but the go URI library does not url encode this bit.

This led to the hover cache having a different key to associate hovers to files, so they weren't displayed.

### Checklist

- [x] Linted

### Screenshots / GIFs

image


cat2608 (approved):
💯

PR #389, opened 2023-10-26
fix: handle empty result from FindStringSubmatch [HEAD-961]

PeterSchafer:
### Description

Handle the special case when there is no match.

### Checklist

- [ ] Linted


michelkaporin (approved):
It's good to accompany every fix PR with a test that ensures the bug doesn't happen again 😉

bastiandoetsch (approved):

PR #392, opened 2023-10-31
fix: conditionally set analytics flag [HEAD-879]

cat2608:
### Description

On the second time `writeSettings` is called, it doesn't get `enableAnalytics` flag. This change avoid this flag to be overridden after initialisation

### DIFF
--- a/application/config/config.go
+++ b/application/config/config.go
@@ -415,7 +415,7 @@ func (c *Config) SetSeverityFilter(severityFilter lsp.SeverityFilter) bool {
 	}
 
 	filterModified := c.filterSeverity != severityFilter
-	log.Debug().Str("method", "SetSeverityFilter").Msgf("Setting severity filter: %v", severityFilter)
+	log.Debug().Str("method", "SetSeverityFilter").Interface("severityFilter", severityFilter).Msg("Setting severity filter:")
 	c.filterSeverity = severityFilter
 	return filterModified
 }
--- a/application/server/configuration.go
+++ b/application/server/configuration.go
@@ -420,7 +420,7 @@ func updateProductEnablement(settings lsp.Settings) {
 }
 
 func updateSeverityFilter(s lsp.SeverityFilter) {
-	log.Debug().Str("method", "updateSeverityFilter").Msgf("Updating severity filter: %v", s)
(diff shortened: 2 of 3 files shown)


PeterSchafer (dismissed):
Please review/adapt the unit test associated with the changed logic

application/config/config.go
```diff
@@ -415,7 +415,7 @@ func (c *Config) SetSeverityFilter(severityFilter lsp.SeverityFilter) bool {
 	}
 
 	filterModified := c.filterSeverity != severityFilter
-	log.Debug().Str("method", "SetSeverityFilter").Msgf("Setting severity filter: %v", severityFilter)
+	log.Debug().Str("method", "SetSeverityFilter").Interface("severityFilter", severityFilter).Msg("Setting severity filter:")
```
cat2608, line 418: |Before|After|
|-|-|
|before|after|

application/server/configuration.go
```diff
@@ -145,7 +145,10 @@ func writeSettings(settings lsp.Settings, initialize bool) {
 	updateRuntimeInfo(settings)
 	updateAutoScan(settings)
 	updateSnykLearnCodeActions(settings)
-	config.CurrentConfig().SetAnalyticsEnabled(settings.EnableAnalytics)
+
+	if initialize {
```
bastiandoetsch, line 149: nice find!

application/server/configuration_test.go
```diff
@@ -283,7 +283,7 @@ func Test_UpdateSettings(t *testing.T) {
 			ScanningMode:                "manual",
 			AuthenticationMethod:        lsp.OAuthAuthentication,
 			SnykCodeApi:                 sampleSettings.SnykCodeApi,
-			EnableAnalytics:             true,
+			EnableAnalytics:             false, // when updating settings, this is always false [HEAD-975]
```
bastiandoetsch, line 286: do we have a test for initialization, then?

bastiandoetsch (approved):

application/server/configuration_test.go
cat2608: I added a new unit: `"analytics is set to true"`

PR #393, opened 2023-10-31
fix: incorrect issue cache state [HEAD-972]

PeterSchafer:
### Description

In certain constellations, the displayed number of issues varied over time even if the existing issues didn't change. There was a bug in a library used for caching issues, which triggered this bug. The PR updates the library in the affected use case.

Since this is a timing issue, it is difficult to provide a test to prove the fix.

### Checklist

- [ ] Linted

### DIFF
--- a/domain/ide/workspace/folder.go
+++ b/domain/ide/workspace/folder.go
@@ -24,7 +24,7 @@ import (
 	"strings"
 	"sync"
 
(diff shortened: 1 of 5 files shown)


bastiandoetsch (approved):

bastiandoetsch:
Commit signing needs a squash :)

PR #404, opened 2023-11-10
feat: redact token in logging [HEAD-1019]

bastiandoetsch:
### Description

We don't want to output the token in LS, so we now direct all log output through a filter/scrubber

### Checklist

- [x] Linted

### Screenshots / GIFs


cat2608 (approved):
🙏 Thank you

PR #408, opened 2023-11-17
fix: check pointer to avoid crash on failed HTTP requests

PeterSchafer:
### Description

if a request fails, the changed code didn't check the response pointer, which caused a crash of the application.

### Checklist

- [x] Linted

### DIFF
--- a/infrastructure/code/backend_service.go
+++ b/infrastructure/code/backend_service.go
@@ -180,6 +180,10 @@ func (s *SnykCodeHTTPClient) doCall(ctx context.Context,
 				Str("responseBody", string(responseBody)).
 				Str("snyk-request-id", requestId).
 				Msg("RECEIVED FROM REMOTE")
+		} else {
+			log.Trace().
+				Str("snyk-request-id", requestId).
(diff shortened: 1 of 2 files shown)


infrastructure/code/backend_service.go
```diff
@@ -174,10 +174,13 @@ func (s *SnykCodeHTTPClient) doCall(ctx context.Context,
 
 		response, body, err := s.httpCall(req)
 		responseBody = body
-		log.Trace().Str("response.Status", response.Status).
-			Str("responseBody", string(responseBody)).
-			Str("snyk-request-id", requestId).
-			Msg("RECEIVED FROM REMOTE")
+
+		if response != nil && responseBody != nil {
+			log.Trace().Str("response.Status", response.Status).
```
michelkaporin, line 179: Would probably make sense to log `RECEIVED FROM REMOTE` in that case anyway as it'd help with debugging.

michelkaporin (approved):

infrastructure/code/backend_service.go
PeterSchafer: Done!

PR #424, opened 2024-01-03
refactor: adopt log scrubbing from framework

thisislawatts:
### Description

Switch to using logging scrubber which was moved to the [framework](https://github.com/snyk/go-application-framework) for reuse across a few components.

### Checklist

- [X] Tests added and all succeed
- [X] Linted


bastiandoetsch (approved):
LGTM 🎉

PR #436, opened 2024-01-22
chore: enable smoke tests for Standard environment [IDE-56]

cat2608:
### Description

This PR enables:

- `Test_SmokeWorkspaceScanOssAndCode`
- `Test_SmokeWorkspaceScanIacAndCode`
- `Test_SmokeWorkspaceScanWithTwoUploadBatches`

for the Standard environment

### Checklist

- [ ] Linted

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,6 +82,10 @@ instance-test:
 	@echo "==> Running instance tests with proxy"
 	@export SMOKE_TESTS=1 && cd application/server && go test -run Test_SmokeWorkspaceScanOssAndCode && cd -
 
+smoke-test-all:
+	@echo "==> Running all smoke tests with proxy"
+	@export SMOKE_TESTS=1 && cd application/server && go test -run Test_Smoke && cd -
+
 ## build: Build binary for default local system's OS and architecture.
 .PHONY: build
 build:
--- a/.github/workflows/instance-tests.yaml
+++ b/.github/workflows/instance-tests.yaml
@@ -17,16 +17,16 @@ on:
   pull_request:
   workflow_dispatch:
   schedule:
-    - cron: '3 * * * *'  # hourly
+    - cron: "3 * * * *" # hourly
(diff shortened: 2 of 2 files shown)


acke (approved):
Look great to me!

Makefile
```diff
@@ -82,6 +82,10 @@ instance-test:
 	@echo "==> Running instance tests with proxy"
 	@export SMOKE_TESTS=1 && cd application/server && go test -run Test_SmokeWorkspaceScanOssAndCode && cd -
 
+smoke-test-all:
+	@echo "==> Running all smoke tests with proxy"
+	@export SMOKE_TESTS=1 && cd application/server && go test -run Test_Smoke && cd -
```
bastiandoetsch, line 87: This doesn't run all Smoke Tests, just the ones in application/server 🙂. I'd therefore suggest renaming the make target.

cat2608: @bastiandoetsch it makes sense :)

Do you have any suggestions? Looking at the test we currently have: "_Running instance tests with proxy_" (`instance-test`) which runs only 1 of 3 smoke tests, I'm trying to find a name for the change I introduced but I still can't figure it out since its differences are:

- it runs for `Standard` environment vs `M1` and `M3`,
- and it's invoking all the Smoke test that follow the naming convention `Test_Smoke...` instead of only `Test_SmokeWorkspaceScanOssAndCode`


Do you think that's a good approach `standard-smoke-tests` as the Makefile command and echo _Running Test_Smoke* tests with proxy_?


Here's a summary of the changes:

|Makefile CMD|Environments|Units|@echo
|-|-|-|-|
|`instance-test`|M1 and M3|Test_SmokeWorkspaceScanOssAndCode|_Running instance tests with proxy_|
|`standard-smoke-tests`|Standard|Test_Smoke*|_Running Test_Smoke* tests with proxy_

PR #437, opened 2024-01-23
feat: add Slack Notification When Smoke Test Fails

acke:
IDE-57

### Description

Adding Slack notification to #dx_local_alerts if the smoke test cases fail.

### Checklist

- [ ] Linted

### Screenshots / GIFs

### DIFF
--- a/.github/workflows/instance-tests.yaml
+++ b/.github/workflows/instance-tests.yaml
@@ -86,7 +86,5 @@ jobs:
         #if: ${{ failure() || cancelled() }}
         if: ${{ always()}}
         uses: slackapi/slack-github-action@v1.24
-        with:
-          status ${{job.status}}
         env:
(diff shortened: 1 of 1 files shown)


acke:
Adding Slack notification to #dx_local_alerts if the smoke test cases fail.

.github/workflows/instance-tests.yaml
```diff
@@ -81,3 +81,15 @@ jobs:
           SNYK_API: ${{ secrets.SNYK_API }}
         run: |
           make smoke-test-all
+          
+      - name: Slack Notification
+        if: ${{ failure() || cancelled() }}
+        #if: ${{ always()}}
```
bastiandoetsch, line 87: Maybe remove this? I think we don't need it anymore :)

PR #441, opened 2024-01-30
fix: Language Server analytics permitted function implemented

acke:
### Description

Since we are having more geographically distributed environments, we should only send analytics to the MT US environments, excluding FedRAMP.

### Checklist

- [x] Linted

### Screenshots / GIFs

https://snyksec.atlassian.net/browse/IDE-24


application/config/config.go
```diff
@@ -872,3 +875,16 @@ func (c *Config) IdeName() string {
 func (c *Config) IsFedramp() bool {
 	return c.Engine().GetConfiguration().GetBool(configuration.IS_FEDRAMP)
 }
+
+func (c *Config) IsAnalyticsPermitted() bool {
+	u, err := url.Parse(c.Engine().GetConfiguration().GetString(configuration.API_URL))
+
+	if err != nil {
+		log.Trace().Err(err).Msg("unable to parse configured API_URL")
```
bastiandoetsch, line 883: Please use `c.Logger().With()...`

application/config/config.go
```diff
@@ -872,3 +875,16 @@ func (c *Config) IdeName() string {
 func (c *Config) IsFedramp() bool {
 	return c.Engine().GetConfiguration().GetBool(configuration.IS_FEDRAMP)
 }
+
+func (c *Config) IsAnalyticsPermitted() bool {
+	u, err := url.Parse(c.Engine().GetConfiguration().GetString(configuration.API_URL))
```
bastiandoetsch, line 880: Initialization order may be interesting here - is the endpoint already set when IsAnalyticsPermitted is called?

acke: We went through the server's `initlialize` handler, and made sure that all calls to `IsAnalyticsPermitted` happen after the API URL has been set in the configuration 👍  `updateTelemetry` is the first function inside the `initialize` handler that is checking if analytics are permitted, and it's being called after setting the API URL. Further calls are done in the `initialized` handler, which also happens afterwards.

application/config/config.go
```diff
@@ -67,12 +67,16 @@ const (
...
-	currentConfig      *Config
-	mutex              = &sync.Mutex{}
-	LicenseInformation = "License information\n FILLED DURING BUILD"
+	Version                        = "SNAPSHOT"
+	LsProtocolVersion              = "development"
+	Development                    = "true"
+	currentConfig                  *Config
+	mutex                          = &sync.Mutex{}
+	LicenseInformation             = "License information\n FILLED DURING BUILD"
+	analyticsPermittedEnvironments = map[string]bool{
```
bastiandoetsch, line 76: Could be beneficial to store these in the Go Application Framework config (gafConfig). As this applies also to CLI.

bastiandoetsch (approved):
Nice work. Marked an optional todo - but can be merged.

PR #454, opened 2024-03-07
feat: add feature flag status request [IDE-171]

cat2608:
### Description

### Checklist

- [ ] Linted

### Screenshots / GIFs


infrastructure/snyk_api/snyk_api.go
```diff
@@ -78,28 +91,38 @@ func NewSnykApiClient(client func() *http.Client) SnykApiClient {
...
+	return response, err
+}
 
-	var response SastResponse
-	unmarshalErr := json.Unmarshal(responseBody, &response)
-	if unmarshalErr != nil {
-		fmtErr := fmt.Errorf("%v: %v", err, responseBody)
-		log.Err(fmtErr).Str("method", method).Msg("couldn't unmarshal SastResponse")
-		return SastResponse{}, err
+func (s *SnykApiClientImpl) FeatureFlagSettings(featureFlagType FeatureFlagType) (FFResponse, error) {
```
teodora-sandu, line 110: The argument should just be a string, since we don't want to pre-define all supported feature flags here. Or at least I don't think we do, I don't see a use for that. Let me know if I'm missing something!

infrastructure/snyk_api/snyk_api.go
```diff
@@ -78,28 +91,38 @@ func NewSnykApiClient(client func() *http.Client) SnykApiClient {
...
+	method := "FeatureFlagSettings"
+	var response FFResponse
+	log.Debug().Str("method", method).Msgf("API: Getting %s", featureFlagType.String())
+	path := path.Join("/cli-config/feature-flag/", string(featureFlagType))
+	organization := config.CurrentConfig().Organization()
+	if organization != "" {
+		path += "?org=" + url.QueryEscape(organization)
+	}
```
bastiandoetsch, line 118: I'm not sure, this parameter carries over to the feature flag endpoint, or how to specify the org here. It may be, that it uses the default org, we should confirm the behaviour.

infrastructure/snyk_api/snyk_api.go
```diff
@@ -141,10 +164,26 @@ func (s *SnykApiClientImpl) doCall(method string,
...
 	if r.StatusCode >= 200 && r.StatusCode <= 399 {
 		return nil
 	}
 
 	return NewSnykApiError("Unexpected response code: "+r.Status, r.StatusCode)
 }
+
+func (f FeatureFlagType) String() string {
+	return string(f)
+}
```
bastiandoetsch, line 189: Wondering if this is actually needed, or not.

infrastructure/snyk_api/snyk_api.go
cat2608: Form `/check` there is an ENUM defining the FF names so I decided to define a list for them here too for robustness.

I decided to go with the `FeatureFlagType` because we have a predefined `enum` that centralises the names of our feature flags in the repo defining the FFs 🙌

infrastructure/snyk_api/snyk_api.go
cat2608: I've added the test `"Get feature flag status when disabled for a ORG"` to document the behaviour of the API when called with an `org_id` which does not have the FF enabled.

For example, my default org does have `useWorkspaces` enabled but the org `hammerhead` doesn't and it gets the following message:

```json
{
  "ok": false,
  "userMessage": "Org hammerhead doesn't have 'use-workspaces' feature enabled"
}
```

This response can be retrieved by calling: `/v1/cli-config/feature-flags/useWorkspaces?org=org_id`

infrastructure/snyk_api/snyk_api.go
cat2608: I saw this pattern in the policies repo but can see now that for our case, it's unnecessary.
I'm removing it 🙏

infrastructure/snyk_api/fake_api_client.go
```diff
@@ -28,12 +28,30 @@ type FakeApiClient struct {
...
+func (f *FakeApiClient) addCallForMethod(method string, args []any) {
+	mutex.Lock()
+	defer mutex.Unlock()
+
+	if f.Calls == nil {
+		f.Calls = make(map[string][][]any)
+	}
+	f.Calls[method] = append(f.Calls[method], args)
+}
+
```
cat2608, line 54: Added `addCallForMethod` to extend the testing for **Feature Flags**, without impacting existing consumers of the `FakeApiClient`. 

The current `addCall` is utilised across various tests and extending it to accommodate the new API call impact other consumers.

We could merge `addCall` with `addCallForMethod` into a single, more generic method in a future PR.

bastiandoetsch (approved):

bastiandoetsch:
LGTM! 🎉

PR #455, opened 2024-03-08
feat: use code-client-go to create and extend bundle [IDE-175]

teodora-sandu:
### Description

Starts using the `code-client-go` library for uploading and analyzing Snyk Code files with ignores. Updates to the latest `code-client-go` as a result and includes the bundle creation/extension of the files in the workspace.

Tested in IntelliJ by running `make build-debug` in `snyk-ls` and `make build` in the CLI , then changing the path to the CLI in IntelliJ. I enabled the `snykCodeConsistentIgnores` feature flag and the `snyk.preview.snyk.code.ls.enabled` registry flag, then opened IntelliJ on https://github.com/snyk/target-service

Also added smoke tests, which I was able to run in my IDE. See screenshot for proof of passing. The token was a service account token created for our new test org.

### Checklist

- [x] Linted

### Screenshots / GIFs

Screenshot 2024-03-15 at 15 22 47

Screenshot 2024-03-25 at 17 19 02


cat2608:
> I'm not sure why, maybe the tests don't run on the branch after all?

@teodora-sandu you can run manually the Instance Test workflow for your branch. Go to https://github.com/snyk/snyk-ls/actions/workflows/instance-tests.yaml and select your branch in the dropdown I hope it helps :)

application/server/server_test.go
```diff
@@ -964,28 +964,81 @@ func Test_CodeActionResolve_ShouldExecuteCommands(t *testing.T) {
 	assert.Equal(t, expected, serviceMock.ExecutedCommands()[0].CommandId)
 }
 
-func Test_SmokeWorkspaceScanOssAndCode(t *testing.T) {
+func Test_SmokeWorkspaceScan(t *testing.T) {
```
bastiandoetsch, line 967: While we're refactoring... maybe we can move this to another file?

infrastructure/code/code.go
```diff
@@ -194,7 +192,7 @@ func (sc *Scanner) Scan(ctx context.Context, path string, folderPath string) (is
 
 	var results []snyk.Issue
 	if sc.useIgnoresFlow() {
-		results, err = sc.UploadAndAnalyzeWithIgnores(span.Context(), folderPath)
+		results, err = sc.UploadAndAnalyzeWithIgnores(span.Context(), folderPath, files, startTime, changedFiles)
```
bastiandoetsch, line 195: This startTime parameter... can we implement this differently? I don't think we should have it in the signature. Also, looking at the code, it is wrong to exclude collecting the files, IMHO.

infrastructure/code/code.go
```diff
@@ -302,33 +300,49 @@ func (sc *Scanner) UploadAndAnalyze(ctx context.Context,
...
+	scanStartTime time.Time,
+	changedFiles map[string]bool,
 ) (issues []snyk.Issue, err error) {
-	response, err := codeClient.UploadAndAnalyze()
+	if scanStartTime.IsZero() {
+		scanStartTime = time.Now()
+	}
+
+	scanMetrics := codeClientObservability.NewScanMetrics(scanStartTime, 0)
+	c := config.CurrentConfig()
```
bastiandoetsch, line 314: while we are doing new signatures anyway, I'd inject the config into the function. This is something that would beneficial everywhere (but is a lot of work):
- easier to configure tests
- better for parallel processing as we don't work with global state, instead of session state

infrastructure/code/code_analytics.go
bastiandoetsch, line 1: Worth checking back with @Arvi3d, if we can skip the whole analytics / Amplitude stuff.

infrastructure/code/code_error_reporter.go
bastiandoetsch, line 1: Wondering if this could be just a wrapper around the existing sentry error reporter.

infrastructure/code/issue_enhancer.go
```diff
@@ -253,7 +252,7 @@ func (b *IssueEnhancer) createOpenSnykLearnCodeAction(issue snyk.Issue) (ca *sny
 	lesson, err := b.learnService.GetLesson(issue.Ecosystem, issue.ID, issue.CWEs, issue.CVEs, issue.IssueType)
 	if err != nil {
 		log.Err(err).Msg("failed to get lesson")
-		b.errorReporter.CaptureError(err)
+		b.errorReporter.CaptureError(err, codeClientObservability.ErrorReporterOptions{ErrorDiagnosticPath: ""})
```
bastiandoetsch, line 255: I really like that Options solution here 👍

internal/testutil/test_setup.go
```diff
@@ -34,12 +34,12 @@ const (
 
 func IntegTest(t *testing.T) *config.Config {
 	t.Helper()
-	return prepareTestHelper(t, integTestEnvVar)
+	return prepareTestHelper(t, integTestEnvVar, false)
 }
 
-func SmokeTest(t *testing.T) *config.Config {
+func SmokeTest(t *testing.T, useConsistentIgnores bool) *config.Config {
```
bastiandoetsch, line 40: After rollout, we should remove that extra config.

infrastructure/code/code.go
teodora-sandu: I've had another look at this and we could keep a similar signature to the existing `UploadAndAnalyze` function, so we don't pass the startTime as a parameter. 

In terms of us excluding collecting the files. I agree but the current code excludes them and I'd like to use these metrics to compare how the two flows perform so I don't want to change how they work

infrastructure/code/code_analytics.go
teodora-sandu: https://snyk.slack.com/archives/C06LTKLFY2F/p1711363049600899

teodora-sandu: We can remove it so I will do that in https://github.com/snyk/code-client-go/pull/15 first

infrastructure/code/code.go
teodora-sandu: Actually, now that we've removed the Amplitude analytics this isn't even needed anymore

infrastructure/code/code.go
teodora-sandu: Instead of injecting in the function I ended up refactoring the code in https://github.com/snyk/code-client-go/pull/16 because we were passing down the host too much. This was inspired by the work done in https://github.com/snyk/code-client-go/pull/14

cat2608 (approved):
Tested it locally with IntelliJ and FF enabled

PR #456, opened 2024-03-08
refactor: extract issue enhancer funtionality [IDE-175]

teodora-sandu:
### Description

This part of the code will soon have some changes so that we can call off to the new Snyk Code scanning functionality in `code-client-go`. `code-client-go` will return SARIF and inside `snyk-ls` we are responsible for converting the SARIF to the `snyk.Issue` type used in the Language Server and enhance that issue with code actions etc. At the moment this functionality to enhance issues is done within the `Bundle` type, but it makes more sense to extract it into a new `IssueEnhancer` type (name is up for debate) and so then we can use this functionality both in the existing flow and then when we start using `code-client-go`. 

I've come to this conclusion by working on https://github.com/snyk/code-client-go/pull/2 and https://github.com/snyk/snyk-ls/pull/455 but these two PRs were getting too big to ask for reviews. Hopefully this PR makes sense but if not, we can chat about it.

I've tested it by running `make build-debug` and then running the `vscode-extension` by pointing at this binary. The UI still works as expected. 

### Checklist

- [x] Linted

### Screenshots / GIFs

Screenshot 2024-03-08 at 14 30 30

### DIFF
--- a/infrastructure/code/code.go
+++ b/infrastructure/code/code.go
@@ -378,15 +378,14 @@ func (sc *Scanner) createBundle(ctx context.Context,
 		errorReporter: sc.errorReporter,
 		limitToFiles:  limitToFiles,
 		notifier:      sc.notifier,
-		issueEnhancer: IssueEnhancer{
-			SnykCode:      sc.BundleUploader.SnykCode,
-			instrumentor:  sc.BundleUploader.instrumentor,
-			errorReporter: sc.errorReporter,
-			learnService:  sc.learnService,
-			notifier:      sc.notifier,
-			requestId:     requestId,
-			rootPath:      rootPath,
-		},
+		issueEnhancer: newIssueEnhancer(
+			sc.BundleUploader.SnykCode,
+			sc.BundleUploader.instrumentor,
+			sc.errorReporter,
+			sc.notifier,
+			sc.learnService,
+			requestId,
+			rootPath),
 	}
 	var err error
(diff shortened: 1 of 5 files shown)


infrastructure/code/code.go
```diff
@@ -377,8 +377,16 @@ func (sc *Scanner) createBundle(ctx context.Context,
 		rootPath:      rootPath,
 		errorReporter: sc.errorReporter,
 		limitToFiles:  limitToFiles,
-		learnService:  sc.learnService,
 		notifier:      sc.notifier,
+		issueEnhancer: IssueEnhancer{
```
bastiandoetsch, line 381: suggestion: Better to use a dedicated constructor (New), I think.

infrastructure/code/code.go
```diff
@@ -377,8 +377,16 @@ func (sc *Scanner) createBundle(ctx context.Context,
...
+			SnykCode:      sc.BundleUploader.SnykCode,
+			instrumentor:  sc.BundleUploader.instrumentor,
+			errorReporter: sc.errorReporter,
+			learnService:  sc.learnService,
+			notifier:      sc.notifier,
+			requestId:     requestId,
```
bastiandoetsch, line 387: RequestID is for one Snyk Code scan, issueEnhancer potentially for a session. How do we treat updates to the requestID, or asked differently - can we make the requestID transient?

bastiandoetsch (approved):
Love it!

PR #457, opened 2024-03-08
refactor: refactor the converter to use types from code-client-go [IDE-175]

teodora-sandu:
### Description
The same exact `SarifResponse` type is now in `code-client-go`. This type is needed for the new Snyk Code scanning functionality which will be behind a feature flag but in order to use it in `snyk-ls` we'll need some conversion functions that generate `snyk.Issue` types. Rather than duplicate this code, I thought we could change it so that both the existing and old flow use this new type. 

In order to do that, I had to change the conversion functions so they take in the element they're trying to convert. And I've exposed the functionality via a new `SarifConverter`, since there are other conversion functions in this file that are not related. Hopefully this makes it easier to tell what belongs to what.

I've tested it by running `make build-debug` and then running the `vscode-extension` by pointing at this binary. The UI still works as expected.
### Checklist

- [x] Linted

### Screenshots / GIFs

Screenshot 2024-03-08 at 14 30 30


cat2608 (approved):
Thank you 🙌

PR #461, opened 2024-03-11
feat: start branching off to new flow for snyk code ignores [IDE-172]

teodora-sandu:
### Description

In https://snyksec.atlassian.net/browse/IDE-172 we want to start using `code-client-go` behind the feature flag to branch off to the new flow and be able to manipulate ignored findings in our IDEs. `code-client-go` is currently not open sourced yet and we don't have a defined contract yet for what the data we return will look like, so for now this PR adds a separate flow for the Snyk Code scanning capabilities with ignores, which returns fake issues with fake ignore metadata. 

This PR will then unblock us working on the UI while the backend is sorted out.

Therefore, in this PR we are defining where the ignore information gets stored in the `snyk.Issue` type used by the Language Server.

I've run `make build-debug` and tested VSCode by changing the location of the LS binary.

I haven't tested it with IntelliJ because of problems with building the CLI.
### Checklist

- [x] Linted

### DIFF
--- a/infrastructure/code/code.go
+++ b/infrastructure/code/code.go
@@ -509,7 +509,7 @@ func (sc *Scanner) trackResult(success bool, scanMetrics *ScanMetrics) {
 }
 
 func (sc *Scanner) useIgnoresFlow() bool {
-	response, err := sc.SnykApiClient.FeatureFlagSettings(snyk_api.FeatureFlagSnykCodeConsistentIgnores)
+	response, err := sc.SnykApiClient.FeatureFlagStatus(snyk_api.FeatureFlagSnykCodeConsistentIgnores)
 	if err != nil {
 		log.Debug().Err(err).Msg("Failed to check if the ignores experience is enabled")
 		return false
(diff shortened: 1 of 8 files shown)


infrastructure/code/code.go
```diff
@@ -418,3 +507,15 @@ func (sc *Scanner) trackResult(success bool, scanMetrics *ScanMetrics) {
...
+	response, err := sc.SnykApiClient.FeatureFlagSettings(snyk_api.FeatureFlagSnykCodeConsistentIgnores)
+	if err != nil {
+		log.Debug().Err(err).Msg("Failed to check if the ignores experience is enabled")
+		return false
+	}
+	if !response.Ok && response.UserMessage != nil {
+		log.Info().Msg(*response.UserMessage)
+	}
+	return response.Ok
+}
```
cat2608, line 521: Just for discussion: `FeatureFlagSettings` wasn't a great choice of name (my bad 😢)
I think the `Settings` part does not make sense. Better if `FeatureFlagStatus`?

teodora-sandu: I'm not very good at naming stuff but I guess you're right, `FeatureFlagStatus` is slightly better! I've renamed it

PR #464, opened 2024-03-12
feat: add ignore details [IDE-176]

teodora-sandu:
### Description

Starts using `code-client-go` after https://github.com/snyk/code-client-go/pull/8 gets merged. Also prepares the data for ignored findings to be consumed by IDEs. 

Test it by:
- uncommenting the replace directive, 
- running `make build-debug`,
- adding a replace directive for `snyk-ls` and `code-client-go` in the CLI in the `cliv2` folder
- running `make build` from the CLI 
- running the IntelliJ plugin in debug mode

### Checklist

- [x] Linted

### Screenshots / GIFs

Screenshot 2024-03-12 at 19 12 30

### DIFF
--- a/infrastructure/code/convert.go
+++ b/infrastructure/code/convert.go
@@ -28,6 +28,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/hexops/gotextdiff"
 	"github.com/hexops/gotextdiff/myers"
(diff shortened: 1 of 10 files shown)


infrastructure/code/convert.go
```diff
@@ -391,6 +391,21 @@ func (s *SarifConverter) toIssues(baseDir string) (issues []snyk.Issue, err erro
 				CWEs:                testRule.Properties.Cwe,
 			}
 
+			// this can be an array of multiple suppressions in SARIF
+			// but we only store one ignore for now
```
bastiandoetsch, line 395: Why?

teodora-sandu: There can only be one ignore per finding, as far as I know. I'm not sure what would happen though if an ignore expires and the user wants to "renew" it - do they create another on top of it but only one is "active" or would it just get replaced. Maybe I should ask

teodora-sandu: https://snyk.slack.com/archives/C06MZKWG0NB/p1710328751971279?thread_ts=1710150868.607839&cid=C06MZKWG0NB

bastiandoetsch (commented):
Looks good so far, but I think the tests need some love :)

infrastructure/code/convert.go
```diff
@@ -391,12 +392,46 @@ func (s *SarifConverter) toIssues(baseDir string) (issues []snyk.Issue, err erro
...
 	return issues, errs
 }
 
+func (s *SarifConverter) getIgnoreDetails(result codeClient.Result) (bool, *snyk.IgnoreDetails) {
+	isIgnored := false
+	var ignoreDetails *snyk.IgnoreDetails
+
+	// this can be an array of multiple suppressions in SARIF
+	// but we only store one ignore for now
+	if len(result.Suppressions) == 1 {
```
bastiandoetsch, line 408: Hm... for resiliency, shouldn't this be `> 0` and then only take the first element for ignoring, maybe logging that there are more?

teodora-sandu: That's a good idea!

bastiandoetsch (approved):
Nice!

PR #467, opened 2024-03-15
fix: update code-client-go [IDE-172]

teodora-sandu:
### Description

`code-client-go` had a small problem in `go.mod` where the minor version and toolchain were supplied, causing `snyk-ls` to not be able to be used in the CLI.

We fixed this in https://github.com/snyk/code-client-go/pull/11 (which also involved removing the circular dependency `code-client-go`->`sny-ls`->`code-client-go`) so now updating the version, so that eventually we can merge https://github.com/snyk/cli/pull/5100.


acke (approved):
Looks good to me.

PR #469, opened 2024-03-18
chore: exclude notifications for `push` events

cat2608:
### Description

- Previously, a Slack notification was sent whenever a build failed or was cancelled, regardless of the event type. This included push events to working branches
- This change will reduce the number of Slack notifications, particularly for ongoing work in development branches.
- Failures in more critical scenarios, like pull requests or main branch updates, will continue to trigger Slack notifications

### Checklist

- [ ] Linted

### Screenshots / GIFs

### DIFF
--- a/.github/workflows/instance-tests.yaml
+++ b/.github/workflows/instance-tests.yaml
@@ -83,7 +83,7 @@ jobs:
           make instance-standard-test
 
       - name: Slack Notification
-        if: ${{ github.event_name != 'push' && (failure() || cancelled()) }}
+        if: ${{ (github.event_name != 'push' || github.ref == 'refs/heads/main') && (failure() || cancelled()) }}
         with:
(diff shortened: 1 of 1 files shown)


bastiandoetsch (approved):

.github/workflows/instance-tests.yaml
```diff
@@ -83,7 +83,7 @@ jobs:
           make instance-standard-test
 
       - name: Slack Notification
-        if: ${{ failure() || cancelled() }}
+        if: ${{ github.event_name != 'push' && (failure() || cancelled()) }}
```
cmars, line 86: Hmm... would we want to notify on a push to `main`? I think we might be wanting to exclude non-main branches here. Unless there's context I'm missing?

cat2608: @cmars , I have edited the condition. I hope this match pushes to `main`
Thanks :)

PR #479, opened 2024-04-03
feat: add Code Issue panel styling [IDE-215]

cat2608:
### Description

This change add the `<style />` for Code Issue panel. There are some adjustments to be done. They will get addressed in a follow up PR

### Checklist

- [ ] Linted

intellij2html


infrastructure/code/code_html.go
```diff
@@ -46,27 +47,29 @@ func getDataFlowHeadingHtml(issue snyk.CodeIssueData) string {
...
-		<div class="data-flow-row">
-		  <span class="data-flow-number">%d</span>
-		  <span class="data-flow-blank"> </span>
-		  <span class="data-flow-filepath" file-path="%s" start-line="%d" end-line="%d" start-character="%d" end-character="%d">%s:%d</span>
-		  <span class="data-flow-delimiter">|</span>
-		  <span class="data-flow-text">%s</span>
-		</div>`,
+		  <tr class="data-flow-row">
+		    <td class="data-flow-number">%d</td>
+		    <td class="data-flow-clickable-row" file-path="%s" start-line="%d" end-line="%d" start-character="%d" end-character="%d">%s:%d</td>
```
teodora-sandu, line 57: If you change the class here we will need to change the logic in https://github.com/snyk/snyk-intellij-plugin/pull/497 a bit so we're getting the right class. Just FYI, depends who merges their PR first 😆

infrastructure/code/template/details.html
```diff
@@ -46,18 +297,18 @@
...
       <div class="data-flow-body">
         ${dataFlow}
       </div>
     </article>
 
     <!-- External example fixes -->
-    <article class="external-example-fixes">
-      <h2>External example fixes</h2>
-      <p id="info-top" class="font-light">
+    <article class="example-fixes">
```
teodora-sandu, line 307: For the example fixes, how do we plan on swapping between the tabs? I think it makes more sense to add the JavaScript in here (`snyk-ls`) for changing what div is visible since it's mostly styling that will take care of that. Also, the behaviour of the tabs would stay the same regardless of IDE.

infrastructure/code/template/details.html
```diff
@@ -22,6 +22,257 @@
   <meta charset="UTF-8">
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <style>
+    html {
```
teodora-sandu, line 25: Not blocking but I did notice that the scrolling is not as easy as it used to be. I need to use the scrollbar to scroll whereas before I could use my touchpad

infrastructure/code/template/details.html
```diff
@@ -22,6 +22,257 @@
...
+      box-sizing: border-box;
+    }
+
+    body {
+      height: 100%;
+      width: 100%;
+      margin: 0;
+      padding: 0;
+      overflow: hidden;
+      color: #d4d4d4;
```
teodora-sandu, line 42: A question about the colours. Did Design end up using intelliJ theme colours so that when we swap between light/dark and various different themes we can maintain the nice look of the plugin?

infrastructure/code/code_html.go
teodora-sandu: I'll do it in https://github.com/snyk/snyk-intellij-plugin/pull/497

infrastructure/code/template/details.html
teodora-sandu: We agreed I could do it in a follow-up PR as part of the ticket I'm working on

infrastructure/code/template/details.html
cat2608: If it's okay with you, this topic and the scrolling issue you pointed out before will be address when meeting the designer 🏗️

teodora-sandu (approved):

PR #481, opened 2024-04-04
fix: improve resiliency of CLI download [IDE-90]

bastiandoetsch:
### Description

This changes the download of the CLI in a minor way:
- it does not search for a CLI in the system path anymore and chooses an arbitrary file that is called snyk
- it downloads the CLI in a directory below the actual destination directory, so chances are high, it can finish once it's begun
- it creates the necessary directory tree before trying to move

### Checklist

- [x] Linted

### Screenshots / GIFs


acke (approved):
Looks good to me.

PR #488, opened 2024-04-11
refactor: use `html/template`

cat2608:
### Description

🚀  This PR shifts from in-code string concatenation to use Go's `html/template` package for generating the Code Issue panel.

- **Centralized HTML Generation**: All HTML content is now generated via templates. This approach enhances maintainability and readability of the HTML structure within the codebase.
- **Conditional Rendering**: Implemented conditionals within the templates to handle edge cases, such as the absence of CWEs or example fixes.

### Checklist

- [x] Linted

### Screenshots / GIFs
In the following clip we can check:

1. ℹ️ Issue is not ignored
2. ⚠️ Issue is ignored: badge, warning and ignore details are shown
3. 💅 Styled scrollbar

https://github.com/snyk/snyk-ls/assets/1948377/21ff29ab-4d1a-40cb-8d31-aed4d5ebf96c


infrastructure/code/code_html.go
```diff
@@ -30,124 +31,207 @@ import (
...
+	</tbody></table>`
+
+	html, err := templateHelper(tmpl, "dataFlow", items)
+	if err != nil {
+		log.Error().Msg(err.Error())
+		return ""
+	}
+
+	return html
+}
```
cat2608, line 186: Tabs vs Spaces... 🥲

infrastructure/code/code_html.go
```diff
@@ -30,172 +31,157 @@ import (
...
+	StartLineValue int
+}
+
+type ExampleCommit struct {
+	CommitURL string
+	IconSVG   template.HTML
+	RepoName  string
+	DiffHTML  template.HTML
+}
+
```
bastiandoetsch, line 57: There is already structs for DataFlowElements and ExampleCommits. Should we just extend them? Or doesn't it make sense?

infrastructure/code/code_html.go
```diff
@@ -30,172 +31,157 @@ import (
...
-	dataFlowCount := len(issue.DataFlow)
-	stepWord := "step"
+func init() {
+	funcMap := template.FuncMap{
+		"vendorIconSvg": getVendorIconSvg,
+		"repoName":      getRepoName,
+		"codeDiffHtml":  getCodeDiffHtml,
+		"trimCWEPrefix": trimCWEPrefix,
+		"idxMinusOne":   idxMinusOne,
+	}
```
bastiandoetsch, line 70: so tidy 😍

infrastructure/code/code_html.go
```diff
@@ -30,172 +31,157 @@ import (
...
 
-	if dataFlowCount > 1 {
-		stepWord += "s"
+	var err error
+	globalTemplate, err = template.New("code").Funcs(funcMap).Parse(detailsHtmlTemplate)
```
bastiandoetsch, line 73: Should we take the product name? E.g. `product.Code`?

infrastructure/code/code_html.go
```diff
@@ -30,172 +31,157 @@ import (
...
+	for i, flow := range issue.DataFlow {
+		items = append(items, DataFlowItem{
+			Number:         i + 1,
+			FilePath:       flow.FilePath,
+			StartLine:      flow.FlowRange.Start.Line,
+			EndLine:        flow.FlowRange.End.Line,
+			StartCharacter: flow.FlowRange.Start.Character,
+			EndCharacter:   flow.FlowRange.End.Character,
+			FileName:       filepath.Base(flow.FilePath),
+			Content:        flow.Content,
```
bastiandoetsch, line 150: You'll need a left trim here :)

bastiandoetsch:
Looking forward to screenshots :)

infrastructure/code/code_html.go
cat2608: The idea is that I'm preparing the data that will be injected in the template:

```html
<tbody>
  {{range .DataFlowTable}}
  <tr class="data-flow-row">
    <td class="data-flow-number">{{.Number}}</td>
    <td class="data-flow-clickable-row" file-path="{{.FilePath}}" start-line="{{.StartLine}}"
      end-line="{{.EndLine}}" start-character="{{.StartCharacter}}" end-character="{{.EndCharacter}}">
      {{.FileName}}:{{.StartLineValue}}
    </td>
    <td class="data-flow-delimiter">|</td>
    <td class="data-flow-text">{{.Content}}</td>
  </tr>
  {{end}}
</tbody>
```

and the local `struct` for `DataFlowItem` is specific for that template:
```go
DataFlowItem{
	Number:         i + 1,
	FilePath:       flow.FilePath,
	StartLine:      flow.FlowRange.Start.Line,
	EndLine:        flow.FlowRange.End.Line,
	StartCharacter: flow.FlowRange.Start.Character,
	EndCharacter:   flow.FlowRange.End.Character,
	FileName:       filepath.Base(flow.FilePath),
	Content:        flow.Content,
	StartLineValue: flow.FlowRange.Start.Line + 1,
}
```

Does this make sense? Was it what you meat?

infrastructure/code/code_html.go
cat2608: Do you mean replacing the hardcoded string "code" with `product.ProductCode` which is defined in the package `product`?

```go
const (
	ProductOpenSource           Product = "Snyk Open Source"
	ProductCode                 Product = "Snyk Code"
	ProductInfrastructureAsCode Product = "Snyk IaC"
	ProductContainer            Product = "Snyk Container"
	ProductUnknown              Product = ""
)
```

The string for the `New()` function is relevant when we are dealing with more than one template. Currently we have only one which represents the Code Issue template which was organised by `<articles />` to keep some semantics when building the layout. 

I'm not sure if it's worth importing the package `product` to get the string `"Snyk Code"`:

```go
template.New(string(product.ProductCode))
```

WDYT?

infrastructure/code/code_html.go
cat2608: I have tried the template with the same issue reported by @Arvi3d and it seems that the rendering is taking care of the white space.

no-trim

infrastructure/code/template/details.html
```diff
@@ -308,7 +329,7 @@
 
 <body>
   <!-- Ignore header -->
-  <div class="ignore-warning-wrapper ${visibilityClass}">
+  <div class="ignore-warning-wrapper {{.VisibilityClass}}">
```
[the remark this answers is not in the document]
cat2608: Yes! Thank you! I'll make the change. 🙌

infrastructure/code/code_html.go
bastiandoetsch: I was wondering if we need the additional type. On the other hand, it doesn't hurt to have it.

infrastructure/code/code_html.go
bastiandoetsch: It's ok, to import the package, I think. We should not run into dependencies, and using a specific constant is in my opinion cleaner. The alternative would be to define a constant, but I believe using the product.ProductCode constant would be clearer, as we then also have the reference that ties the template directly to the product.

bastiandoetsch (approved):
This looks great (and even reduces the lines of code)! Thanks for the good job!

infrastructure/code/template/details.html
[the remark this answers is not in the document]
cat2608: @jen-snyk Fixed! Way better now!!! Thank you 🎉

PR #492, opened 2024-04-17
feat: add cache cleanups, and cache removal handling

bastiandoetsch:
### Description

### Checklist

- [x] Linted

### Screenshots / GIFs


domain/ide/workspace/folder.go
```diff
@@ -182,14 +218,14 @@ func (f *Folder) scan(ctx context.Context, path string) {
...
-		return
-	}
+	//issuesSlice := f.IssuesForFile(path)
+	//if issuesSlice != nil {
+	//	log.Info().Str("method", method).
+	//		Int("issueSliceLength", len(issuesSlice)).
+	//		Msgf("Cached results found: Skipping scan for %s", path)
+	//	f.processResults(snyk.ScanData{Issues: issuesSlice})
+	//	return
+	//}
```
bastiandoetsch, line 228: I don't see the logic or necessity to skip a scan, when we already have data. WDYT?

domain/ide/workspace/folder_test.go
```diff
@@ -46,21 +47,6 @@ import (
...
-	scanner.AddTestIssue(NewMockIssue("1", filePath))
-	f := NewFolder(folderPath, "Test", scanner, hover.NewFakeHoverService(), snyk.NewMockScanNotifier(), notification.NewNotifier())
-	ctx := context.Background()
-
-	f.ScanFile(ctx, filePath)
-	f.ScanFile(ctx, filePath)
-
-	assert.Equal(t, 1, scanner.Calls())
-}
-
```
bastiandoetsch, line 63: I don't understand why and do not agree. WDYT?

domain/ide/workspace/workspace.go
```diff
@@ -174,10 +174,9 @@ func (w *Workspace) ChangeWorkspaceFolders(ctx context.Context, params lsp.DidCh
 	for _, folder := range params.Event.Added {
 		f := NewFolder(uri.PathFromUri(folder.Uri), folder.Name, w.scanner, w.hoverService, w.scanNotifier, w.notifier)
 		w.AddFolder(f)
-	}
-
-	if config.CurrentConfig().IsAutoScanEnabled() {
-		w.ScanWorkspace(ctx)
+		if config.CurrentConfig().IsAutoScanEnabled() {
+			f.ScanFolder(ctx)
+		}
```
bastiandoetsch, line 179: Adding a workspace folder should not trigger a workspace scan. It should just trigger a scan of the added folder.

infrastructure/code/cache_test.go
bastiandoetsch, line 1: Same here, cache testing moved to its own file

infrastructure/code/cache_test.go
```diff
@@ -0,0 +1,195 @@
...
+
+		for i := 0; i < len(results); i++ {
+			select {
+			case path := <-evictionChan:
+				log.Debug().Msg("evicted from cache" + path)
+			case <-time.After(time.Second):
+				t.Fatal("timeout waiting for eviction")
+			}
+		}
+	})
```
bastiandoetsch, line 120: This is a new test

application/codeaction/codeaction_test.go
```diff
@@ -41,7 +41,7 @@ type mockIssuesProvider struct {
 	mock.Mock
 }
 
-func (m *mockIssuesProvider) Issues() map[string][]snyk.Issue {
+func (m *mockIssuesProvider) Issues() snyk.IssuesByFile {
```
bastiandoetsch, line 44: Introduced a new type to make handling of all these maps a bit easier

domain/ide/command/logout_test.go
```diff
@@ -66,6 +66,8 @@ func TestLogoutCommand_Execute_ClearsIssues(t *testing.T) {
 	w.AddFolder(folder)
 
 	ctx := context.Background()
+	path := filepath.Join(folder.Path(), "path1")
```
bastiandoetsch, line 69: A lot of tests had incorrect filepathes that weren't inside the folder.

domain/ide/workspace/folder.go
```diff
@@ -79,21 +79,58 @@ type Folder struct {
...
 	// so we get the global issues first, then append the scanner-local issues
-	issues := make(map[string][]snyk.Issue)
-	f.documentDiagnosticCache.Range(func(key string, value []snyk.Issue) bool {
-		issues[key] = value
+	issues := snyk.IssuesByFile{}
+	f.documentDiagnosticCache.Range(func(path string, value []snyk.Issue) bool {
+		if f.Contains(path) {
+			issues[path] = value
+		} else {
+			panic("issue found in cache that does not pertain to folder")
```
bastiandoetsch, line 121: Same here, we should not have issues in our global folder cache that do not belong to the folder.

domain/ide/workspace/folder.go
```diff
@@ -79,21 +79,58 @@ type Folder struct {
...
 		return true
 	})
 
 	// scanner-local issues: if the scanner is an IssueProvider, we append the issues it knows about
 	issueProvider, scannerIsIssueProvider := f.scanner.(snyk.IssueProvider)
 	if scannerIsIssueProvider {
 		cachedScannerIssues := issueProvider.Issues()
-		for key, value := range cachedScannerIssues {
-			issues[key] = append(issues[key], value...)
+		for path, issuesForPath := range cachedScannerIssues {
```
bastiandoetsch, line 130: A scanner is multi-folder capable. So we only take the issues that belong to the folder.

domain/ide/workspace/folder_test.go
teodora-sandu: You do not agree that it should **not** rescan if the results are cached? To me the test makes sense, isn't that the whole point of the caching?

infrastructure/code/cache.go
```diff
@@ -0,0 +1,109 @@
...
+package code
+
+import (
+	"github.com/erni27/imcache"
+
+	"github.com/snyk/snyk-ls/domain/snyk"
+	"github.com/snyk/snyk-ls/internal/product"
+)
+
+func (sc *Scanner) addToCache(results []snyk.Issue) {
```
teodora-sandu, line 26: This question doesn't relate to this PR necessarily but I'm wondering why we need to have these functions in the code package? There doesn't seem to be anything code specific in here and I'm wondering if these functions can be re-used by other scanners to. So in the delegating concurrent scanner instead of delegating down for this for no reason we do all this in there

application/server/server.go
```diff
@@ -451,7 +451,7 @@ func textDocumentDidOpenHandler() jrpc2.Handler {
 			return nil, nil
 		}
 
-		filteredIssues := workspace.FilterIssues(folder.Issues(), config.CurrentConfig().DisplayableIssueTypes())
+		filteredIssues := folder.FilterIssues(folder.Issues(), config.CurrentConfig().DisplayableIssueTypes())
```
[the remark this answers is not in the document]
bastiandoetsch: We didn't need it on workspace level, I think.

domain/ide/workspace/folder_test.go
bastiandoetsch: Caching makes sense, when we  _access_ information, but not when we try to create. `ScanFile` is supposed to trigger the analysis and the creation of diagnostics. It is e.g. called by `DidSaveTextDocument`. `DidOpenTextDoument` instead only reads the cache, and does not trigger a scan. Stopping a scan because we have cached data is counter-intuitive for me. I'd expect a scan to update the cache, not use it.

infrastructure/code/cache.go
teodora-sandu: Nevermind, ignore me. `sc.issueCache` is Code's cache

bastiandoetsch: Yeah, most likely it could be extracted to be used by all scanners. That may be a good refactor when we update the other scanners to have their own caches instead of using a global cache.

domain/ide/workspace/folder.go
```diff
@@ -468,18 +532,26 @@ func (f *Folder) IssuesForRange(filePath string, requestedRange snyk.Range) (mat
...
+		for path := range issues {
+			if f.Contains(path) {
+				cacheProvider.ClearIssues(path)
+			}
+		}
+	}
 }
 
 func (f *Folder) ClearDiagnosticsByIssueType(removedType product.FilterableIssueType) {
+	// TODO implement scanner variant
```
teodora-sandu, line 554: What about this todo?

domain/ide/workspace/folder.go
```diff
@@ -468,18 +532,26 @@ func (f *Folder) IssuesForRange(filePath string, requestedRange snyk.Range) (mat
...
+		issues := cacheProvider.Issues()
+		for path := range issues {
+			if f.Contains(path) {
+				cacheProvider.ClearIssues(path)
+			}
+		}
+	}
 }
 
 func (f *Folder) ClearDiagnosticsByIssueType(removedType product.FilterableIssueType) {
```
teodora-sandu, line 553: Could we maybe have `ClearDiagnostics` take in an optional parameter so that we don't have both the `ClearDiagnostics` and `ClearDiagnosticsByIssueType`? We have a lot of caching functions at the moment that are making it a bit hard to understand what's going on

teodora-sandu: This might be a personal preference so feel free to disagree

teodora-sandu (approved):
Works as far as I can tell (at least saving a file triggers a new scan) and there are smoke tests added, which increase my confidence. The code has gotten quite complicated but I'm not sure what else we could do. Will approve, I might have a look after at the code to see if there's any tidying we could do ⭐

infrastructure/code/code.go
```diff
@@ -83,7 +83,8 @@ type Scanner struct {
 	codeScanner  codeClient.CodeScanner
 	// this is the local scanner issue cache. In the future, it should be used as source of truth for the issues
 	// the cache in workspace/folder should just delegate to this cache
-	issueCache *imcache.Cache[string, []snyk.Issue]
+	issueCache          *imcache.Cache[string, []snyk.Issue]
```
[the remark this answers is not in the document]
bastiandoetsch: Not sure I understand your proposal :)

domain/ide/workspace/folder.go
bastiandoetsch: This looks pretty redundant, I agree.

PR #493, opened 2024-04-18
refactor: use function options pattern

teodora-sandu:
### Description

Updates the `code-client-go` library and picks up the logging improvements and function options pattern changes.


go.mod
```diff
@@ -12,6 +12,7 @@ require (
 	github.com/golang/mock v1.6.0
 	github.com/gomarkdown/markdown v0.0.0-20230922112808-5421fefb8386
 	github.com/google/uuid v1.6.0
+	github.com/hashicorp/go-uuid v1.0.3
```
bastiandoetsch, line 15: Why not use the google lib that is already imported?

bastiandoetsch (approved):

PR #502, opened 2024-05-09
refactor: oss issue data [IDE-283]

teodora-sandu:
### Description

Refactors the types used for OSS so that:
- they're used consistently and so the OSS scan deals with `oss.ossIssue`, the rendered HTML deals with `snyk.OssIssueData`, and the `snyk.scan` notification deals with `lsp.OssIssueData`, which is converted from `snyk.OssIssueData` (before this change the rendered HTML dealt with `oss.ossIssue` directly, and so the data given to `snyk.scan` and to the HTML logic differed)
- as a result, we are able to easily include new fields in the `snyk.scan` notification that can be used to build the UI (e.g. `MatchingIssues`, `Lesson`, `Identifiers`)

This will be the building block for rendering OSS vulnerabilities in IntelliJ via the LS, but not from HTML yet.

### Checklist

- [x] Linted

### Screenshots / GIFs

Tested in the VSCode extension by running `make build-debug` in this repo and then changing the path to LS in VSCode to it.

Before:
Screenshot 2024-05-09 at 11 23 22

After:
Screenshot 2024-05-09 at 11 23 57
Screenshot 2024-05-09 at 11 24 34

Proof of using the new LS FROM LOGS:
```
[Info  - 11:24:16] 2024-05-09T11:24:16+01:00 INF - snyk-ls: v20240509.112159-SNAPSHOT-390c0ac (/Users/teodorasandu/Documents/repos/snyk-ls/build/snyk-ls)
```

### DIFF
--- a/application/server/notification/scan_notifier.go
+++ b/application/server/notification/scan_notifier.go
@@ -117,6 +117,34 @@ func (n *scanNotifier) appendOssIssues(scanIssues []lsp.ScanIssue, folderPath st
 			continue // skip non-oss issues
 		}
 
+		matchingIssues := make([]lsp.OssIssueData, len(additionalData.MatchingIssues))
+		for i, matchingIssue := range additionalData.MatchingIssues {
+			matchingIssues[i] = lsp.OssIssueData{
+				License: matchingIssue.License,
+				Identifiers: lsp.OssIdentifiers{
+					CWE: issue.CWEs,
+					CVE: issue.CVEs,
+				},
+				Description:       matchingIssue.Description,
+				Language:          matchingIssue.Language,
+				PackageManager:    matchingIssue.PackageManager,
+				PackageName:       matchingIssue.PackageName,
+				Name:              matchingIssue.Name,
+				Version:           matchingIssue.Version,
+				Exploit:           matchingIssue.Exploit,
+				CVSSv3:            matchingIssue.CVSSv3,
+				CvssScore:         strconv.FormatFloat(matchingIssue.CvssScore, 'f', 2, 64), // convert float64 to string with 2 decimal places
+				FixedIn:           matchingIssue.FixedIn,
+				From:              matchingIssue.From,
+				UpgradePath:       matchingIssue.UpgradePath,
+				IsPatchable:       matchingIssue.IsPatchable,
+				IsUpgradable:      matchingIssue.IsUpgradable,
+				ProjectName:       matchingIssue.ProjectName,
+				DisplayTargetFile: matchingIssue.DisplayTargetFile,
+				Details:           matchingIssue.Details,
+			}
+		}
+
 		scanIssues = append(scanIssues, lsp.ScanIssue{
 			Id:       additionalData.Key,
 			Title:    additionalData.Title,
@@ -146,6 +174,8 @@ func (n *scanNotifier) appendOssIssues(scanIssues []lsp.ScanIssue, folderPath st
 				ProjectName:       additionalData.ProjectName,
 				DisplayTargetFile: additionalData.DisplayTargetFile,
 				Details:           additionalData.Details,
+				MatchingIssues:    matchingIssues,
+				Lesson:            additionalData.Lesson,
 			},
 		})
 	}
--- a/infrastructure/oss/issue.go
+++ b/infrastructure/oss/issue.go
@@ -19,19 +19,14 @@ package oss
 import (
 	_ "embed"
 	"fmt"
-	"net/url"
-	"strings"
 
 	"github.com/gomarkdown/markdown"
-	"github.com/pkg/errors"
-	"github.com/rs/zerolog/log"
 
 	"github.com/snyk/snyk-ls/application/config"
 	"github.com/snyk/snyk-ls/domain/observability/error_reporting"
 	"github.com/snyk/snyk-ls/domain/snyk"
 	"github.com/snyk/snyk-ls/infrastructure/learn"
 	"github.com/snyk/snyk-ls/internal/product"
-	"github.com/snyk/snyk-ls/internal/util"
 )
 
 var issuesSeverity = map[string]snyk.Severity{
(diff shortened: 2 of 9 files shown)


infrastructure/oss/issue.go
```diff
@@ -41,127 +36,6 @@ var issuesSeverity = map[string]snyk.Severity{
 	"medium":   snyk.Medium,
 }
 
-func (i *ossIssue) AddCodeActions(learnService learn.Service, ep error_reporting.ErrorReporter) (actions []snyk.
```
teodora-sandu, line 44: I just moved these functions to the actual definition of the type

infrastructure/oss/issue_html.go
```diff
@@ -32,7 +35,7 @@ func replaceVariableInHtml(html string, variableName string, variableValue strin
 	return strings.ReplaceAll(html, fmt.Sprintf("${%s}", variableName), variableValue)
 }
 
-func getIdentifiers(issue *ossIssue) string {
+func getIdentifiers(id string, issue snyk.OssIssueData) string {
```
teodora-sandu, line 38: Changed these functions to work on the `snyk` domain

application/server/notification/scan_notifier.go
```diff
@@ -117,6 +117,34 @@ func (n *scanNotifier) appendOssIssues(scanIssues []lsp.ScanIssue, folderPath st
 			continue // skip non-oss issues
 		}
 
+		matchingIssues := make([]lsp.OssIssueData, len(additionalData.MatchingIssues))
+		for i, matchingIssue := range additionalData.MatchingIssues {
```
teodora-sandu, line 121: This now includes the matching issues in the type, so that both the HTML renderer using in. VSCode and the native UI in the IntelliJ plugin can use it to build the following:
Screenshot 2024-05-09 at 13 38 39

infrastructure/oss/issue.go
```diff
@@ -170,21 +44,22 @@ func toIssue(
...
-	title := issue.Title
-
-	if config.CurrentConfig().Format() == config.FormatHtml {
-		title = string(markdown.ToHTML([]byte(title), nil, nil))
-	}
-	remediationAdvice := getRemediationAdvice(issue)
-
 	// find all issues with the same id
-	matchingIssues := []ossIssue{}
+	matchingIssues := []snyk.OssIssueData{}
```
teodora-sandu, line 48: This is just re-ordering of the code so that we can use the `additionalData` to compute the remediation advice

internal/lsp/message_types.go
```diff
@@ -1066,6 +1066,8 @@ type OssIssueData struct {
 	ProjectName       string         `json:"projectName"`
 	DisplayTargetFile string         `json:"displayTargetFile"`
 	Details           string         `json:"details,omitempty"`
+	MatchingIssues    []OssIssueData `json:"matchingIssues"`
```
teodora-sandu, line 1069: Now the LS includes these fields in the `snyk.scan` notification

infrastructure/oss/types.go
```diff
@@ -56,10 +67,183 @@ type ossIssue struct {
...
+
+func (i *ossIssue) GetExtendedMessage(issue ossIssue) string {
+	title := issue.Title
+	description := issue.Description
+
+	if config.CurrentConfig().Format() == config.FormatHtml {
+		title = string(markdown.ToHTML([]byte(title), nil, nil))
+		description = string(markdown.ToHTML([]byte(description), nil, nil))
+	}
+	summary := fmt.Sprintf("### Vulnerability %s %s %s \n **Fixed in: %s | Exploit maturity: %s**",
```
cmars, line 135: thought: Looks like we're rendering in code fragments, where we could be exposing the data model and rendering entirely within the template. Probably out of scope for this PR, but it looks like the OSS template needs to be migrated to Go templates to make that happen.

Could we get that in a followup? 🙏

teodora-sandu: Just clarifying first that this is mostly moving around code, so the HTML here is not something new I added. You're right though that we should improve this! @cat2608 I remember you already mentioned that we might want a ticket in the future  for when we start rending the HTML in IntelliJ too for OSS. Do you remember if we created one? If not I can add a ticket, though it might not be done soon since it's out of scope for this cycle. But we should definitely add it to the next cooldown

cat2608: @teodora-sandu we don't have a ticket to migrate OSS to `HTML/template`. 
Please, create one if you can 🙏 Many thanks 🙌

application/server/notification/scan_notifier.go
```diff
@@ -117,6 +117,34 @@ func (n *scanNotifier) appendOssIssues(scanIssues []lsp.ScanIssue, folderPath st
 			continue // skip non-oss issues
 		}
 
+		matchingIssues := make([]lsp.OssIssueData, len(additionalData.MatchingIssues))
```
cat2608, line 120: **Non-blocking**: I noticed that `folderPath` isn't used in this function, as well as in others called from the same place.:

```go
	if pr == product.ProductInfrastructureAsCode {
		scanIssues = n.appendIacIssues(scanIssues, folderPath, issues)
	} else if pr == product.ProductCode {
		scanIssues = n.appendCodeIssues(scanIssues, folderPath, issues)
	} else if pr == product.ProductOpenSource {
		scanIssues = n.appendOssIssues(scanIssues, folderPath, issues)
	}
```

Maybe we could clean them up too? We could simplify the function signatures by removing it?

internal/lsp/message_types.go
cat2608: `MatchingIssues` field is a slice of `OssIssueData`. Is it correct to say that this is a recursive data structure? Is it possible that with large data this could get complex? Is there a way to test it for deeply nested data if it make sense?

cat2608 (approved):
Tested it locally 🚀

cat2608:
Ups, I think we are missing the **Learn** kink with this changes, could it be? I just noticed that on my local test and we can see it too at your **before** and **after** screenshots

application/server/notification/scan_notifier.go
teodora-sandu: I noticed that too but I'm already changing a lot of code and I had to set a limit to the number of refactorings I can do 😓

teodora-sandu:
Good spot @cat2608 ! I found the problem and added a test to make sure we don't break this in the future. Here is the latest
Screenshot 2024-05-10 at 11 42 42

PR #503, opened 2024-05-10
chore: update code-client to use new interface

PeterSchafer:
### Description

This PR updates to the latest version of code-client-go, to make use of its new interface.

### Checklist

- [ ] Linted

### Screenshots / GIFs


infrastructure/code/code.go
```diff
@@ -360,7 +361,12 @@ func (sc *Scanner) UploadAndAnalyzeWithIgnores(ctx context.Context,
 	requestId := span.GetTraceId() // use span trace id as code-request-id
 	log.Info().Str("requestId", requestId).Msg("Starting Code analysis.")
 
-	sarif, bundleHash, err := sc.codeScanner.UploadAndAnalyze(ctx, requestId, path, files, changedFiles)
+	target, err := scan.NewRepositoryTarget(path)
+	if err != nil {
+		log.Warn().Err(err)
+	}
```
cat2608, line 367: Is it possible that we need a return here so we don't propagate a "corrupted" `target`? Maybe we could validate this idea with a sad path test?

PeterSchafer: I decided to just log an error, because I wanted to avoid to code to much assumption on the behaviour in snyk-ls and leave it to code-client-go to react on targets that it can't deal with. This means I would actually like to ignore the error here but the linter wants me to handle it. So logging seems like a reasonable thing.

cat2608 (approved):

PR #504, opened 2024-05-10
feat: include ruleId in LS response for OSS [IDE-238] 

teodora-sandu:
### Description

Adds a `ruleId` field to the LS message so that in IntelliJ we can include it in the rendered UI.

### Checklist

- [x] Linted


acke (approved):
Looks good to me.

PR #508, opened 2024-05-15
fix: fix pact-ruby-standalone to v2.4.2

cat2608:
### Description

This PR fixes `pact-ruby-standalone` to v2.4.2 in GHAs

### Checklist

- [ ] Linted


.github/workflows/build.yaml
```diff
@@ -46,7 +46,7 @@ jobs:
         if: steps.cache-pact.outputs.cache-hit != 'true'
         run: |
           cd ~
-          curl -fsSL https://raw.githubusercontent.com/pact-foundation/pact-ruby-standalone/master/install.sh | bash
+          curl -fsSL https://raw.githubusercontent.com/pact-foundation/pact-ruby-standalone/master/install.sh | PACT_CLI_VERSION=v2.4.2 sh
```
cat2608, line 49: I'll try to move the `PACT_CLI_VERSION` to an env var in the next iteration.

https://docs.github.com/en/actions/learn-github-actions/variables

cmars, line 49: Could we replace this line and the one below with `make tools` to consolidate the pact version in one place (the makefile)?

Just a non-blocking thought.

cmars (approved):

PR #509, opened 2024-05-15
chore: explicitly set token authentication when using the CLI

PeterSchafer:
### Description

The CLI default authentication will be switching to OAuth, since snyk-ls uses the CLI for token based authentication, it is necessary to explicitly state the `auth-type` to be token.

This can be tested with the `preview` version of the CLI, which already uses OAuth as default.
The expected behaviour is that snyk-ls still uses token based authentication unless it is connecting to fedramp instances or the authentication type is changed.

### Checklist

- [x] Linted

### Screenshots / GIFs


cmars (approved):
Tested locally with VSCode.

Wondering why oauth didn't work... maybe because we're testing the token v1-style instead of using `/rest/self` in LS?

PeterSchafer:
Closing since this change will be replaced by another solution in the CLI

PR #510, opened 2024-05-15
fix: ignored vuln links should work with custom endpoint [IDE-249]

ShawkyZ:
### Description

- Calculate SnykUi endpoint from config.currentConfig.SnykAPI
- Replace hardcoded SnykUI URL for OSS and Code html.

### Checklist

- [ ] Linted

### Screenshots / GIFs

### DIFF
--- a/application/config/config.go
+++ b/application/config/config.go
@@ -61,6 +61,7 @@ const (
 	FormatMd              = "md"
 	snykCodeTimeoutKey    = "SNYK_CODE_TIMEOUT" // timeout as duration (number + unit), e.g. 10m
 	DefaultSnykApiUrl     = "https://snyk.io/api"
+	DefaultSnykUiUrl      = "https://app.snyk.io/"
 	DefaultDeeproxyApiUrl = "https://deeproxy.snyk.io"
 	pathListSeparator     = string(os.PathListSeparator)
 	windows               = "windows"
@@ -351,7 +352,7 @@ func (c *Config) SnykCodeApi() string { return c.snykCodeApiUrl }
 func (c *Config) SnykUi() string {
 	parsedUrl, err := url.Parse(c.snykApiUrl)
 	if err != nil {
-		return "https://app.snyk.io"
+		return DefaultSnykUiUrl
 	}
 	host := parsedUrl.Host
 	appHostname := strings.Replace(host, "api", "app", 1)
--- a/application/config/config.go
+++ b/application/config/config.go
@@ -350,13 +350,12 @@ func (c *Config) LogPath() string {
 func (c *Config) SnykApi() string     { return c.snykApiUrl }
 func (c *Config) SnykCodeApi() string { return c.snykCodeApiUrl }
 func (c *Config) SnykUi() string {
-	parsedUrl, err := url.Parse(c.snykApiUrl)
-	if err != nil || !parsedUrl.IsAbs() {
+	snykUiUrl, err := getCustomEndpointUrlFromSnykApi(c.snykApiUrl, "app")
+	if err != nil && snykUiUrl == "" {
 		return DefaultSnykUiUrl
 	}
-	host := parsedUrl.Host
-	appHostname := strings.Replace(host, "api", "app", 1)
-	return fmt.Sprintf("%s://%s", parsedUrl.Scheme, appHostname)
+
+	return snykUiUrl
 }
 func (c *Config) SnykCodeAnalysisTimeout() time.Duration { return c.snykCodeAnalysisTimeout }
 func (c *Config) IntegrationName() string {
@@ -585,16 +584,20 @@ func getCodeApiUrlFromCustomEndpoint(endpoint string) (string, error) {
 	}
 
 	// Use Snyk API endpoint to determine deeproxy API URL
-	endpointUrl, err := url.Parse(strings.Trim(endpoint, " "))
-	if err != nil {
+	return getCustomEndpointUrlFromSnykApi(endpoint, "deeproxy")
+}
+
+func getCustomEndpointUrlFromSnykApi(snykApi string, subdomain string) (string, error) {
+	snykApiUrl, err := url.Parse(strings.Trim(snykApi, " "))
+	if err != nil && !snykApiUrl.IsAbs() {
 		return "", err
 	}
 
 	m := regexp.MustCompile(`^(ap[pi]\.)?`)
-	endpointUrl.Host = m.ReplaceAllString(endpointUrl.Host, "deeproxy.")
-	endpointUrl.Path = ""
+	snykApiUrl.Host = m.ReplaceAllString(snykApiUrl.Host, subdomain+".")
+	snykApiUrl.Path = ""
 
-	return endpointUrl.String(), nil
+	return snykApiUrl.String(), nil
 }
 
 func snykCodeAnalysisTimeoutFromEnv() time.Duration {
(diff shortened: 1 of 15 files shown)


application/config/config.go
```diff
@@ -346,8 +346,17 @@ func (c *Config) LogPath() string {
 	defer c.m.Unlock()
 	return c.logPath
 }
-func (c *Config) SnykApi() string                        { return c.snykApiUrl }
-func (c *Config) SnykCodeApi() string                    { return c.snykCodeApiUrl }
+func (c *Config) SnykApi() string     { return c.snykApiUrl }
+func (c *Config) SnykCodeApi() string { return c.snykCodeApiUrl }
+func (c *Config) SnykUi() string {
+	parsedUrl, err := url.Parse(c.snykApiUrl)
```
cmars, line 352: Be careful here... an empty string would not error here, but it'd have an invalid host.

[parsedUrl.IsAbs()](https://pkg.go.dev/net/url#URL.IsAbs) could help guard against that.

See https://go.dev/play/p/fO4rfoYnZsX

application/config/config.go
```diff
@@ -346,8 +346,17 @@ func (c *Config) LogPath() string {
...
+	if err != nil {
+		return "https://app.snyk.io"
```
cmars, line 354: Wondering if this should be a const.

cmars (commented):
Almost LGTM; I'd be careful about the empty string case (which is easy to have happen, as it's the default value for string in Go)

application/config/config.go
ShawkyZ: Agreer

application/config/config.go
ShawkyZ: Thanks Casey, I added the check for IsAbs.

application/config/config.go
```diff
@@ -346,8 +347,17 @@ func (c *Config) LogPath() string {
...
+func (c *Config) SnykCodeApi() string { return c.snykCodeApiUrl }
+func (c *Config) SnykUi() string {
+	parsedUrl, err := url.Parse(c.snykApiUrl)
+	if err != nil || !parsedUrl.IsAbs() {
+		return DefaultSnykUiUrl
+	}
+	host := parsedUrl.Host
+	appHostname := strings.Replace(host, "api", "app", 1)
+	return fmt.Sprintf("%s://%s", parsedUrl.Scheme, appHostname)
+}
```
cat2608, line 360: Shall we add a unit test for this logic? Maybe we could check if `c.snykApiUrl` is not a valid or absolute URL and the happy paths? `TestSnykCodeApi` seems to be covering a few interesting cases that we could have a look?

ShawkyZ: I added a couple of tests similar to the `TestSnykCodeApi` tests

application/config/config_test.go
```diff
@@ -300,3 +300,39 @@ func Test_IsTelemetryEnabled(t *testing.T) {
 	assert.False(t, c.IsTelemetryEnabled())
 	assert.True(t, c.Engine().GetConfiguration().GetBool(configuration.ANALYTICS_DISABLED))
 }
+
+func TestSnykUiEndpoint(t *testing.T) {
+	c := New()
+	t.Run("Default Api Endpoint with /api prefix", func(t *testing.T) {
+		uiEndpoint := c.SnykUi()
+		assert.Equal(t, "https://snyk.io", uiEndpoint)
```
ShawkyZ, line 308: @cat2608 I'm not really sure here if this is right though. 
If the the API endpoint doesn't have the `api` or `app` prefix. Is it correct to to use `snyk.io` as the SnykUi endpoint?

Maybe we should always append `app` as a prefix subdomain ?

application/config/config.go
```diff
@@ -61,6 +61,7 @@ const (
 	FormatMd              = "md"
 	snykCodeTimeoutKey    = "SNYK_CODE_TIMEOUT" // timeout as duration (number + unit), e.g. 10m
 	DefaultSnykApiUrl     = "https://snyk.io/api"
```
bastiandoetsch, line 63: Not necessary for this PR, but should probably be changed to `https://api.snyk.io`

application/config/config_test.go
```diff
@@ -300,3 +300,53 @@ func Test_IsTelemetryEnabled(t *testing.T) {
 	assert.False(t, c.IsTelemetryEnabled())
 	assert.True(t, c.Engine().GetConfiguration().GetBool(configuration.ANALYTICS_DISABLED))
 }
+
+func TestSnykUiEndpoint(t *testing.T) {
```
bastiandoetsch, line 304: suggestion(nitpick): add snykgov test case, e.g. `api.fedramp.snykgov.io` -> `app.fedramp.snykgov.io`

infrastructure/code/code_html.go
```diff
@@ -104,6 +105,7 @@ func getCodeDetailsHtml(issue snyk.Issue) string {
 		"RepoCount":          additionalData.RepoDatasetSize,
 		"ExampleCount":       len(additionalData.ExampleCommitFixes),
 		"ExampleCommitFixes": prepareExampleCommitFixes(additionalData.ExampleCommitFixes),
+		"SnykUi":             config.CurrentConfig().SnykUi(),
```
bastiandoetsch, line 108: suggestion(nitpick): probably would be better for future maintainability to inject a config to avoid potential multi-threading issues.

infrastructure/code/code_html_test.go
```diff
@@ -108,6 +109,44 @@ func Test_Code_Html_getCodeDetailsHtml_ignored(t *testing.T) {
 	assert.Contains(t, codePanelHtml, `class="ignore-details-section"`)
 }
 
+func Test_Code_Html_getCodeDetailsHtml_ignored_customEndpoint(t *testing.T) {
+	_ = testutil.UnitTest(t)
```
bastiandoetsch, line 113: question: why not assign the returned config to a var and use it later on instead of using `config.CurrentConfig()`?

infrastructure/oss/issue_html.go
```diff
@@ -95,6 +95,11 @@ func getIntroducedBy(issue snyk.OssIssueData) string {
 	}
 }
 
+func getVulnHtmlAnchor(packageManager string, module string) string {
+	snykUi := config.CurrentConfig().SnykUi()
+	return fmt.Sprintf("<a href='%s/test/%s'>%s</a>", snykUi, packageManager, module)
```
bastiandoetsch, line 100: This is a completely unnecessary comment. Just wanted to add that I recently read, that `fmt.Sprintf` uses more heap allocations than `strings.Builder`. So it may be more efficient to use the latter - but in this case performance should not be an issue.

infrastructure/oss/issue_html_test.go
```diff
@@ -101,3 +103,35 @@ func Test_OssDetailsPanel_html_withLearn(t *testing.T) {
 
 	assert.True(t, strings.Contains(issueDetailsPanelHtml, "Learn about this vulnerability"))
 }
+
+func Test_OssDetailsPanel_html_withLearn_withCustomEndpoint(t *testing.T) {
+	_ = testutil.UnitTest(t)
```
bastiandoetsch, line 108: see above, same question.

infrastructure/oss/issue_html.go
bastiandoetsch: Interesting read regarding this: https://stackoverflow.com/questions/1760757/how-to-efficiently-concatenate-strings-in-go/47798475#47798475

bastiandoetsch (approved):

PR #513, opened 2024-05-20
feat: add Learn link to header template [IDE-301]

cat2608:
### Description

This PR adds the Lesson Link to the header template

Note: styling in next iterations

### Checklist

- [x] Linted

### Screenshots / GIFs

image


domain/snyk/issues.go
```diff
@@ -101,6 +101,7 @@ type CodeIssueData struct {
 	HasAIFix           bool               `json:"hasAIFix"`
 	DataFlow           []DataFlowElement  `json:"dataFlow,omitempty"`
 	Details            string             `json:"details"`
+	LessonUrl          string             `json:"url"`
```
bastiandoetsch, line 104: suggestion: this could even be added one level up, as we have learn lessons for all products, not only code. WDYT?

infrastructure/code/issue_enhancer.go
```diff
@@ -80,6 +106,9 @@ func (b *IssueEnhancer) addIssueActions(ctx context.Context, issues []snyk.Issue
 		return
 	}
 
+	// Populate HTML template
+	b.enhanceIssuesDetails(issues)
+
```
bastiandoetsch, line 111: suggestion: let's rename `addIssueActions` func to something like `addActionsAndDetails` or `enrichIssue`, as it's now having an unexpected side effect.

bastiandoetsch (approved):

infrastructure/code/issue_enhancer.go
cat2608: @bastiandoetsch I think I found a better place for this enhancement; I made `Scanner` responsible for enhancing the issues before adding the issues to the cache. WDYT?

domain/snyk/issues.go
cat2608: It makes sense since `OssIssueData` implements it in its own way, so we could unify them by moving `LessonUrl` to the Issue struct as you suggest.

infrastructure/code/issue_enhancer.go
bastiandoetsch: Might have a slight performance penalty (iterating over issues again), but should be fine :)

PR #514, opened 2024-05-20
feat: use progress tracker from code-client-go [IDE-222]

teodora-sandu:
### Description

Uses the new progress tracker factory and tracker from `code-client-go` in https://github.com/snyk/code-client-go/pull/49. 

To test (currently blocked, see explanation underneath the screen recording):
- Run `make build-debug` in `snyk-ls`
- Run `make build` in `cli` in the branch in https://github.com/snyk/cli/pull/5263
- Change the path to the CLI in IntelliJ in the Snyk Settings to the local CLI binary built in the previous step and disable automatically managing binaries
- Make sure to use an org that has the `snykCodeConsistentIgnores` FF enabled (e.g. `ide-consistent-ignores-test`)

### Checklist

- [x] Linted

### Screenshots / GIFs

Tested with `target-service` because it's smaller and quicker to scan.

https://github.com/snyk/snyk-ls/assets/81559517/03d1405d-b2ea-4e18-a430-2ad078f9ad39

There's a bug for this feature that is caused by the `/v1` path being added to some of the APIs so the flow won't be successful now (it was at the time of taking that screen recording): https://snyksec.atlassian.net/browse/IDE-353.


infrastructure/code/code_tracker.go
```diff
@@ -0,0 +1,93 @@
...
+	})
+
+	t.send(lsp.ProgressParams{
+		Token: t.token,
+		Value: lsp.WorkDoneProgressBegin{
+			WorkDoneProgressKind: lsp.WorkDoneProgressKind{Kind: "begin"},
+			Title:                title,
+			Message:              message,
+			Cancellable:          false,
+			Percentage:           0,
```
bastiandoetsch, line 69: I'm not sure if this should be 0 or 1. I THINK that 0 is ok for non-quantifiable progress, though.

infrastructure/code/code_tracker.go
```diff
@@ -0,0 +1,93 @@
...
+		},
+	})
+}
+
+func (t *tracker) End(message string) {
+	if t.finished {
+		panic("Called end progress twice. This breaks LSP in Eclipse fix me now and avoid headaches later")
```
bastiandoetsch, line 76: Ok, now that we're at it: Let's add some interpunctuation: `This breaks LSP in Eclipse. Fix me now and avoid headaches later`

infrastructure/code/code_tracker.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	t.finished = true
+	t.send(lsp.ProgressParams{
+		Token: t.token,
+		Value: lsp.WorkDoneProgressEnd{
+			WorkDoneProgressKind: lsp.WorkDoneProgressKind{Kind: "end"},
```
bastiandoetsch, line 82: nitpick(non-blocking): `end` should probably be a constant

infrastructure/code/code_tracker.go
```diff
@@ -0,0 +1,93 @@
...
+
+	t.send(lsp.ProgressParams{
+		Token: t.token,
+		Value: nil,
+	})
+
+	t.send(lsp.ProgressParams{
+		Token: t.token,
+		Value: lsp.WorkDoneProgressBegin{
+			WorkDoneProgressKind: lsp.WorkDoneProgressKind{Kind: "begin"},
```
bastiandoetsch, line 65: nitpick(non-blocking): `begin` should probably be a constant.

bastiandoetsch (approved):

infrastructure/code/code_tracker.go
teodora-sandu: Yep 0 is for unquantifiable, you're right: https://github.com/snyk/snyk-ls/blob/22e4e7e42de1361adac84c3f5ee72f2555204a44/internal/progress/progress.go#L146-L148

PR #526, opened 2024-05-30
feat: display Code Issues information in tabs [IDE-271]

cat2608:
### Description

This PR updates the Code Issue panel template to render the data in three columns.

### Checklist

- [ ] Linted

### Screenshots / GIFs

>[!NOTE]
>Styling will be addressed by #522

https://github.com/snyk/snyk-ls/assets/1948377/beba2503-6664-42e1-bb77-a0345e95eb6c

### DIFF
--- a/infrastructure/code/template/details.html
+++ b/infrastructure/code/template/details.html
@@ -91,6 +91,7 @@
       flex-direction: column;
     }
 
+    .issue-overview-header,
     .ignore-details-header,
     .data-flow-header,
     .example-fixes-header {
@@ -580,7 +581,6 @@
           <!-- Content for Vulnerability Overview -->
           <section class="issue-overview">
             <div class="overview-text">
-              <h2>Details</h2>
               {{.IssueOverview}}
             </div>
           </section>
@@ -615,8 +615,14 @@
     const EXAMPLE_TAB_CONTENT_SELECTOR = '.example-fixes .tab-content';
     const IGNORE_WARNING_SELECTOR = '.ignore-warning-wrapper';
     const FIRST_EXAMPLE_CONTENT_ID = '#tab-content-0';
+    const ISSUE_OVERVIEW_MD_HEADER = '.issue-overview .overview-text h2';
 
     document.addEventListener('DOMContentLoaded', () => {
+      const overviewHeader = document.querySelector(ISSUE_OVERVIEW_MD_HEADER);
+      if (overviewHeader) {
+        overviewHeader.classList.add('issue-overview-header');
+      }
+
       const mainTabLinks = document.querySelectorAll(MAIN_TAB_ITEM_SELECTOR);
       const mainTabContents = document.querySelectorAll(MAIN_TAB_CONTENT_SELECTOR);
 
--- a/infrastructure/code/template/details.html
+++ b/infrastructure/code/template/details.html
@@ -605,8 +605,6 @@
 
   <!-- Scripts -->
   <script nonce="{{.Nonce}}">
-    const ideType = 'snyk-vscode';
-
     const MAIN_TAB_NAV_SELECTOR = '.main-tabs-nav';
     const MAIN_TAB_ITEM_SELECTOR = '.main-tabs-nav .tab-item';
     const MAIN_TAB_CONTENT_SELECTOR = '.tab-container > .tab-content';
@@ -615,14 +613,9 @@
     const EXAMPLE_TAB_CONTENT_SELECTOR = '.example-fixes .tab-content';
     const IGNORE_WARNING_SELECTOR = '.ignore-warning-wrapper';
     const FIRST_EXAMPLE_CONTENT_ID = '#tab-content-0';
-    const ISSUE_OVERVIEW_MD_HEADER = '.issue-overview .overview-text h2';
 
+    // Ensure the document is fully loaded before executing script to manipulate DOM.
     document.addEventListener('DOMContentLoaded', () => {
-      const overviewHeader = document.querySelector(ISSUE_OVERVIEW_MD_HEADER);
-      if (overviewHeader) {
-        overviewHeader.classList.add('issue-overview-header');
-      }
-
       const mainTabLinks = document.querySelectorAll(MAIN_TAB_ITEM_SELECTOR);
       const mainTabContents = document.querySelectorAll(MAIN_TAB_CONTENT_SELECTOR);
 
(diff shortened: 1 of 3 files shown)


infrastructure/code/template/details.html
```diff
@@ -542,22 +605,101 @@ <h2 class="example-fixes-header">External example fixes</h2>
...
+    const EXAMPLE_TAB_NAV_SELECTOR = '.example-fixes-tabs-nav';
+    const EXAMPLE_TAB_ITEM_SELECTOR = '.example-fixes-tabs-nav .tab-item';
+    const EXAMPLE_TAB_CONTENT_SELECTOR = '.example-fixes .tab-content';
+    const IGNORE_WARNING_SELECTOR = '.ignore-warning-wrapper';
+    const FIRST_EXAMPLE_CONTENT_ID = '#tab-content-0';
+    const ISSUE_OVERVIEW_MD_HEADER = '.issue-overview .overview-text h2';
+
+    document.addEventListener('DOMContentLoaded', () => {
+      const overviewHeader = document.querySelector(ISSUE_OVERVIEW_MD_HEADER);
+      if (overviewHeader) {
```
teodora-sandu, line 622: Question: Why do we need to add this class conditionally?

infrastructure/code/template/details.html
```diff
@@ -542,22 +605,101 @@ <h2 class="example-fixes-header">External example fixes</h2>
...
+        const targetContentId = clickedTab.id.replace('tab-link-', 'tab-content-');
+
+        exampleTabLinks.forEach(tab => {
+          tab.classList.toggle('is-selected', tab === clickedTab);
+        });
+
+        exampleTabContents.forEach(content => {
+          content.classList.toggle('is-selected', content.id === targetContentId);
+          if (content.id === targetContentId) {
+            lastSelectedFixContent = content; // Track the last selected content
```
teodora-sandu, line 698: Question: What's the reason for having to store this?

infrastructure/code/template/details.html
```diff
@@ -542,22 +605,101 @@ <h2 class="example-fixes-header">External example fixes</h2>
...
-        content.classList.toggle('is-selected', content.id === targetContentId);
+      // Restore the last selected tab on page load
+      mainTabContents.forEach(content => {
+        content.classList.toggle('is-selected', content.getAttribute('data-content') === lastSelectedTab);
       });
-    });
 
+      mainTabLinks.forEach(tab => {
+        tab.classList.toggle('is-selected', tab.getAttribute('data-tab') === lastSelectedTab);
+      });
```
teodora-sandu, line 674: I just tested it and the old UI doesn't support this, as far as I can tell. What's the reason for adding this logic? I think it makes it a bit harder to understand so if you agree that the old UI doesn't have this, then we should remove it just to make our lives easier

infrastructure/code/template/details.html
```diff
@@ -385,10 +356,53 @@
...
+    .ignore-details-value {
+      flex-grow: 1;
+      text-align: left;
+      align-self: stretch;
+    }
   </style>
 </head>
 
-<body>
+<body class="snyk-ide">
```
teodora-sandu, line 405: Question: What's this here for?

teodora-sandu (approved):
Looks great! I think the JavaScript could be a bit simplified but if you agree with the comments I left then I could do that as part of my work. That way we can merge this PR, unblock your next ticket, and unblock my work too.  WDYT?

infrastructure/code/template/details.html
teodora-sandu: Actually, maybe I didn't understand what this does. Is it there to just make sure we don't end up with empty content under the example fix tab? 
Screenshot 2024-05-31 at 17 41 59

Either way it might be worth adding a comment here 😄

infrastructure/code/template/details.html
cat2608: It not a conditional addition. We are making sure that we are calling `.classList.add` only if the element exists in the `DOM`. 

When we query an element that is not yet set in the `DOM`, it returns `null`, which can cause an error that might stop the execution of subsequent JavaScript. In this example, the rest of the elements are no longer responsive:

image

The `DOMContentLoaded` event ensures that our script runs after the HTML is fully parsed. However, this doesn't always guarantee that every element we intend to manipulate is present. This is particularly critical for content coming from **markdown** in `additionalData.Text`, which we render in the **Overview** panel. For this case, _defensive programming_ is necessary.

infrastructure/code/template/details.html
cat2608: It a leftover from my tests with scoped CSS, we can delete it 💪

infrastructure/code/template/details.html
cat2608: By the way, my initial intention in adding the class name `issue-overview-header` to the `<h2>` tags sourced from the markdown in `additionalData.Text` was to ensure that these headings are styled consistently with the rest of the template. 

|`additionalData.Text`|Markdown Rendered Without Additional Styles|
|-|-|
|image||


However, I just realised that after your changes in #526, these headings and body will inherit the correct styles, won't they? 🙌 



|suggestion.css|
|-|
||

infrastructure/code/template/details.html
cat2608: I have refactored a bit more the code and added some comments. As I've shown before, I needed to find a way to keep the state of the latest content shown for the Code Diff suggestion and also to have it selected on page load.

This logic does not exist in the current VSCode implementation because we do not have tabs within tabs structure. Eventually we will replace the tabs for the Code Examples with the arrows.

infrastructure/code/template/details.html
```diff
@@ -438,91 +452,140 @@ <h2 class="severity-title">{{.IssueTitle}}</h2>
...
+        <span data-tab="fix-analysis" id="fix-analysis-tab"
+          class="tab-item is-selected sn-fix-analysis fix-analysis-tab">
+          Fix Analysis
+        </span>
+        <span data-tab="vuln-overview" id="vuln-overview-tab" class="tab-item sn-vuln-overview vuln-overview-tab">
+          Vulnerability Overview
+        </span>
+        {{end}}
       </div>
-    </article>
```
cat2608, line 446: Some class names were brought over from the current mark-up in the `vscode-plugin` project, and it's possible that not all of them are needed.

infrastructure/code/template/details.html
teodora-sandu: Possibly! I'll try to see if I can cleanup any redundant classes/styling

PR #527, opened 2024-05-31
feat: ai fix html [IDE-375]

teodora-sandu:
### Description
Moves over the common HTML and styling from VSCode so that eventually it can be used to render this panel in IntelliJ. In https://github.com/snyk/vscode-extension/pull/467 we can see the changes needed to add extra styling and javascript for VSCode and in https://github.com/snyk/snyk-intellij-plugin/pull/541 we completely hide this panel for IntelliJ.

### Checklist

- [x] Linted

### Screenshots / GIFs

See screenshots in linked PRs.

### DIFF
--- a/infrastructure/code/template/details.html
+++ b/infrastructure/code/template/details.html
@@ -118,6 +118,11 @@
       padding-right: 12px;
     }
 
+    .is-external-icon {
+      margin-bottom: 8px;
+      margin-left: 2px;
+    }
+
     .styled-link {
       text-decoration: none;
     }
@@ -663,10 +668,7 @@
       </div>
       <a class="lesson-link styled-link is-external" target="_blank" rel="noopener noreferrer" href="{{.LessonUrl}}">
         Learn about this vulnerability
-        <svg width="9" height="9" viewBox="0 0 9 9" xmlns="http://www.w3.org/2000/svg" fill="none">
-          <path d="M4.99998 0L6.64648 1.6465L3.14648 5.1465L3.85348 5.8535L7.35348 2.3535L8.99998 4V0H4.99998Z" fill="#888"/>
-          <path d="M8 8H1V1H4.5L3.5 0H1C0.4485 0 0 0.4485 0 1V8C0 8.5515 0.4485 9 1 9H8C8.5515 9 9 8.5515 9 8V5.5L8 4.5V8Z" fill="#888"/>
-        </svg>
+        {{.ExternalIcon}}
       </a>
     </div>
     {{end}}
--- a/infrastructure/code/template/details.html
+++ b/infrastructure/code/template/details.html
@@ -661,14 +661,13 @@
       <div class="lesson-icon">
         {{.LessonIcon}}
       </div>
-      <a class="lesson-link styled-link" target="_blank" rel="noopener noreferrer" href="{{.LessonUrl}}">Learn about
-        this vulnerability</a>
-      <div class="learn--link is-external">
+      <a class="lesson-link styled-link is-external" target="_blank" rel="noopener noreferrer" href="{{.LessonUrl}}">
+        Learn about this vulnerability
         <svg width="9" height="9" viewBox="0 0 9 9" xmlns="http://www.w3.org/2000/svg" fill="none">
           <path d="M4.99998 0L6.64648 1.6465L3.14648 5.1465L3.85348 5.8535L7.35348 2.3535L8.99998 4V0H4.99998Z" fill="#888"/>
           <path d="M8 8H1V1H4.5L3.5 0H1C0.4485 0 0 0.4485 0 1V8C0 8.5515 0.4485 9 1 9H8C8.5515 9 9 8.5515 9 8V5.5L8 4.5V8Z" fill="#888"/>
         </svg>
-      </div>
+      </a>
     </div>
     {{end}}
   </header>
--- a/infrastructure/code/template/details.html
+++ b/infrastructure/code/template/details.html
@@ -39,16 +39,6 @@
       background: #595a5c;
     }
 
-    html,
-    body {
-      height: 100%;
-      font-size: 16px;
-      display: flex;
-      flex-direction: column;
-      margin: 0;
-      padding: 0;
-    }
-
     *,
     *:before,
     *:after {
@@ -57,8 +47,6 @@
 
     body {
       font-family: var(--default-font);
-      font-weight: 400;
-      font-size: 0.875rem;
       overflow-y: auto;
       overflow-x: hidden;
     }
@@ -87,8 +75,8 @@
     .issue-overview-header,
     .ignore-details-header,
     .data-flow-header,
+    .ai-fix-header,
     .example-fixes-header {
-      font-size: 0.75rem;
       line-height: 1.25rem;
     }
 
@@ -108,7 +96,6 @@
       align-items: center;
       margin-bottom: 1em;
       line-height: 1em;
-      font-size: 0.85rem;
     }
 
     .lesson-container {
@@ -125,6 +112,10 @@
     .lesson-link {
       display: flex;
       align-items: center;
+    }
+
+    .is-external {
+      padding-right: 12px;
     }
 
     .styled-link {
@@ -156,7 +147,6 @@
       display: flex;
       align-items: center;
       justify-content: flex-start;
-      font-size: 0.85rem;
       line-height: 1.125rem;
       padding: 0.5em;
       border-radius: 0.25em;
@@ -165,10 +155,13 @@
     }
 
     .ignore-badge {
-      font-size: 0.75rem;
       padding: 0.35em 0.35em;
       border-radius: 0.25em;
       margin-left: 1em;
+    }
+
+    .position-line:hover{
+      cursor: pointer;
     }
 
     .data-flow-body {
@@ -226,7 +219,6 @@
     }
 
     .tab-item {
-      font-size: 0.75rem;
       font-weight: 500;
       float: left;
       display: block;
@@ -236,6 +228,10 @@
       margin-right: 4px;
       cursor: pointer;
       transition: background-color 0.3s, border-color 0.3s;
+    }
+
+    .tab-item-icon {
+      color: inherit;
     }
 
     .example-repo-link {
@@ -251,7 +247,6 @@
 
     .example-line {
       display: block;
-      padding: 2px 0 18px 0;
     }
 
     .example-line-number {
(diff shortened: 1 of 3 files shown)


infrastructure/code/template/details.html
```diff
@@ -396,6 +660,12 @@ <h2 class="severity-title">{{.IssueTitle}}</h2>
       </div>
       <a class="lesson-link styled-link" target="_blank" rel="noopener noreferrer" href="{{.LessonUrl}}">Learn about
         this vulnerability</a>
+      <div class="learn--link is-external">
+        <svg width="9" height="9" viewBox="0 0 9 9" xmlns="http://www.w3.org/2000/svg" fill="none">
+          <path d="M4.99998 0L6.64648 1.6465L3.14648 5.1465L3.85348 5.8535L7.35348 2.3535L8.99998 4V0H4.99998Z" fill="#888"/>
```
teodora-sandu, line 665: Adds the icon that marks the learn link as external

infrastructure/code/template/details.html
```diff
@@ -408,21 +678,13 @@ <h2 class="severity-title">{{.IssueTitle}}</h2>
...
-          Fix Analysis
-        </span>
-        <span data-tab="vuln-overview" id="vuln-overview-tab" class="tab-item vuln-overview-tab">
-          Vulnerability Overview
-        </span>
-        {{else}}
-        <span data-tab="fix-analysis" id="fix-analysis-tab"
-          class="tab-item is-selected sn-fix-analysis fix-analysis-tab">
+        {{end}}
+        <span data-tab="fix-analysis" id="fix-analysis-tab" class="tab-item {{if not .IsIgnored}}is-selected{{end}} sn-fix-analysis fix-analysis-tab">
```
teodora-sandu, line 682: Some refactoring here so we have less code duplication and warnings about using the same id twice

infrastructure/code/template/details.html
```diff
@@ -451,84 +713,169 @@ <h2 class="severity-title">{{.IssueTitle}}</h2>
...
-                      end-line="{{.EndLine}}" start-character="{{.StartCharacter}}" end-character="{{.EndCharacter}}">
-                      {{.FileName}}:{{.StartLineValue}}
-                    </td>
-                    <td class="data-flow-delimiter">|</td>
-                    <td class="data-flow-text">{{.Content}}</td>
-                  </tr>
-                  {{end}}
-                </tbody>
-              </table>
+          <section id="suggestion-info" class="delimiter-top">
```
teodora-sandu, line 716: Instead of deciding what to render when generating the template, we use classes to decide whether to hide panels if the issue has an AI fix. This way, in Intellij we can remove/add the `hidden` class such that we always hide AI fixes

infrastructure/code/template/details.html
```diff
@@ -39,16 +39,6 @@
       background: #595a5c;
     }
 
-    html,
```
teodora-sandu, line 42: These got moved to Intellij in https://github.com/snyk/snyk-intellij-plugin/pull/541 because in VSCode we have different styling as seen in https://github.com/snyk/vscode-extension/pull/467

infrastructure/code/template/details.html
```diff
@@ -600,8 +947,9 @@ <h2 class="example-fixes-header">Fixed Code Examples</h2>
       };
 
       // Event listeners for main and nested tabs
+      // TODO: move this into the HTML as an onclick
```
teodora-sandu, line 950: I think we can simplify a lot of this code by just adding an `onclick` attribute to these elements and in IntelliJ and VSCode we only implement the functions that these elements use. The changes are already complicated enough in this PR and its other two in our IDEs so I'd like to do that in a follow-up if there's time.

infrastructure/code/template/details.html
cat2608: What's your opinion about moving all the SVGs to `infrastructure/code/code_html.go` together with:
- Severity icons
- GitHub icon (I should have called it GitHub icon, sorry 🙏 )
- Lesson icon

https://github.com/snyk/snyk-ls/blob/4850808bc5da0b297fbc565b21f335999bd61cc3/infrastructure/code/code_html.go#L271-L286

cat2608: We need to wrap the SVG with the a tag so it becomes clickable:

infrastructure/code/template/details.html
cat2608: The mark-up is 😍 , thank you! 
It seems that there are some issues with the listeners injected from the IDE? You can see that error when switching between issues.

infrastructure/code/template/details.html
```diff
@@ -600,8 +950,9 @@ <h2 class="example-fixes-header">Fixed Code Examples</h2>
...
       document.querySelector(MAIN_TAB_NAV_SELECTOR).addEventListener('click', onMainTabClicked);
-      document.querySelector(EXAMPLE_TAB_NAV_SELECTOR).addEventListener('click', onExampleTabClicked);
+      document.querySelector(EXAMPLE_TAB_NAV_SELECTOR)?.addEventListener('click', onExampleTabClicked);
```
cat2608, line 955: Maybe we need to go for something similar to this?

https://github.com/snyk/snyk-ls/pull/526#discussion_r1623497561

infrastructure/code/template/details.html
teodora-sandu: What is the reason for having the icons in there? I chose not to put it in that file because I didn't see the point of it.

teodora-sandu: The icon was not clickable in the old view either, so I didn't make it clickable here

infrastructure/code/template/details.html
teodora-sandu: How did you manage to test this view? The consistent ignores backend is broken at the moment (it was broken on Friday and I explained in the PR description and it seems still broken today), so I haven't been able to test this part of the UI (where we have an ignored issue)

infrastructure/code/template/details.html
cat2608: - In the current implementation, the icon is clickable:

https://github.com/snyk/snyk-ls/assets/1948377/3c6a40fe-01cb-4f08-8406-53f9cbde82f0


- **SVG location**: with the SVGs and additional styling, the `details.html` has almost 1000 lines. Modularizing some bits makes the file more manageable and easier to maintain:

  - **Reusability**: Breaking down the code into smaller components makes it easier to understand, maintain, and update reducing the complexity of the `details.html` file.
  - **Scalability**: Moving SVGs to a separate file allows for reusability across the project, ensuring consistency and reducing redundancy. As and example of reusability, you wisely commented in [here](https://github.com/snyk/snyk-intellij-plugin/pull/535#discussion_r1621929683): "_in the future we will have more than Snyk Code styling but still want to share stylesheets or colours across Snyk OSS, Snyk IaC, and Snyk Container_".
  - **Collaboration**: Clean, modular code improves collaboration among team members by reducing merge conflicts and making it easier to work on the same project.

infrastructure/code/template/details.html
cat2608: In dev, everything is working fine 🙌

cat2608: Also, did we say that is the issue is ignored, we don't show AI Fixes? I'm not sure now and I can't find references to this case.

infrastructure/code/template/details.html
teodora-sandu: Having the SVGs in the Go code is not the nicest: you don't benefit from the HTML syntax highlighting and we're only moving the complexity from one file to another, which is not really the right file to hold this kind of code. I had a look at other ways to do this. Maybe when we end up re-using this code for other product lines we could look at doing what they suggest here: https://stackoverflow.com/questions/24093923/optimising-html-template-composition. For now I'll do what you suggested to be consistent 👍

infrastructure/code/template/details.html
teodora-sandu: Based on the designs we're meant to show the `Fix Analysis` tab: https://www.figma.com/design/hcI2QHUtHfcIjgrpYlMqff/Holistic-Ignores?node-id=2292-64060&t=nsuVsneQwSDwzp71-4. I'll change it in this PR

infrastructure/code/template/details.html
cat2608: Yeah! That makes a lot of sense 🙌  There are a few [downsides](https://stackoverflow.com/a/5448974) when using CSS rules to hide components, but I thing for our context that is the IDE's land, we don't need to care about:

- **Initial Load**: All hidden elements are still sent to the IDE, which can increase the initial load time and the size of the HTML document. 
- **Rendering Overhead**: The IDE still needs to parse and render these hidden elements, which can add some overhead and potentially slow down the IDE, especially in larger projects.
- **Data Exposure**: For our case, I guess it's not a big problem but all the hidden elements are still accessible from the DevTool meaning that with just removing the class `hidden` the element is exposed.

cat2608 (approved):
All rendering looks awesome! Just the JS seems to not work sometimes. I'm approving it so we can continue it in later iterations since you said you'd like to inline the onclicks :)

infrastructure/code/template/details.html
teodora-sandu: Thanks it does work in dev indeed! I've pushed a fix in https://github.com/snyk/vscode-extension/pull/467/commits/b21e4068514cf60caa82c48230e71b0c7c87c734

infrastructure/code/template/details.html
teodora-sandu: Yep in this case, since it's temporary situation, this is okay 😄

teodora-sandu:
> Just the JS seems to not work sometimes.

Which JS is that @cat2608 ? I probably will raise a ticket to add the onclicks because I think the manual testing of the feature is more important than code improvements at this moment, so if there are parts of the JS that don't work maybe I can work on those instead

cat2608:
@teodora-sandu this was my setup when testing your change and when I left my comment [here](https://github.com/snyk/snyk-ls/pull/527#discussion_r1632828256)

- LS: `feat/ai-fix-html` at aec12f511bd9d66ebee0dd6aac2a26917b9ff400
- vscode-plugin: `feat/ai-fix-html` at `eb5c2aee5a60eea04288c0721d0909a1c5357688` 

https://github.com/snyk/snyk-ls/assets/1948377/c37c087b-da12-4ff6-9944-522b76b2a231

---

Now I can see that you have added `b21e4068514cf60caa82c48230e71b0c7c87c734` in `vscode-plugin` which reads `fix: add event listener conditionally`. I tested it again and the issue about the event listeners has gone! 🙌 

The only thing remaining is the Fix Action itself:

https://github.com/snyk/snyk-ls/assets/1948377/82cfa715-bf69-46fd-9c73-1ca951c045f3

teodora-sandu:
Ah the error you're seeing is a backend error, the call to get the fix diffs seems to return `null`. The AI fixes were not working on Friday and they seem to still be slightly broken today. I will cover that in https://snyksec.atlassian.net/browse/IDE-380. If you apply the patch I provide in the PR description it does work correctly, because it's using the old backend

PR #529, opened 2024-06-03
refactor: js logic for swapping tabs

teodora-sandu:
### Description

Refactors the logic for switching tabs in HTML for both the main tabs and the example commit fixes tabs.

The selectors we were using were not specific enough so when switching between the main tabs, we would remove the `is-selected` class from the example commit fixes tabs too. By adding a more specific class to each tab and selecting based on that, we don't need to have complicated logic for remembering what the last selected tab was. This greatly simplifies the code in the JS.

### Checklist

- [x] Linted

### Screenshots / GIFs

https://github.com/snyk/snyk-ls/assets/81559517/14bcc012-0a84-410a-970f-8f5aee03ae1a

### DIFF
--- a/infrastructure/code/template/details.html
+++ b/infrastructure/code/template/details.html
@@ -418,7 +418,7 @@
       <div class="tab-container">
         {{if .IsIgnored}}
         <!-- Start Ignore Details section for tab id="ignore-details-tab" -->
-        <div data-content="ignore-details" id="ignore-details-content" class="tab-content is-selected">
+        <div data-content="ignore-details" id="ignore-details-content" class="tab-content main-tab-content is-selected">
           <div class="ignore-details">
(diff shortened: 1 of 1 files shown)


cat2608 (approved):
🚀 Awesome! Thank you!

Just a note about my environment when testing this:

|Project|Branch|
|-|-|
|`snyk-ls`|`refactor/html-tab-js-logic`|
|`vscode-extension`|`main`|
|`cli`|`main`|
|`snyk-intellij-plugin`|`master`|


Using my local `snyk-ls` in your branch, for IntelliJ I see the template with zero styling - It's expected, right?

teodora-sandu:
Yep I have't updated the `cli` to include the latest `snyk-ls` yet so that explains why IntelliJ doesn't have it yet @cat2608 😄

PR #534, opened 2024-06-05
fix: support multiple date formats for ignored-on field in API response [IDE-388]

cat2608:
### Description

- Change `IgnoredOn` from `time.Time` to `string` to avoid parsing errors.
- Use date string as received with a fallback to a formatted version of today's date
- Update error log

> [!TIP]
**How to test it**
If you have ignored an issue before today, you can compare VSCode PROD and this branch and see that we don't fallback to today's day if the attribute exists and it's valid

### Checklist

- [ ] Linted

### Screenshots / GIFs

|Before|After|
|-|-|
|before-ignored-on-fix|after-ignored-on-fix|

> [!NOTE]  
> Data from the API
image

### DIFF
--- a/infrastructure/code/convert.go
+++ b/infrastructure/code/convert.go
@@ -412,24 +412,34 @@ func (s *SarifConverter) getIgnoreDetails(result codeClientSarif.Result) (bool, 
 		if suppression.Properties.Expiration != nil {
 			expiration = *suppression.Properties.Expiration
 		}
-		ignoredOn, err := time.Parse(time.RFC3339, suppression.Properties.IgnoredOn)
-		if err != nil {
-			// We don't want to fail just because of this parsing logic
-			log.Error().
-				Err(err).
-				Msg("failed to parse ignoredOn timestamp " +
-					suppression.Properties.IgnoredOn)
-			ignoredOn = time.Now()
-		}
+
 		ignoreDetails = &snyk.IgnoreDetails{
 			Category:   string(suppression.Properties.Category),
 			Reason:     suppression.Justification,
 			Expiration: expiration,
-			IgnoredOn:  ignoredOn,
+			IgnoredOn:  parseDateFromString(suppression.Properties.IgnoredOn),
 			IgnoredBy:  suppression.Properties.IgnoredBy.Name,
 		}
 	}
 	return isIgnored, ignoreDetails
+}
+
+func parseDateFromString(date string) time.Time {
+	logger := log.With().Str("method", "convert.parseDateFromString").Logger()
+	layouts := []string{
+		"Mon Jan 02 2006", // TODO: when this gets fixed, we can remove this option [IGNR-365]
+		time.RFC3339,      // Standard format
+	}
+
+	for _, layout := range layouts {
+		if t, err := time.Parse(layout, date); err == nil {
+			return t
+		}
+	}
+
+	// Fallback to today's date if parsing fails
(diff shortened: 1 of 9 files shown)


infrastructure/code/convert_test.go
```diff
@@ -956,7 +957,39 @@ func Test_Result_getIgnoreDetails(t *testing.T) {
 		assert.Equal(t, "reason", ignoreDetails.Reason)
 		assert.Equal(t, "category", ignoreDetails.Category)
 		assert.Equal(t, "expiration", ignoreDetails.Expiration)
-		assert.Equal(t, 2024, ignoreDetails.IgnoredOn.Year())
+		assert.Equal(t, "June 5, 2024", ignoreDetails.IgnoredOn) // Expected formatted date
```
cat2608, line 960: I think this assertion could fail next year

infrastructure/code/convert.go
```diff
@@ -412,20 +413,21 @@ func (s *SarifConverter) getIgnoreDetails(result codeClientSarif.Result) (bool,
 		if suppression.Properties.Expiration != nil {
 			expiration = *suppression.Properties.Expiration
 		}
-		ignoredOn, err := time.Parse(time.RFC3339, suppression.Properties.IgnoredOn)
+		ignoredOn, err := time.Parse("Mon Jan 02 2006", suppression.Properties.IgnoredOn)
```
cat2608, line 416: This change is because the date is coming as a string in this format: `"Wed Jun 05 2024"`

infrastructure/code/fake_code_client_scanner.go
```diff
@@ -292,7 +292,7 @@ func getSarifResponseJson2(filePath string) string {
 								"properties": {
 									"category": "wont-fix",
 									"expiration": "13 days",
-									"ignoredOn": "2024-02-23T16:08:25Z",
+									"ignoredOn": "Wed Jun 05 2024",
 									"ignoredBy": {
 										"name": "Neil M",
 										"email": "test@test.io"
```
cat2608, line 298: IDE-258

cmars: Our REST APIs should be consistent on date time formatting, expressed in the OpenAPI spec as:

```
type: string
format: date-time
```

So the original was correct. This looks like a regression or bug in the API. Please open a bug with the API owner 🙏

cmars (changes requested):
API should be using RFC3339, not a human-readable (and highly locale subjective) string representation.

infrastructure/code/fake_code_client_scanner.go
cmars: Oh... this is SARIF. Hmm. Consulting the SARIF spec.

cmars: Looks like this is free-form metadata in the suppression object. I think we should still push for RFC3339 (unambigious, machine readable, consistent with our other APIs...)

teodora-sandu (approved):
Looks good! I think you might need to change the type in https://github.com/snyk/snyk-intellij-plugin/blob/ffb78950ade7b3d1b8cbc325cb9161465216659c/src/main/kotlin/snyk/common/lsp/Types.kt#L545 too. In VSCode it doesn't look like we define it anywhere so you're good

cat2608:
Following @cmars suggestion, I have asked to the API owners to please review this attribute. More details [here](https://snyksec.atlassian.net/browse/IGNR-365)

PR #542, opened 2024-06-11
feat: change dataflow based on designs [IDE-382]

teodora-sandu:
### Description

This PR refactors the dataflow section to match the new [designs](https://www.figma.com/design/hcI2QHUtHfcIjgrpYlMqff/Holistic-Ignores?node-id=2292-63654&t=iIxZM8o3pih8EY4M-0).

I haven't been able to test this for dataflow that have multiple files. This is because `juice-goof` is failing to scan at the moment. 

### Checklist

- [x] Linted

### Screenshots / GIFs
Screenshot 2024-06-11 at 16 27 42
Screenshot 2024-06-11 at 17 15 14


infrastructure/code/template/details.html
```diff
@@ -824,18 +855,24 @@ <h2 class="data-flow-header">
...
+                      <tr class="data-flow-row">
+                        <td class="data-flow-number">{{.Number}}</td>
+                        <td class="data-flow-clickable-row" file-path="{{.FilePath}}" start-line="{{.StartLine}}"
+                          end-line="{{.EndLine}}" start-character="{{.StartCharacter}}" end-character="{{.EndCharacter}}">
+                          line:{{.StartLineValue}}
+                        </td>
+                        <td class="data-flow-delimiter">|</td>
+                        <td class="data-flow-text">{{.Content}}</td>
+                      </tr>
+                    {{end}}
```
cat2608, line 875: I replicated two folders from Juice Shop and it seems that we are not placing the file names correctly between blocks:

- Figma design:

Data Flow Figma

- IntelliJ
image

I pushed this folders to the testing repo I shared with you :)

infrastructure/code/template/details.html
```diff
@@ -60,6 +60,10 @@
       overflow-y: auto;
     }
 
+    section {
+      padding: 10px;
+    }
+
```
cat2608, line 66: Looking great! 
Do you think we could add some rules to the Data Flow section to make it horizontally scrollable? 

I quickly tested these [rules](https://www.geeksforgeeks.org/make-a-div-horizontally-scrollable-using-css/) in VSCode and IntelliJ and I compared with and without the rules and they look like the following:


https://github.com/snyk/snyk-ls/assets/1948377/de8492f3-2a85-4548-aec8-488143b4837b


https://github.com/snyk/snyk-ls/assets/1948377/e83f7b4c-68c4-4da9-b100-16a7c4553898



For IntelliJ we would need to style the scrollbar too 😬

cat2608 (approved):
The functionality works great! Left a comment in case the layout could be improved.

infrastructure/code/template/details.html
teodora-sandu: I remember we have a ticket to take care of this: https://snyksec.atlassian.net/browse/IDE-256. It would make sense to make it consistent with what's already in there in the old UI, which I would need to check if it supports scrolling. I think we could also just wrap the text to the next line if it's too long, which is what I preferred. But basically I think I'd prefer to do it in another ticket if that's okay?

cat2608: It's approved! As you like 🙌 
Thank you

PR #544, opened 2024-06-14
feat: allow oauth2 authentication

bastiandoetsch:
### Description

This PR overhauls the authentication in Language Server.

1. By default, it registers OAuth2 as primary provider and Token (CLI) Authentication as secondary. If OAuth2 fails, it automatically falls back to token authentication.
2. By specifying `token` as authentication method during initialization, or as configuration update, Token-only authentication can be forced.
3. The whole authentication & initialization code has been moved to streamline dependencies (top->down). Authentication does not have backlinks to the snyk domain anymore.

### Checklist

- [x] Linted

### Screenshots / GIFs


domain/snyk/authentication_functions.go
```diff
@@ -42,11 +40,6 @@ func GetActiveUser() (*ActiveUser, error) {
 		return nil, errors.New("no credentials found")
 	}
 	conf := c.Engine().GetConfiguration().Clone()
-	if c.AuthenticationMethod() == lsp.OAuthAuthentication {
```
teodora-sandu, line 45: Question: Because we're enabling it by default we don't need to set these as they're configured already [here](https://github.com/snyk/go-application-framework/blob/cc0fb551c613340e4f9c5bca833e2de4caf48210/pkg/app/app.go#L185) right?

infrastructure/cli/environment_test.go
```diff
@@ -41,11 +45,22 @@ func TestAddConfigValuesToEnv(t *testing.T) {
 		c.SetIntegrationVersion(expectedIntegrationVersion)
 		c.SetIdeVersion(expectedIdeVersion)
 		c.SetIdeName(expectedIdeName)
+		s := oauth2.Token{
```
teodora-sandu, line 47: Should we keep the old test and just add a new oauth2 env test?

teodora-sandu (approved):
Works so I'm approving it but I left a comment to try and figure out a way to make this code a bit easier to understand. Not a blocker though!

infrastructure/authentication/auth_configuration.go
```diff
@@ -0,0 +1,94 @@
+/*
```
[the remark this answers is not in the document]
bastiandoetsch: Hm... will have a look. Sounds like a good idea, will see if we can do that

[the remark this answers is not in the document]
bastiandoetsch: @teodora-sandu does that look better to you?

domain/snyk/authentication_functions.go
bastiandoetsch: I ran into some problems when not setting it explicitly, so I added it. In general, the FF should be removed from GAF.

infrastructure/cli/environment_test.go
bastiandoetsch: Hm, the expected default behaviour changes, so I would change this test, too.

bastiandoetsch: I can add a test for the Token auth, though :)

PR #547, opened 2024-06-18
fix: only log msg when not succeeding to send event to endpoint [IDE-418]

acke:
### Description

LS should not return an error to the IDEs if the analytics service isn't available or the call returns an error. Instead, it should just log.

### Checklist

- [x] Linted

### Screenshots / GIFs

### DIFF
--- a/infrastructure/analytics/analytics.go
+++ b/infrastructure/analytics/analytics.go
@@ -46,7 +46,6 @@ func SendAnalyticsToAPI(c *config.Config, payload []byte) error {
 
 	if err != nil {
 		logger.Err(err).Msg("error invoking workflow")
-		return err
 	}
 	return nil
 }


bastiandoetsch (approved):

infrastructure/analytics/analytics.go
```diff
@@ -46,7 +46,6 @@ func SendAnalyticsToAPI(c *config.Config, payload []byte) error {
 
 	if err != nil {
 		logger.Err(err).Msg("error invoking workflow")
-		return err
```
cmars, line 49: Please add a comment why we don't return the error here.

acke: LS returns an error the IDEs if analytics-service not present. This should not happen - it should fail silently, with just a log message. As Basti pointed out.

PR #549, opened 2024-06-19
feat: create script / workflows for automatic PR creation

bastiandoetsch:
### Description

This adds automation for creating a CLI PR on releasing language server. It also provides a script that can be triggered manually to create a PR with the latest released LS version.

### Checklist

- [ ] Linted

### Screenshots / GIFs

### DIFF
--- a/.github/create-cli-pr.sh
+++ b/.github/create-cli-pr.sh
@@ -1,0 +1,31 @@
+#
+# © 2024 Snyk Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


ShawkyZ (approved):
Looks good 
Thank you !

Maybe in the future we could move create-cli-pr to a common repo to to also use it in code-client-go and other repos eventually.
It would also be cool if the command that needs to run to upgrade dependencies (for example scripts/upgrade-snyk-go-dependencies.go) would come as an argument to the workflow if triggered from another workflow.

PR #552, opened 2024-06-19
fix: consistent ignores styling [IDE-386]

teodora-sandu:
### Description

Changes some colours in order to match the [designs](https://www.figma.com/design/hcI2QHUtHfcIjgrpYlMqff/Holistic-Ignores?node-id=2292-61688&t=larzTCeHFfRtz6av-0)

I tried to split into commits but there were so many small changes that I ended up combining some of them. The best way to review this is to run the code with https://github.com/snyk/vscode-extension/pull/476 and https://github.com/snyk/snyk-intellij-plugin/pull/552 and compare the designs.

I mostly address the different styling improvements called out in https://snyksec.atlassian.net/browse/IDE-386, but also some other aspects I noticed while comparing the UI to the designs. 

The delimiters were hidden before (as they're not there in the original UI) but the designs make them visible, so I did too in this PR.

### Checklist

- [x] Linted


cat2608 (approved):
🔝

PR #554, opened 2024-06-20
feat: add commits in PR body

bastiandoetsch:
### Description

image

### Checklist

- [ ] Linted

### Screenshots / GIFs

### DIFF
--- a/.github/create-cli-pr.sh
+++ b/.github/create-cli-pr.sh
@@ -1,3 +1,4 @@
+#!/bin/bash
 #
 # © 2024 Snyk Limited
 #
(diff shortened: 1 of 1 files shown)


.github/create-cli-pr.sh
```diff
@@ -14,30 +15,37 @@
...
-gh repo clone git@github.com:snyk/cli.git $CLI_DIR -- --depth=1
-pushd $CLI_DIR
+
+
+gh repo clone git@github.com:snyk/cli.git "$CLI_DIR" -- --depth=1
+pushd "$CLI_DIR/cliv2"
+  LS_COMMIT_HASH=$(grep snyk-ls go.mod| cut -d "-" -f 4)
+popd
+
+WHAT_CHANGED=$(git whatchanged "$LS_COMMIT_HASH"...HEAD)
```
ShawkyZ, line 28: should we use git log instead of git whatchanged? 
> The command is kept primarily for historical reasons;

https://stackoverflow.com/a/18585297
according to this answer equivelant should be 
```git log --raw --no-merges```

ShawkyZ (approved):

.github/create-cli-pr.sh
bastiandoetsch: Interesting. I like `whatchanged` better, as it's quite explicit :)

PR #558, opened 2024-06-24
fix: configure AI fix boolean even if issue is ignored [IDE-435]

teodora-sandu:
### Description

We want to continue hiding the code actions and code lenses for ignored findings that have AI fixes. But we want the Fix Analysis tab to show AI fixes, so that issues that are temporarily ignored because of noise can be fixed if the developer chooses to. So the `HasAIFix` boolean needs to be configured.

Also sets the reason for an ignore to a default `None given` which matches what the UI shows.

### Checklist

- [x] Linted

### DIFF
--- a/infrastructure/code/issue_enhancer_test.go
+++ b/infrastructure/code/issue_enhancer_test.go
@@ -241,7 +241,7 @@ func Test_addIssueActions(t *testing.T) {
 		assert.Len(t, fakeIssues[0].CodeActions, 2)
 	})
 
-	t.Run("Includes AI fixes if issue is not autofixable", func(t *testing.T) {
+	t.Run("Does not include AI fixes if issue is not autofixable", func(t *testing.T) {
 		setupCodeSettings()
 		defer t.Cleanup(resetCodeSettings)
 		fakeIssues := setupFakeIssues(false, false)
@@ -255,7 +255,7 @@ func Test_addIssueActions(t *testing.T) {
 		assert.Len(t, fakeIssues[0].CodeActions, 1)
 	})
 
-	t.Run("Does not include AI fixes if issue is ignored", func(t *testing.T) {
+	t.Run("Does not include AI fixes even if it is autofixable if issue is ignored", func(t *testing.T) {
 		setupCodeSettings()
 		defer t.Cleanup(resetCodeSettings)
 		fakeIssues := setupFakeIssues(true, true)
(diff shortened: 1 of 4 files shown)


application/server/server_smoke_test.go
```diff
@@ -136,11 +144,12 @@ func Test_SmokeWorkspaceScan(t *testing.T) {
 			file2:                "maven-compat/src/test/java/org/apache/maven/repository/legacy/LegacyRepositorySystemTest.java",
 			useConsistentIgnores: true,
 			hasVulns:             true,
+			endpoint:             "https://api.dev.snyk.io",
 		},
 	}
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
-			runSmokeTest(t, tc.repo, tc.commit, tc.file1, tc.file2, tc.useConsistentIgnores, tc.hasVulns, "")
+			runSmokeTest(t, tc.repo, tc.commit, tc.file1, tc.file2, tc.useConsistentIgnores, tc.hasVulns, tc.endpoint)
```
teodora-sandu, line 152: Although we were configuring the endpoint for the `OSS and Code with V1 endpoint` test we weren't passing it down to the function and so always using the non-v1 endpoint.

infrastructure/code/issue_enhancer_test.go
```diff
@@ -264,7 +264,7 @@ func Test_addIssueActions(t *testing.T) {
 
 		issueData, ok := fakeIssues[0].AdditionalData.(snyk.CodeIssueData)
 		require.True(t, ok)
-		assert.False(t, issueData.HasAIFix)
+		assert.True(t, issueData.HasAIFix)
```
cat2608, line 267: With this PR change, is it correct to say that this unit test description needs to be updated too?

It currently reads:

    "Does not include AI fixes if issue is ignored"

However, as far as I understand, we'd like to assert that:

    "Includes AI fixes even if issue is ignored"

Is this correct?

teodora-sandu: Not exactly, the AI fixes are still not included in the code actions and code lenses. Which is what this test verifies. The side-effect of setting `HasAIFix` in this code is not great but I didn't want to refactor the code smell right now since we're in a bit of a time crunch to get this into the CLI before the 26th of June

cat2608: Then updating the unit description to something more align to this context also does not work? 

It's just that the title says `Does not include AI fixes if issue is ignored` and the assertion is `assert.True(t, issueData.HasAIFix)`.

A description like:

```
Sets HasAIFix to true but does not include AI fixes in Code Actions or Code Lenses if issue is ignored
```
could maybe work?

> [!NOTE]  
> This is not a change request

teodora-sandu: I just think it's a bit too verbose. I've changed the test name a bit (and for the one above it, since it was just incorrect) but I don't specify HasAIFix in it. Thanks for the suggestion!

PR #559, opened 2024-06-24
feat: warn, if LS does not match client requirements

bastiandoetsch:
### Description

Language Server & IDEs synchronize over the LsProtocolVersion. If these mismatch, functionality could be buggy or not working. This PR enables warnings to the user, if a client specifies its required protocol version:

- new setting option in InitializationOptions to transmit the required protocol version from the client
- language server checks its own protocol version, and if not the same, sends a `ShowMessageRequest` language server protocol dialog to the client
- the dialog contains a button to open a web browser to download either the correct & needed CLI or the standalone language server.
- The download URL takes release channels into account

### Checklist

- [x] Linted

### Screenshots / GIFs

image

image


infrastructure/cli/install/releases.go
```diff
@@ -17,16 +17,23 @@
...
 	"io"
 	"net/http"
+	"path"
+	"runtime"
+	"strings"
+	"time"
 
+	"github.com/pkg/errors"
+
+	http2 "github.com/snyk/code-client-go/http"
```
ShawkyZ, line 32: why do we need this here instead of just using http.Client ?

ShawkyZ (approved):

PR #563, opened 2024-06-26
fix: trigger folder scan when `.snyk` file changes [IDE-253]

cat2608:
### Description

This PR triggers a workspace scan when the `.snyk` file was saved.

### Checklist

- [x] Linted

### Screenshots / GIFs

https://github.com/snyk/snyk-ls/assets/1948377/d9f761f9-d692-4c48-b77e-15bc5181896a

### DIFF
--- a/application/server/server.go
+++ b/application/server/server.go
@@ -540,6 +540,14 @@ func textDocumentDidOpenHandler() jrpc2.Handler {
 	})
 }
 
+func isDotSnykFile(uri sglsp.DocumentURI) bool {
+	return strings.HasSuffix(string(uri), ".snyk")
+}
+
+func dotSnykFileDidChangeScan(ctx context.Context, workspaceFolderPath *workspace.Folder) {
+	go workspaceFolderPath.ScanFolder(ctx)
+}
+
 func textDocumentDidSaveHandler() jrpc2.Handler {
 	return handler.New(func(_ context.Context, params sglsp.DidSaveTextDocumentParams) (any, error) {
 		// The context provided by the JSON-RPC server is canceled once a new message is being processed,
@@ -553,6 +561,11 @@ func textDocumentDidSaveHandler() jrpc2.Handler {
 		filePath := uri.PathFromUri(params.TextDocument.URI)
 
 		f := workspace.Get().GetFolderContaining(filePath)
+
+		if isDotSnykFile(params.TextDocument.URI) {
+			dotSnykFileDidChangeScan(bgCtx, f)
+			return nil, nil
+		}
 
 		autoScanEnabled := cfg.IsAutoScanEnabled()
 		if f != nil {
--- a/application/server/server.go
+++ b/application/server/server.go
@@ -554,13 +554,13 @@ func textDocumentDidSaveHandler() jrpc2.Handler {
 		di.FileWatcher().SetFileAsSaved(params.TextDocument.URI)
 		filePath := uri.PathFromUri(params.TextDocument.URI)
 
-		ws := workspace.Get()
-		if ws != nil && autoScanEnabled && uri.IsDotSnykFile(params.TextDocument.URI) {
-			go ws.ScanWorkspace(bgCtx)
+		f := workspace.Get().GetFolderContaining(filePath)
+
+		if f != nil && autoScanEnabled && uri.IsDotSnykFile(params.TextDocument.URI) {
+			go f.ScanFolder(bgCtx)
 			return nil, nil
 		}
 
-		f := ws.GetFolderContaining(filePath)
 		if f != nil {
 			if autoScanEnabled {
 				go f.ScanFile(bgCtx, filePath)
--- a/application/server/server.go
+++ b/application/server/server.go
@@ -540,34 +540,27 @@ func textDocumentDidOpenHandler() jrpc2.Handler {
 	})
 }
 
-func isDotSnykFile(uri sglsp.DocumentURI) bool {
-	return strings.HasSuffix(string(uri), ".snyk")
-}
-
-func dotSnykFileDidChangeScan(ctx context.Context, workspaceFolderPath *workspace.Folder) {
-	go workspaceFolderPath.ScanFolder(ctx)
-}
-
 func textDocumentDidSaveHandler() jrpc2.Handler {
 	return handler.New(func(_ context.Context, params sglsp.DidSaveTextDocumentParams) (any, error) {
 		// The context provided by the JSON-RPC server is canceled once a new message is being processed,
 		// so we don't want to propagate it to functions that start background operations
 		bgCtx := context.Background()
-		cfg := config.CurrentConfig()
-		logger := cfg.Logger().With().Str("method", "TextDocumentDidSaveHandler").Logger()
-
+		c := config.CurrentConfig()
+		logger := c.Logger().With().Str("method", "TextDocumentDidSaveHandler").Logger()
 		logger.Info().Interface("params", params).Msg("Receiving")
+
+		autoScanEnabled := c.IsAutoScanEnabled()
+
 		di.FileWatcher().SetFileAsSaved(params.TextDocument.URI)
 		filePath := uri.PathFromUri(params.TextDocument.URI)
 
-		f := workspace.Get().GetFolderContaining(filePath)
-
-		if isDotSnykFile(params.TextDocument.URI) {
-			dotSnykFileDidChangeScan(bgCtx, f)
+		ws := workspace.Get()
+		if ws != nil && autoScanEnabled && uri.IsDotSnykFile(params.TextDocument.URI) {
+			go ws.ScanWorkspace(bgCtx)
 			return nil, nil
 		}
 
-		autoScanEnabled := cfg.IsAutoScanEnabled()
+		f := ws.GetFolderContaining(filePath)
 		if f != nil {
 			if autoScanEnabled {
 				go f.ScanFile(bgCtx, filePath)
--- a/application/server/server.go
+++ b/application/server/server.go
@@ -545,23 +545,24 @@ func textDocumentDidSaveHandler() jrpc2.Handler {
 		// The context provided by the JSON-RPC server is canceled once a new message is being processed,
 		// so we don't want to propagate it to functions that start background operations
 		bgCtx := context.Background()
-		c := config.CurrentConfig()
-		logger := c.Logger().With().Str("method", "TextDocumentDidSaveHandler").Logger()
+		cfg := config.CurrentConfig()
+		logger := cfg.Logger().With().Str("method", "TextDocumentDidSaveHandler").Logger()
 
 		logger.Info().Interface("params", params).Msg("Receiving")
 		di.FileWatcher().SetFileAsSaved(params.TextDocument.URI)
 		filePath := uri.PathFromUri(params.TextDocument.URI)
 
 		f := workspace.Get().GetFolderContaining(filePath)
-		autoScanEnabled := config.CurrentConfig().IsAutoScanEnabled()
-		if f != nil && autoScanEnabled {
-			go f.ScanFile(bgCtx, filePath)
-		} else {
+
+		autoScanEnabled := cfg.IsAutoScanEnabled()
+		if f != nil {
 			if autoScanEnabled {
-				logger.Warn().Str("documentURI", filePath).Msg("Not scanning, file not part of workspace")
+				go f.ScanFile(bgCtx, filePath)
 			} else {
 				logger.Warn().Msg("Not scanning, auto-scan is disabled")
 			}
+		} else if autoScanEnabled {
+			logger.Warn().Str("documentURI", filePath).Msg("Not scanning, file not part of workspace")
 		}
 		return nil, nil
 	})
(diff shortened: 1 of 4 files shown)


application/server/server.go
```diff
@@ -477,28 +477,42 @@ func textDocumentDidOpenHandler() jrpc2.Handler {
 	})
 }
 
+func isDotSnykFile(uri sglsp.DocumentURI) bool {
```
bastiandoetsch, line 480: should we move this to internal/uri or infrastructure/filefilter?

application/server/server.go
```diff
@@ -477,28 +477,42 @@ func textDocumentDidOpenHandler() jrpc2.Handler {
...
+	return strings.HasSuffix(string(uri), ".snyk")
+}
+
+func dotSnykFileDidChangeScan(ctx context.Context, workspaceFolderPath *workspace.Folder) {
+	go workspaceFolderPath.ScanFolder(ctx)
```
bastiandoetsch, line 485: Maybe it would be easier/better to use the scan.workspaceFolder command, as there may be additional logic that would be beneficial to be executed - haven't checked, but would be possible that e.g. Trust is handled etc.

application/server/server.go
```diff
@@ -477,28 +477,42 @@ func textDocumentDidOpenHandler() jrpc2.Handler {
...
+}
+
 func textDocumentDidSaveHandler() jrpc2.Handler {
 	return handler.New(func(_ context.Context, params sglsp.DidSaveTextDocumentParams) (any, error) {
 		// The context provided by the JSON-RPC server is canceled once a new message is being processed,
 		// so we don't want to propagate it to functions that start background operations
 		bgCtx := context.Background()
-		c := config.CurrentConfig()
-		logger := c.Logger().With().Str("method", "TextDocumentDidSaveHandler").Logger()
+		cfg := config.CurrentConfig()
```
bastiandoetsch, line 493: usually, it's `c` everywhere - so I'd keep it like that :)

application/server/server.go
```diff
@@ -477,28 +477,42 @@ func textDocumentDidOpenHandler() jrpc2.Handler {
...
 		filePath := uri.PathFromUri(params.TextDocument.URI)
 
 		f := workspace.Get().GetFolderContaining(filePath)
-		autoScanEnabled := config.CurrentConfig().IsAutoScanEnabled()
-		if f != nil && autoScanEnabled {
-			go f.ScanFile(bgCtx, filePath)
-		} else {
+
+		if isDotSnykFile(params.TextDocument.URI) {
+			dotSnykFileDidChangeScan(bgCtx, f)
```
bastiandoetsch, line 503: I think this should be dependent on the auto-scan setting too. Although this scenario is right now not defined. I wonder though how we transmit the need to rescan to get new results. Maybe a notification? WDYT?

application/server/server.go
bastiandoetsch, line 485: This is a potential panic, if the pointer is nil (e.g. when the file is not contained in the workspaceFolder.

application/server/server.go
```diff
@@ -477,28 +477,42 @@ func textDocumentDidOpenHandler() jrpc2.Handler {
...
+
+		autoScanEnabled := cfg.IsAutoScanEnabled()
+		if f != nil {
 			if autoScanEnabled {
-				logger.Warn().Str("documentURI", filePath).Msg("Not scanning, file not part of workspace")
+				go f.ScanFile(bgCtx, filePath)
 			} else {
 				logger.Warn().Msg("Not scanning, auto-scan is disabled")
 			}
+		} else if autoScanEnabled {
```
bastiandoetsch, line 514: why is this dependent on the autoscan boolean?

application/server/server_test.go
```diff
@@ -772,6 +772,57 @@ func Test_textDocumentDidSaveHandler_shouldAcceptDocumentItemAndPublishDiagnosti
...
 
+func createTemporaryDirectoryWithSnykFile(t *testing.T) (snykFilePath string, folderPath string) {
+	t.Helper()
+
+	temp := t.TempDir()
+	temp = filepath.Clean(temp)
+	temp, err := filepath.Abs(temp)
+	if err != nil {
+		t.Fatalf("couldn't get abs folder path of temp dir: %v", err)
+	}
```
bastiandoetsch, line 783: you can use `assert.NoError(t, err)` which is more concise.

application/server/server.go
cat2608: This bit was just a small refactor to reduce one level of nesting. It's the first [commit](https://github.com/snyk/snyk-ls/pull/563/commits/06891fae540bf5799aed5037a00cc932c2cb37ce)

application/server/server.go
cat2608: >I wonder though how we transmit the need to rescan to get new results. Maybe a notification? WDYT?

Could you please elaborate a bit here? Are you saying we need to notify the user that after a local change to the `.snyk` file, the workspace will be re-scanned?

So far, I've been doing it in the background: when `textDocumentDidSaveHandler` the scan occurs.

application/server/server.go
cat2608: Done 🙌

acke (approved):
Looks good to me.

application/server/server.go
```diff
@@ -547,21 +547,28 @@ func textDocumentDidSaveHandler() jrpc2.Handler {
...
 		filePath := uri.PathFromUri(params.TextDocument.URI)
 
-		f := workspace.Get().GetFolderContaining(filePath)
-		autoScanEnabled := config.CurrentConfig().IsAutoScanEnabled()
-		if f != nil && autoScanEnabled {
-			go f.ScanFile(bgCtx, filePath)
-		} else {
+		ws := workspace.Get()
+		if ws != nil && autoScanEnabled && uri.IsDotSnykFile(params.TextDocument.URI) {
+			go ws.ScanWorkspace(bgCtx)
```
bastiandoetsch, line 559: Question: why not scan the workspaceFolder? Scanning the workspace could be too much - especially in Eclipse, corporate users often have 30+ projects (=workspace folders) open, and scanning all of them may slow down their workflow & computer. You could pull line 563 up, and then call f.Scan(). WDYT?

application/server/server.go
```diff
@@ -547,21 +547,28 @@ func textDocumentDidSaveHandler() jrpc2.Handler {
...
+			return nil, nil
+		}
+
+		f := ws.GetFolderContaining(filePath)
+		if f != nil {
 			if autoScanEnabled {
-				logger.Warn().Str("documentURI", filePath).Msg("Not scanning, file not part of workspace")
+				go f.ScanFile(bgCtx, filePath)
```
bastiandoetsch, line 566: This may cause a double scan, right? First the scan above and afterwards the f.ScanFile. I'd think we can combine it :)

application/server/server_test.go
```diff
@@ -799,6 +799,55 @@ func Test_textDocumentDidSaveHandler_shouldAcceptDocumentItemAndPublishDiagnosti
...
+ignore:
+  SNYK-JS-QS-3153490:
+    - '*':
+        reason: Ignore me 30 days
+        expires: 2024-08-26T13:55:05.414Z
+        created: 2024-07-26T13:55:05.417Z
+patch: {}
+`
+	err = os.WriteFile(snykFilePath, []byte(yamlContent), 0600)
+	assert.Nil(t, err)
```
bastiandoetsch, line 823: assert.NoError() would be better :)

bastiandoetsch (approved):
Please have a look at my comment regarding double scans - else very nice 🚀

application/server/server.go
bastiandoetsch: From the customer perspective, if they have deactivated auto-scanning, they wouldn't notice a need to re-scan. On the other hand, they are the ones to update the .snyk file so it may be okay.

bastiandoetsch: From the customer perspective, if they have deactivated auto-scanning, they wouldn't notice a need to re-scan. On the other hand, they are the ones to update the .snyk file so it may be okay.

application/server/server.go
```diff
@@ -111,6 +111,16 @@ func textDocumentDidChangeHandler() jrpc2.Handler {
...
+		f := workspace.Get().GetFolderContaining(filePath)
+
+		if f != nil && autoScanEnabled && uri.IsDotSnykFile(params.TextDocument.URI) {
+			go f.ScanFolder(ctx)
+			return nil, nil
+		}
+
 		for _, change := range params.ContentChanges {
 			if packageScanner, ok := di.Scanner().(snyk.PackageScanner); ok {
 				packageScanner.ScanPackages(ctx, c, uri.PathFromUri(params.TextDocument.URI), change.Text)
```
cat2608, line 126: @bastiandoetsch I also added the trigger when a user runs the command `snyk ignore --id='ISSUE_ID' --reason='REASON'` in the terminal.

I need to commit the unit test for this branch and move this logic to a function so both handlers can use it

bastiandoetsch: Oh yeah, probably when VSCode refreshes and the content changes, it is sending this.

application/server/server_test.go
```diff
@@ -799,6 +800,55 @@ func Test_textDocumentDidSaveHandler_shouldAcceptDocumentItemAndPublishDiagnosti
...
+func Test_textDocumentDidSaveHandler_shouldTriggerScanForDotSnykFile(t *testing.T) {
+	loc, jsonRPCRecorder := setupServer(t)
+	config.CurrentConfig().SetSnykCodeEnabled(false)
+	fakeAuthenticationProvider := di.AuthenticationService().Provider().(*snyk.FakeAuthenticationProvider)
+	fakeAuthenticationProvider.IsAuthenticated = true
+
+	_, err := loc.Client.Call(ctx, "initialize", nil)
+	if err != nil {
+		t.Fatalf("initialization failed: %v", err)
+	}
```
bastiandoetsch, line 837: same here - should use `assert.NoError(t, err)`

application/server/server.go
bastiandoetsch: Let me know, once you've pushed it, I'll re-review and merge.

cat2608: For this case, I'm trying to make this test pass:


```go
func Test_textDocumentDidChangeHandler_shouldTriggerScanForDotSnykFile(t *testing.T) {
	loc, jsonRPCRecorder := setupServer(t)
	config.CurrentConfig().SetSnykCodeEnabled(false)

	fakeAuthenticationProvider := di.AuthenticationService().Provider().(*snyk.FakeAuthenticationProvider)
	fakeAuthenticationProvider.IsAuthenticated = true

	_, err := loc.Client.Call(ctx, "initialize", nil)
	if err != nil {
		t.Fatalf("initialization failed: %v", err)
	}

	snykFilePath, folderPath := createTemporaryDirectoryWithSnykFile(t)

	// Simulate a "did change" event for the .snyk file
	sendFileChangedMessage(t, snykFilePath, folderPath, loc)

	// Wait for $/snyk.scan notification
	assert.Eventually(
		t,
		checkForSnykScan(t, jsonRPCRecorder),
		15*time.Second,
		50*time.Millisecond,
	)
}

func sendFileChangedMessage(t *testing.T, filePath, fileDir string, loc server.Local) sglsp.DocumentURI {
	t.Helper()
	c := config.CurrentConfig()
	didChangeParams := sglsp.DidChangeTextDocumentParams{
		TextDocument: sglsp.VersionedTextDocumentIdentifier{
			TextDocumentIdentifier: sglsp.TextDocumentIdentifier{
				URI: uri.PathToUri(filePath),
			},
			Version: 1,
		},
		ContentChanges: []sglsp.TextDocumentContentChangeEvent{
			{
				Text: "SNYK-JS-QS-3153490", // simulates: `snyk ignore --id='SNYK-JS-QS-3153490'`
			},
		},
	}
	workspace.Get().AddFolder(workspace.NewFolder(c, fileDir,
		"Test",
		di.Scanner(),
		di.HoverService(),
		di.ScanNotifier(),
		di.Notifier()))

	_, err := loc.Client.Call(ctx, "textDocument/didChange", didChangeParams)
	if err != nil {
		t.Fatal(err)
	}

	return didChangeParams.TextDocument.URI
}
```

I'm following a similar approach written for other cases but I wasn't able to solve this error yet:

```
024-06-28T16:07:55+01:00 INF - Fake Authentication - successful.
2024-06-28T16:07:55+01:00 DBG - IsAuthenticated: fake auth successful
2024-06-28T16:07:55+01:00 INF ide.workspace.folder.DelegatingConcurrentScanner.ScanFile  - Scan was canceled err={}
--- FAIL: Test_textDocumentDidChangeHandler_shouldTriggerScanForDotSnykFile (15.21s)
    server_test.go:1033: 
        	Error Trace:	/Users/cata/git/snyk/hammerhead/snyk-ls/application/server/server_test.go:1033
        	Error:      	Condition never satisfied
        	Test:       	Test_textDocumentDidChangeHandler_shouldTriggerScanForDotSnykFile
FAIL
```

which is happening here: `domain/snyk/scanner.go`.

We could go for the `textDocumentDidSaveHandler` only and work on the `textDocumentDidChangeHandler` in another PR. WDYT?

application/server/server_test.go
cat2608: This is the case when there is an error and all the other blocks are using `t.Fatal`. In this case I'm only appending the error message.

application/server/server.go
cat2608: @bastiandoetsch here is the initial change needed in IntelliJ https://github.com/snyk/snyk-intellij-plugin/pull/562

PR #564, opened 2024-06-26
feat: refactor the HTML to use html/template and to inject styling for other IDEs [IDE-326]

teodora-sandu:
### Description

This PR refactors the HTML template used for the OSS Sugestion Panel so that it's more secure and uses the `html/template` library. This also refactors the template so that it injects icons directly in LS and so other IDEs don't need to know what icons to inject. It adds a `${ideStyle}` variable so that in both VSCode and IntelliJ we can inject the IDE specific styling, like we do for Snyk Code.

To test this, need to use https://github.com/snyk/vscode-extension/pull/482.

### Checklist

- [x] Linted

### DIFF
--- a/infrastructure/code/code_html.go
+++ b/infrastructure/code/code_html.go
@@ -27,10 +27,9 @@ import (
 	"strings"
 	"time"
 
-	"github.com/gomarkdown/markdown"
-
 	"github.com/snyk/snyk-ls/application/config"
 	"github.com/snyk/snyk-ls/domain/snyk"
+	"github.com/snyk/snyk-ls/internal/html"
 	"github.com/snyk/snyk-ls/internal/product"
 )
 
@@ -66,8 +65,8 @@ var globalTemplate *template.Template
 func init() {
 	funcMap := template.FuncMap{
 		"repoName":      getRepoName,
-		"trimCWEPrefix": TrimCWEPrefix,
-		"idxMinusOne":   IdxMinusOne,
+		"trimCWEPrefix": html.TrimCWEPrefix,
+		"idxMinusOne":   html.IdxMinusOne,
 	}
 
 	var err error
@@ -92,9 +91,9 @@ func getCodeDetailsHtml(issue snyk.Issue) string {
 		"IssueTitle":         additionalData.Title,
 		"IssueMessage":       additionalData.Message,
 		"IssueType":          getIssueType(additionalData),
-		"SeverityIcon":       GetSeverityIconSvg(issue),
+		"SeverityIcon":       html.GetSeverityIconSvg(issue),
 		"CWEs":               issue.CWEs,
-		"IssueOverview":      markdownToHTML(additionalData.Text),
+		"IssueOverview":      html.MarkdownToHTML(additionalData.Text),
 		"IsIgnored":          issue.IsIgnored,
 		"DataFlow":           additionalData.DataFlow,
 		"DataFlowTable":      prepareDataFlowTable(additionalData),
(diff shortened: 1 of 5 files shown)


infrastructure/code/code_html.go
```diff
@@ -279,33 +265,6 @@ func getExternalIconSvg() template.HTML {
...
-		return template.HTML(`<svg fill="none" xmlns="http://www.w3.org/2000/svg" width="32" height="32" viewBox="0 0 16 16">
-			 <rect width="16" height="16" rx="2" fill="#88879E"/>
-			 <path d="M11 11v2H6.705v-2H11ZM7 3v10H5V3h2Z" fill="#fff"/>
-		 </svg>`)
-	default:
-		return ``
-	}
-}
-
 func getGitHubIconSvg() template.HTML {
```
bastiandoetsch, line 268: Wondering, if caching/memory will become an issue with 1000s of issues if we embed the svg like that. Do you know how Chrome handles this?

infrastructure/oss/issue_html.go
```diff
@@ -17,104 +17,118 @@
...
+		"IssueName":          additionalData.Name,
+		"IssueTitle":         additionalData.Title,
+		"IssueType":          getIssueType(additionalData),
+		"SeverityText":       issue.Severity.String(),
+		"SeverityIcon":       html.GetSeverityIconSvg(issue),
+		"VulnerableModule":   additionalData.Name,
+		"IssueOverview":      html.MarkdownToHTML(string(overview)),
+		"CVEs":               additionalData.Identifiers.CVE,
+		"CWEs":               additionalData.Identifiers.CWE,
+		"CvssScore":          fmt.Sprintf("%.1f", additionalData.CvssScore),
```
bastiandoetsch, line 76: As we're at it - should we just update it to CVSSv4? ;)

infrastructure/code/code_html.go
teodora-sandu: Do you mean we might run into trouble ebecause we would be using a lot of memory if we cache the same SVG multiple times?

application/server/notification/scan_notifier.go
```diff
@@ -135,7 +135,7 @@ func (n *scanNotifier) appendOssIssues(scanIssues []lsp.ScanIssue, issues []snyk
 				Version:           matchingIssue.Version,
 				Exploit:           matchingIssue.Exploit,
 				CVSSv3:            matchingIssue.CVSSv3,
-				CvssScore:         strconv.FormatFloat(matchingIssue.CvssScore, 'f', 2, 64), // convert float64 to string with 2 decimal places
+				CvssScore:         strconv.FormatFloat(matchingIssue.CVSSv4, 'f', 2, 64), // convert float64 to string with 2 decimal places
```
bastiandoetsch, line 138: Nice! I think we need to check if it's filled, as it's not always available.

teodora-sandu:
https://github.com/snyk/vscode-extension/pull/482 should be merged first before this.

infrastructure/oss/issue_html.go
```diff
@@ -17,145 +17,171 @@
...
 
 func getExploitMaturity(issue snyk.OssIssueData) string {
 	if len(issue.Exploit) > 0 {
-		return fmt.Sprintf("<div class='summary-item maturity'><div class='label font-light'>Exploit maturity</div>"+
-			"<div class='content'>%s</div></div>", issue.Exploit)
+		return issue.Exploit
 	} else {
 		return ""
 	}
 }
```
cat2608, line 108: Maybe we can remove the else?

```go
func getExploitMaturity(issue snyk.OssIssueData) string {
	if len(issue.Exploit) > 0 {
		return issue.Exploit
	}
	return ""
}
```

cat2608 (approved):

infrastructure/code/code_html.go
cat2608: I've been thinking about this and at the very beginning of this migration, when we were just supporting IntelliJ and writing the HTML and CSS from scratch, we embedded everything in one single HTML file, trading DRY for velocity.

Now that the project has grown and VSCode and IntelliJ are loading CSS rules anyway, I think we could start considering removing the CSS and SVGs from the Language Server and moving them to the IDEs as separate assets. 

### Benefits:
1. **Reduced Redundancy**: Each vulnerability doesn't need to embed the same CSS and SVGs.
2. **Improved Performance**: Smaller HTML payloads and better caching of static assets.

WDYT @teodora-sandu @bastiandoetsch

PR #571, opened 2024-07-08
fix: introduce throughs and detailed paths in html of oss

teodora-sandu:
### Description

Fixes some small UI problems in the OSS suggestion panel:
- the `Introduced Through` overview section should have the packages separated by comma and not as separate components on individual lines
- the `Detailed Paths` section in IntelliJ needs to render only the top 3 elements. It doesn't seem like VSCode deals with that today but it would be a good thing to have.

Have tested it with `target-service`.

### Checklist

- [x] Linted

### Screenshots / GIFs
Before:
Screenshot 2024-07-08 at 15 59 49

After:
Screenshot 2024-07-08 at 15 45 58

### DIFF
--- a/infrastructure/oss/issue_html.go
+++ b/infrastructure/oss/issue_html.go
@@ -62,6 +62,8 @@ func getDetailsHtml(issue snyk.Issue) string {
 	}
 	overview := markdown.ToHTML([]byte(additionalData.Description), nil, nil)
 
+	detailedPaths := getDetailedPaths(additionalData)
+
 	data := map[string]interface{}{
 		"IssueId":            issue.ID,
 		"IssueName":          additionalData.Name,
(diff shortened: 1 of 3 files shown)


infrastructure/oss/issue_html.go
```diff
@@ -114,7 +117,7 @@ type IntroducedThrough struct {
 }
 
 func getIntroducedThroughs(issue snyk.OssIssueData) []IntroducedThrough {
-	introducedThroughs := []IntroducedThrough{}
+	var introducedThroughs []IntroducedThrough
```
bastiandoetsch, line 120: nitpick: shouldn't it be `introducedThrough`?

bastiandoetsch (approved):

infrastructure/oss/issue_html.go
teodora-sandu: It's mean to be a list of `IntroducedThrough`  so that's why it's the extra `s`

PR #582, opened 2024-07-11
fix: only show one code lens and quick fix code action in OSS files [IDE-484][IDE-485]

bastiandoetsch:
### Description

When an OSS package can be upgraded, only one code action & only one code lens should be shown. This PR ensures it.

### Checklist

- [x] Linted

### Screenshots / GIFs
Unreasonable updates before:
image

After:
image

Before:
image

After:
image


application/config/automatic_config.go
```diff
@@ -61,7 +62,12 @@ func (c *Config) normalizePath(foundPath string) (string, bool) {
 	return path, false
 }
 
-func (c *Config) determineMavenHome() {
+func (c *Config) mavenDefaults() {
+	// explicitly and always use headless mode
+	mavenOptsVarName := "MAVEN_OPTS"
+	mavenOpts := fmt.Sprintf("%s %s", os.Getenv(mavenOptsVarName), "-Djava.awt.headless=true")
+	_ = os.Setenv(mavenOptsVarName, mavenOpts)
+
```
bastiandoetsch, line 70: This is unrelated to the actual PR goal, but the JVM was popping up UI and stealing focus - we don't want that in LS.

application/config/config.go
```diff
@@ -248,11 +248,13 @@ func New() *Config {
 
 func initWorkFlowEngine(c *Config) {
 	conf := configuration.NewInMemory()
+	conf.Set(cli_constants.EXECUTION_MODE_KEY, cli_constants.EXECUTION_MODE_VALUE_STANDALONE)
+	enableOAuth := c.authenticationMethod == types.OAuthAuthentication
+	conf.Set(configuration.FF_OAUTH_AUTH_FLOW_ENABLED, enableOAuth)
+
```
bastiandoetsch, line 254: just moved configuration before engine init

application/server/codeaction_handlers.go
bastiandoetsch, line 1: This file was moved, as it tries to inject a code action service in `getCodeActionHandler` before it's actually initialized via di.Init().

application/server/server.go
```diff
@@ -142,32 +141,25 @@ func codeLensHandler() jrpc2.Handler {
...
+		defer c.Logger().Info().Str("method", "CodeLensHandler").
+			Bool("isDirtyFile", isDirtyFile).
+			Int("lensCount", len(lenses)).
+			Msg("SENDING")
+
 		if !isDirtyFile {
 			return lenses, nil
 		}
-
-		return filterCodeFixCodelens(lenses), nil
```
bastiandoetsch, line 155: I removed the return of "non-fix" code lenses - it doesn't make sense to return them either, limiting it to code is not sensible. Also, we don't show any other code lenses than fix code lenses right now, so it should not have visible effects for now.

application/server/server_smoke_test.go
```diff
@@ -523,9 +619,12 @@ func setupRepoAndInitialize(t *testing.T, repo string, commit string, loc server
 	return cloneTargetDir
 }
 
-func checkFeatureFlagStatus(t *testing.T, loc *server.Local, c *config.Config) {
+func checkFeatureFlagStatus(t *testing.T, c *config.Config, loc *server.Local) {
```
bastiandoetsch, line 622: when we pass a config, it should be the first param (or second if it's a test helper)

internal/types/grouping.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+type GroupingFunction func(groupables []Groupable) any
+type GroupingType string
+
+const Quickfix GroupingType = "quickfix-grouping"
+
+func MaxSemver() GroupingFunction {
+	return func(groupables []Groupable) any {
+		if len(groupables) == 0 {
```
ShawkyZ, line 40: a test would be nice

internal/types/command.go
```diff
@@ -67,7 +67,22 @@ type CommandData struct {
 	 * Arguments that the command handler should be
 	 * invoked with.
 	 */
-	Arguments []any
+	Arguments     []any
+	GroupingKey   Key
```
ShawkyZ, line 71: how are the Grouping properties related to CommandData? 
Aren't they only necessary for codeaction? or are there other use cases where we need them ?

infrastructure/oss/issue.go
```diff
@@ -60,6 +60,9 @@ func toIssue(
 					affectedFilePath,
 					issueRange,
 				},
+				GroupingKey:   codeAction.GroupingKey,
```
ShawkyZ, line 63: hmm that's why...
Maybe we can then have "AdditionalProperties" or smth similar in CommandData for these cases?

infrastructure/oss/code_actions.go
```diff
@@ -0,0 +1,162 @@
...
+	"golang.org/x/mod/semver"
+
+	"github.com/snyk/snyk-ls/application/config"
+	"github.com/snyk/snyk-ls/domain/snyk"
+	"github.com/snyk/snyk-ls/infrastructure/learn"
+	"github.com/snyk/snyk-ls/internal/observability/error_reporting"
+	"github.com/snyk/snyk-ls/internal/types"
+)
+
+func (i *ossIssue) AddCodeActions(learnService learn.Service, ep error_reporting.ErrorReporter, affectedFilePath string, issueRange snyk.Range) (actions []snyk.CodeAction) {
```
ShawkyZ, line 33: nice refactoring. so much easier to follow/understand now

PR #593, opened 2024-07-18
fix: parse gradle multiprojects correctly [IDA-39]

bastiandoetsch:
### Description

Gradle multi module projects are reported differently in the JSON output. This PR adjusts parsing and issue key generation.

### Checklist

- [ ] Linted

### Screenshots / GIFs

image

### DIFF
--- a/infrastructure/oss/cli_scanner.go
+++ b/infrastructure/oss/cli_scanner.go
@@ -263,10 +263,15 @@ func (cliScanner *CLIScanner) unmarshallAndRetrieveAnalysis(ctx context.Context,
 	}
 
 	for _, scanResult := range scanResults {
-		targetFilePath := path
 		targetFile := cliScanner.determineTargetFile(scanResult.DisplayTargetFile)
-		if targetFile != "" {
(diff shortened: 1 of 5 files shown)


infrastructure/oss/cli_scanner.go
```diff
@@ -263,10 +263,15 @@ func (cliScanner *CLIScanner) unmarshallAndRetrieveAnalysis(ctx context.Context,
...
 	for _, scanResult := range scanResults {
-		targetFilePath := path
 		targetFile := cliScanner.determineTargetFile(scanResult.DisplayTargetFile)
-		if targetFile != "" {
+		targetFilePath := scanResult.Path
+		if targetFilePath == "" {
+			targetFilePath = workDir
+		}
+		if targetFilePath == "" {
 			targetFilePath = filepath.Join(workDir, targetFile)
```
ShawkyZ, line 272: if targetFilePath is still empty here that means workDir is also empty. So no need to do filepath.Join?

ShawkyZ (approved):

PR #595, opened 2024-07-19
chore: improve error handling and tests

ShawkyZ:
### Description

- Only log err and send a notification to the IDE if base scan fails and don't send it back.
- Removed afero dependency.
- Rewrote vcs tests to not use mocking.
- Fixed an issue with deleting existing cached snapshot when adding a new one in git persistence.
- Changed cache dir perm from 600 to 700 (x perm on dir allows navigating into it).

### Checklist

- [x] Linted

### Screenshots / GIFs


infrastructure/code/code.go
```diff
@@ -189,9 +191,10 @@ func (sc *Scanner) Scan(ctx context.Context, path string, folderPath string) (is
 	results, err := internalScan(ctx, sc, folderPath, logger, filesToBeScanned)
 
 	if err == nil && c.IsDeltaFindingsEnabled() {
-		err = scanAndPersistBaseBranch(ctx, sc, folderPath)
-		if err != nil {
-			logger.Error().Err(err).Msg("couldn't scan base branch for folder " + folderPath)
+		baseScanErr := scanAndPersistBaseBranch(ctx, sc, folderPath)
+		if baseScanErr != nil {
+			go sc.notifier.SendShowMessage(sglsp.MTError, "Couldn't determine the diff between current and base branch. Falling back to showing full scan results.")
```
bastiandoetsch, line 196: You don't really need a goroutine here, but doesn't hurt either. Usually, it's only necessary to use a goroutine, when we're waiting for an answer, e.g. with `ShowMessageRequest`

bastiandoetsch, line 196: `the diff` -> `the difference`

infrastructure/code/code.go
```diff
@@ -298,6 +301,19 @@ func scanAndPersistBaseBranch(ctx context.Context, sc *Scanner, folderPath strin
 	return nil
 }
 
+func normalizeBranchName(branchName string) string {
+	normalized := strings.TrimSpace(branchName)
+	normalized = strings.ToLower(normalized)
+	normalized = strings.ReplaceAll(normalized, " ", "_")
+	reg, err := regexp.Compile(`[^a-z0-9_\-]+`)
+	if err != nil {
+		return ""
```
bastiandoetsch, line 310: In this case, probably a hash would serve? Sth like sha256 of branchname?

bastiandoetsch (approved):

infrastructure/code/code.go
ShawkyZ: I was thinking that it would be better to have the branch name in folder name. So that we could possibly recognize it later.

ShawkyZ: actually it should always compile. 
Moved it to a global var. 
The ReplaceString will replace everything matched. It should match all chars not a-Z0-9\-

PR #599, opened 2024-07-23
fix: check if path is a sub directory correctly

bastiandoetsch:
### Description

Previously, the sub directory check for trust was just checking if the trusted directory had the same prefix. Now it's checking the relative path.

Example:
Trusted: /a
Untrusted: /a-123

Previously, /a-123 would be trusted. This PR fixes that.

### Checklist

- [x] Linted

### Screenshots / GIFs

### DIFF
--- a/internal/testutil/test_setup.go
+++ b/internal/testutil/test_setup.go
@@ -24,6 +24,7 @@ import (
 	"testing"
 
 	"github.com/rs/zerolog"
+	"github.com/stretchr/testify/assert"
 
 	"github.com/snyk/snyk-ls/application/config"
 	"github.com/snyk/snyk-ls/internal/progress"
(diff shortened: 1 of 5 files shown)


internal/testutil/test_setup.go
```diff
@@ -144,3 +150,34 @@ func OnlyEnableCode() {
...
+
+func SetupCustomTestRepo(t *testing.T, rootDir string, url string, targetCommit string, logger *zerolog.Logger) (string, error) {
+	t.Helper()
+	tempDir := filepath.Join(rootDir, util.Murmur(t.Name()))
+	assert.NoError(t, os.MkdirAll(tempDir, 0755))
+	repoDir := "1"
+	absoluteCloneRepoDir := filepath.Join(tempDir, repoDir)
+	cmd := []string{"clone", url, repoDir}
+	logger.Debug().Interface("cmd", cmd).Msg("clone command")
+	clone := exec.Command("git", cmd...)
```
ShawkyZ, line 162: should we use the repo setup logic from go-git here? 
Sample: git_cloner.go -> initGitRepo

ShawkyZ (approved):

internal/testutil/test_setup.go
ShawkyZ: Actually maybe it doesn't make much sense for this case. We can revisit it later

bastiandoetsch: Yeah, we can consolidate that.

PR #603, opened 2024-07-24
chore: tweak parallelization of CLI, add test

bastiandoetsch:
### Description

This sets the max parallel execution of CLI to the number of cores - 2 or 1, whichever is higher. It also fixes a concurrency issue when retrieving open source issues and storing them in the scanner-local caches by using the scanner mutex.

A smoke test was added to check parallelization.

### Checklist

- [x] Linted

### DIFF
--- a/application/server/constants.go
+++ b/application/server/constants.go
@@ -1,0 +1,19 @@
+/*
+ * © 2024 Snyk Limited
+ *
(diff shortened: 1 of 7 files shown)


ShawkyZ (approved):
🚀🍔🌟

PR #611, opened 2024-07-31
fix: parsing oss output [IDE-517]

cat2608:
### Description

- parse oss results correctly for different package managers

### Checklist

- [x] Linted


bastiandoetsch (approved):

ShawkyZ (approved):
🚀🌕

PR #613, opened 2024-08-01
feat: added IsNew to lsp.Issue model [IDE-537]

ShawkyZ:
### Description

### Checklist

- [ ] Linted

### DIFF
--- a/internal/types/lsp.go
+++ b/internal/types/lsp.go
@@ -1064,6 +1064,7 @@ type ScanIssue struct { // TODO - convert this to a generic type
 	FilePath       string        `json:"filePath"`
 	Range          sglsp.Range   `json:"range"`
 	IsIgnored      bool          `json:"isIgnored"`
+	IsNew          bool          `json:"isNew"`
 	IgnoreDetails  IgnoreDetails `json:"ignoreDetails"`
 	AdditionalData any           `json:"additionalData,omitempty"`
 }
--- a/application/server/notification/scan_notifier.go
+++ b/application/server/notification/scan_notifier.go
@@ -283,6 +283,7 @@ func (n *scanNotifier) appendCodeIssues(scanIssues []types.ScanIssue, issues []s
 			FilePath:  issue.AffectedFilePath,
 			Range:     converter.ToRange(issue.Range),
 			IsIgnored: issue.IsIgnored,
+			IsNew:     issue.IsNew,
 			AdditionalData: types.CodeIssueData{
 				Message:            additionalData.Message,
 				Rule:               additionalData.Rule,
--- a/domain/snyk/issues.go
+++ b/domain/snyk/issues.go
@@ -40,7 +40,7 @@ type Issue struct {
 	Severity  Severity
 	IssueType Type
 	IsIgnored bool // If not explicitly it will default to false, so it doesn't break backwards
-	isNew     bool
+	IsNew     bool
 	// compatibility
 	IgnoreDetails *IgnoreDetails // It defaults to nil, so it doesn't break backwards compatibility
 	// Range identifies the location of this issue in its source of origin (e.g. line & character start & end)
@@ -96,12 +96,12 @@ func (i *Issue) EndColumn() int {
 	return i.Range.End.Character
 }
 
-func (i *Issue) IsNew() bool {
-	return i.isNew
+func (i *Issue) GetIsNew() bool {
+	return i.IsNew
 }
 
 func (i *Issue) SetIsNew(isNew bool) {
-	i.isNew = isNew
+	i.IsNew = isNew
 }
 
 func (i *Issue) GetGlobalIdentity() string {
--- a/internal/delta/delta_finder_test.go
+++ b/internal/delta/delta_finder_test.go
@@ -68,7 +68,7 @@ func TestFind_DifferWithEnricher(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Len(t, enrichedList, 2)
 
-	assert.True(t, enrichedList[1].IsNew())
+	assert.True(t, enrichedList[1].GetIsNew())
 }
 
 func TestFind_DifferWithEnricherWithMatcher(t *testing.T) {
@@ -118,7 +118,7 @@ func TestFind_DifferWithEnricherWithMatcher(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Len(t, enrichedList, 3)
 
-	assert.True(t, enrichedList[2].IsNew())
+	assert.True(t, enrichedList[2].GetIsNew())
 	assert.Equal(t, enrichedList[2].GetGlobalIdentity(), currentIssueList[2].GetGlobalIdentity())
 }
 
@@ -156,7 +156,7 @@ func TestFind_DifferWithEnricherWithMatcher_NoNewIssues(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Len(t, enrichedList, 2)
 	for _, enriched := range enrichedList {
-		assert.False(t, enriched.IsNew())
+		assert.False(t, enriched.GetIsNew())
 		assert.NotEmpty(t, enriched.GetGlobalIdentity())
 	}
 }
--- a/internal/delta/findings_identifiable.go
+++ b/internal/delta/findings_identifiable.go
@@ -39,5 +39,5 @@ type Identifiable interface {
 	GetGlobalIdentity() string
 	SetGlobalIdentity(globalIdentity string)
 	SetIsNew(isNew bool)
-	IsNew() bool
+	GetIsNew() bool
 }
--- a/internal/delta/mock_identifiable.go
+++ b/internal/delta/mock_identifiable.go
@@ -65,7 +65,7 @@ func (m *mockIdentifiable) SetIsNew(isNew bool) {
 	m.isNew = isNew
 }
 
-func (m *mockIdentifiable) IsNew() bool {
+func (m *mockIdentifiable) GetIsNew() bool {
 	return m.isNew
 }
 


domain/ide/converter/converter.go
```diff
@@ -65,7 +67,7 @@ func ToCodeAction(issue snyk.Issue, action snyk.CodeAction) types.CodeAction {
 	return types.CodeAction{
 		Title:       action.Title,
 		Kind:        types.QuickFix,
-		Diagnostics: ToDiagnostics([]snyk.Issue{issue}),
+		Diagnostics: ToDiagnostics([]snyk.Issue{issue}, product.ProductUnknown),
```
ShawkyZ, line 70: This doesn't seem correct

domain/ide/converter/converter.go
```diff
@@ -154,30 +156,243 @@ func ToPosition(p snyk.Position) sglsp.Position {
...
+		diagnostic := types.Diagnostic{
 			Range:           ToRange(issue.Range),
 			Severity:        ToSeverity(issue.Severity),
 			Code:            issue.ID,
 			Source:          string(issue.Product),
 			Message:         issue.Message,
 			CodeDescription: types.CodeDescription{Href: types.Uri(s)},
-		})
+		}
+		// TODO: This is not future proof. Find a better way to match.
```
ShawkyZ, line 190: the appendIaCIssues/Code/Oss functions should now be changed to enrich 1 issue instead of receiving a list. This will fix this TODO.

domain/ide/workspace/folder.go
```diff
@@ -262,7 +262,7 @@ func NewFolder(
 
 func (f *Folder) sendEmptyDiagnosticForFile(path string) {
 	config.CurrentConfig().Logger().Debug().Str("filePath", path).Msg("sending empty diagnostic for file")
-	f.sendDiagnosticsForFile(path, []snyk.Issue{})
+	f.sendDiagnosticsForFile(product.ProductUnknown, path, []snyk.Issue{})
```
ShawkyZ, line 265: this is wrong. The caller has the issue Type and it should be passed down to sendEmptyDiagnosticForFile

application/server/server.go
```diff
@@ -522,7 +523,7 @@ func textDocumentDidOpenHandler() jrpc2.Handler {
 			logger.Info().Msg("Sending cached issues")
 			diagnosticParams := types.PublishDiagnosticsParams{
 				URI:         params.TextDocument.URI,
-				Diagnostics: converter.ToDiagnostics(filteredIssues[filePath]),
+				Diagnostics: converter.ToDiagnostics(filteredIssues[filePath], product.ProductUnknown),
```
ShawkyZ, line 526: we need to figure out the product here.

.goreleaser.yaml
```diff
@@ -50,4 +50,4 @@ dist: build
 env:
   - GO111MODULE=on
   - CGO_ENABLED=0
-  - LS_PROTOCOL_VERSION=13
+  - LS_PROTOCOL_VERSION=14
```
bastiandoetsch, line 53: do we tickets to update Eclipse, IntelliJ & VSCode to that version already?

bastiandoetsch: Also, how do we handle protocol version 13 (which hasn't been released in stable either)

bastiandoetsch (commented):
I love that this removed so much code/bloat.

application/server/server_smoke_test.go
```diff
@@ -605,23 +611,67 @@ func checkForScanParams(t *testing.T, jsonRPCRecorder *testutil.JsonRPCRecorder,
...
+	notifications := jsonRPCRecorder.FindNotificationsByMethod("textDocument/publishDiagnostics")
+	for _, n := range notifications {
+		diagnosticsParams := types.PublishDiagnosticsParams{}
+		_ = n.UnmarshalParams(&diagnosticsParams)
+		for _, diagnostic := range diagnosticsParams.Diagnostics {
+			diagnosticCode, ok := diagnostic.Code.(string)
+			if ok && diagnosticCode == "Snyk Error" {
+				continue
+			}
+			if diagnostic.Source != string(p) || filepath.Dir(uri.PathFromUri(diagnosticsParams.URI)) != folderPath {
```
bastiandoetsch, line 629: I would think about using `uri.FolderContains(diagnosticsParams.URI)` - as that also covers sub-directories.

application/server/server_smoke_test.go
```diff
@@ -605,23 +611,67 @@ func checkForScanParams(t *testing.T, jsonRPCRecorder *testutil.JsonRPCRecorder,
...
+				continue
+			}
+
+			var issue types.ScanIssue
+			err := mapSerializedIssueToStruct(diagnostic.Data, &issue)
+
+			if err != nil {
+				continue
```
bastiandoetsch, line 637: Why don't we fail the test here? Shouldn't we be able to map all returned issues?

application/server/server_smoke_test.go
```diff
@@ -605,23 +611,67 @@ func checkForScanParams(t *testing.T, jsonRPCRecorder *testutil.JsonRPCRecorder,
...
+
+	var issueList []types.ScanIssue
+	notifications := jsonRPCRecorder.FindNotificationsByMethod("textDocument/publishDiagnostics")
+	for _, n := range notifications {
+		diagnosticsParams := types.PublishDiagnosticsParams{}
+		_ = n.UnmarshalParams(&diagnosticsParams)
+		for _, diagnostic := range diagnosticsParams.Diagnostics {
+			diagnosticCode, ok := diagnostic.Code.(string)
+			if ok && diagnosticCode == "Snyk Error" {
+				continue
```
bastiandoetsch, line 627: do we expect errors here? why don't they fail the test?

application/server/server_smoke_test.go
```diff
@@ -605,23 +611,67 @@ func checkForScanParams(t *testing.T, jsonRPCRecorder *testutil.JsonRPCRecorder,
...
+			}
+			issueList = append(issueList, issue)
+		}
+	}
+
+	return issueList
+}
+
+func mapSerializedIssueToStruct(m any, result *types.ScanIssue) error {
```
bastiandoetsch, line 646: ah, data is a map[string]string as it's LSPAny type?

bastiandoetsch: Or even define the data attribute directly as types.ScanIssues to avoid this serialization dance.

application/server/server_smoke_test.go
```diff
@@ -605,23 +611,67 @@ func checkForScanParams(t *testing.T, jsonRPCRecorder *testutil.JsonRPCRecorder,
...
+	assert.Greater(t, len(issueList), 0)
+	for _, issue := range issueList {
 		codeIssueData, ok := issue.AdditionalData.(map[string]interface{})
 		if !ok || codeIssueData["hasAIFix"] == false || codeIssueData["rule"] != "WebCookieSecureDisabledByDefault" {
 			continue
 		}
 		call, err := loc.Client.Call(ctx, "workspace/executeCommand", sglsp.ExecuteCommandParams{
 			Command:   types.CodeFixDiffsCommand,
-			Arguments: []any{uri.PathToUri(snykCodeScanParams.FolderPath), uri.PathToUri(issue.FilePath), issue.Id},
+			Arguments: []any{path.Base(issue.FilePath), uri.PathToUri(issue.FilePath), issue.Id},
```
bastiandoetsch, line 674: This doesn't look correct. It should be folderPath, filePath. I'd pass in the folder or folderPath from the caller.

domain/ide/converter/converter.go
```diff
@@ -159,25 +161,225 @@ func ToDiagnostics(issues []snyk.Issue) []types.Diagnostic {
 	// Sending null instead of an empty array leads to stored diagnostics not being cleared.
 	// Do not prefer nil over an empty slice in this case. The next line ensures that even if issues is empty,
 	// the return value of this function will not be null.
```
bastiandoetsch, line 163: Well, this comment didn't age well ;)

domain/ide/converter/converter.go
```diff
@@ -159,25 +161,225 @@ func ToDiagnostics(issues []snyk.Issue) []types.Diagnostic {
...
-	diagnostics := []types.Diagnostic{}
 
-	for _, issue := range issues {
+	var diagnostics []types.Diagnostic
```
bastiandoetsch, line 165: Wondering: Is this empty array or nil?

bastiandoetsch: According to [this](https://www.perplexity.ai/search?q=pending&newFrontendContextUUID=bb514303-e101-4531-a6d5-99fd11f32b2a), it should stay like it was.

domain/ide/hover/service.go
```diff
@@ -36,16 +36,16 @@ type DefaultHoverService struct {
 	hovers       map[string][]Hover[Context]
 	hoverIndexes map[string]bool
 	hoverChan    chan DocumentHovers
-	mutex        *sync.Mutex
+	mutex        *sync.RWMutex
```
bastiandoetsch, line 39: nice small performance enhancement!

internal/types/lsp.go
```diff
@@ -1064,6 +1062,7 @@ type ScanIssue struct { // TODO - convert this to a generic type
 	FilePath       string        `json:"filePath"`
 	Range          sglsp.Range   `json:"range"`
 	IsIgnored      bool          `json:"isIgnored"`
+	IsNew          bool          `json:"isNew"`
```
bastiandoetsch, line 1065: I guess, we are adding this in VSCode and IntelliJ for proto version 14, too?

bastiandoetsch (changes requested):
Some fixes needed :). Let me know when I should have another look. Good job 🎉 !

.goreleaser.yaml
ShawkyZ: yeah we will need to skip in both of them

domain/ide/converter/converter.go
ShawkyZ: it's nil. but append can append to nil.

according to this it doesn't make a difference
https://stackoverflow.com/a/29164565

.goreleaser.yaml
ShawkyZ: we already PRs to bump to 13 for intellij and eclipse. we can adapt these prs

application/server/server_smoke_test.go
ShawkyZ: Agree, it doesn't make any sense lol

application/server/server_smoke_test.go
ShawkyZ: I think it should be ok. We will only need that when we deserialize the the types.*Params. 
Which we don't do except in tests afaik

ShawkyZ: I like that type.go maps the protocol def 1:1. Makes it easier to see and compare

application/server/server_smoke_test.go
ShawkyZ: yup that makes sense

internal/types/lsp.go
ShawkyZ: yeah we should. I'm thinking if we must do it now tho as soon as we release 14 or is it ok to add it whenever we need it

ShawkyZ: We will need it soon anyway when we implement the tabs

bastiandoetsch (approved):

PR #615, opened 2024-08-05
feat: add sending of issue overview tree [IDE-291]

bastiandoetsch:
### Description

image

### Checklist

- [x] Linted


domain/ide/workspace/ui/diagnostics_overview.go
```diff
@@ -0,0 +1,204 @@
...
+
+	"github.com/snyk/snyk-ls/application/config"
+	"github.com/snyk/snyk-ls/domain/snyk"
+	"github.com/snyk/snyk-ls/internal/html"
+	"github.com/snyk/snyk-ls/internal/notification"
+	"github.com/snyk/snyk-ls/internal/product"
+	"github.com/snyk/snyk-ls/internal/types"
+)
+
+//go:embed template/diagnosticsOverview.html
```
ShawkyZ, line 35: TIL

domain/ide/workspace/ui/diagnostics_overview.go
```diff
@@ -0,0 +1,204 @@
...
+// Node represents a tree node
+type Node struct {
+	Icon           template.HTML
+	Text           template.HTML
+	ProductEnabled bool
+}
+
+func SendDiagnosticsOverview(c *config.Config, p product.Product, issuesByFile snyk.IssuesByFile, notifier notification.Notifier) {
+	logger := c.Logger().With().Str("method", "ui.SendDiagnosticsOverview").Logger()
+	if p != "" {
```
ShawkyZ, line 72: nitpick: early exit

domain/ide/workspace/ui/diagnostics_overview.go
```diff
@@ -0,0 +1,204 @@
...
+			pluralSuffix,
+			issuesByFile.SeverityCountsAsString(critical, high, medium, low),
+		)
+	}
+	return rootNodeTitle
+}
+
+func isProductEnabled(c *config.Config, p product.Product) bool {
+	switch p {
+	case product.ProductOpenSource:
```
ShawkyZ, line 195: should we move this to config?

domain/ide/workspace/ui/diagnostics_overview.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+func getFileTypeIcon() template.HTML {
+	return ""
+}
+
+// generateSecurityNonce generates a cryptographically secure random nonce.
+// A nonce is used in the Content Security Policy (CSP) to allow specific
+// inline styles and scripts, helping to prevent Cross-Site Scripting (XSS) attacks.
+func generateSecurityNonce() (string, error) {
```
ShawkyZ, line 132: If I remember we were also calculating nonces for the other HTML views correct?
If this was done in LS maybe it makes sense to move this somewhere else

domain/ide/workspace/ui/diagnostics_overview.go
```diff
@@ -0,0 +1,204 @@
...
+
+//go:embed template/diagnosticsOverview.html
+var diagnosticsOverviewTemplatePath string
+
+//go:embed template/diagnosticsOverview.css
+var diagnosticsOverviewTemplateCSS string
+var diagnosticsOverviewTemplate *template.Template
+
+func init() {
+	logger := config.CurrentConfig().Logger().With().Str("method", "ui.init").Logger()
```
ShawkyZ, line 43: inject the config instead of singleton?

domain/ide/workspace/folder.go
```diff
@@ -610,6 +611,7 @@ func isVisibleSeverity(issue snyk.Issue) bool {
 func (f *Folder) publishDiagnostics(product product.Product, issuesByFile snyk.IssuesByFile) {
 	f.sendHovers(issuesByFile)
 	f.sendDiagnostics(issuesByFile)
+	ui.SendDiagnosticsOverview(f.c, product, issuesByFile, f.notifier)
 	f.sendSuccess(product)
```
ShawkyZ, line 615: I think we need to rethink the flow here

Currently if scan success -> load issues from diagnostics.

In the future:
it doesn't matter if success or not IDE reacts to the new notification to load the HTML panel in the IDE

Maybe we don't need the f.sendSuccess anymore?

ShawkyZ:
I love this change btw. I'm so happy that we will delete lots of stuff from the IDEs after this

domain/ide/workspace/folder.go
bastiandoetsch: Actually, yes and no. We will send this by scan & product. So actually it could be combined. Although I like that scan status reporting is separated from the actual overview html. I'm open for both, though.

domain/ide/workspace/ui/diagnostics_overview.go
bastiandoetsch: Can't change the signature of `init()`, else I'd be with you :)

bastiandoetsch: We could move the template parsing into a singleton though, if it's important.

domain/ide/workspace/ui/diagnostics_overview.go
bastiandoetsch: Yeah, this is a real lazy repeat. I'll DRY it up.

bastiandoetsch: Actually, I didn't find repetition, but it's gonna be good to have it in the `html` package anyway. I think, the IaC PR can re-use it.

domain/ide/workspace/ui/diagnostics_overview.go
bastiandoetsch: Yeah, I was originally thinking of having it in product, but then it would have cyclical dependencies. But `config` sounds good.

bastiandoetsch: Actually, we have that code twice - just realized it's already in scanNotifier - so one more reason to move it.

domain/ide/workspace/ui/template/diagnosticsOverview.html
```diff
@@ -0,0 +1,94 @@
...
+  netNewIssuesButton.addEventListener("click", () => {
+    retrieveIssues(true)
+  })
+
+  allIssuesButton.addEventListener("click", () => {
+    retrieveIssues(false)
+  })
+
+  function retrieveIssues(deltaFindingsEnabled) {
+    // retrieve issues, injected/overwritten by IDE
```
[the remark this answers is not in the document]
bastiandoetsch: Yeah, that's the plan

ShawkyZ (approved):

PR #623, opened 2024-08-13
Implement cross process locking for oauth2 refreshes [IDE-561]

bastiandoetsch:
### Description

This changes our internal storage to use a json file storage internally, implementing file locking.

### Checklist

- [x] Linted

### DIFF
--- a/internal/storage/storage.go
+++ b/internal/storage/storage.go
@@ -16,7 +16,14 @@
 
 package storage
 
-import "github.com/snyk/go-application-framework/pkg/configuration"
+import (
+	"context"
+	"time"
+
+	"github.com/adrg/xdg"
+
+	"github.com/snyk/go-application-framework/pkg/configuration"
+)
 
 type StorageCallbackFunc func(string, any)
 
@@ -27,8 +34,21 @@ type StorageWithCallbacks interface {
 }
 
 type storage struct {
-	data      map[string]any
-	callbacks map[string]StorageCallbackFunc
+	data        map[string]any
+	callbacks   map[string]StorageCallbackFunc
+	jsonStorage *configuration.JsonStorage
+}
+
+func (s *storage) Refresh(config configuration.Configuration, key string) error {
+	return s.jsonStorage.Refresh(config, key)
+}
+
+func (s *storage) Lock(ctx context.Context, retryDelay time.Duration) error {
+	return s.jsonStorage.Lock(ctx, retryDelay)
+}
+
+func (s *storage) Unlock() error {
+	return s.jsonStorage.Unlock()
 }
 
 type storageOption func(*storage)
@@ -40,13 +60,25 @@ func (s *storage) Set(key string, value any) error {
 	if callback != nil {
 		callback(key, value)
 	}
+
+	err := s.jsonStorage.Set(key, value)
+	if err != nil {
+		return err
+	}
 	return nil
 }
 
 func NewStorage(opts ...storageOption) StorageWithCallbacks {
+	// we ignore the error and just work without locking
+	file, err := xdg.ConfigFile("snyk/ls-config")
+	if err != nil {
+		panic(err)
+	}
+
 	s := &storage{
-		data:      make(map[string]any),
-		callbacks: make(map[string]StorageCallbackFunc),
+		data:        make(map[string]any),
+		callbacks:   make(map[string]StorageCallbackFunc),
+		jsonStorage: configuration.NewJsonStorage(file),
 	}
 
 	for _, opt := range opts {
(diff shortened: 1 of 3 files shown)


internal/storage/storage.go
```diff
@@ -40,13 +60,25 @@ func (s *storage) Set(key string, value any) error {
...
+	if err != nil {
+		return err
+	}
 	return nil
 }
 
 func NewStorage(opts ...storageOption) StorageWithCallbacks {
+	// we ignore the error and just work without locking
+	file, err := xdg.ConfigFile("snyk/ls-config")
+	if err != nil {
```
ShawkyZ, line 74: How will this be ignored if it will panic? Or are you relying on the recovery mechanism?

internal/storage/storage.go
```diff
@@ -27,8 +34,21 @@ type StorageWithCallbacks interface {
...
 type storage struct {
-	data      map[string]any
-	callbacks map[string]StorageCallbackFunc
+	data        map[string]any
+	callbacks   map[string]StorageCallbackFunc
+	jsonStorage *configuration.JsonStorage
+}
+
+func (s *storage) Refresh(config configuration.Configuration, key string) error {
+	return s.jsonStorage.Refresh(config, key)
```
ShawkyZ, line 43: so we will rely on GAF's jsonStorage impl? 
I think that makes sense.

I'm wondering though how it will be in the CLI. 
CLI will use JsonStorage's impl in this case that makes me wonder if we need the interface really since all impl are the same or is there a difference?

internal/storage/storage.go
```diff
@@ -40,13 +60,25 @@ func (s *storage) Set(key string, value any) error {
...
+	if err != nil {
+		panic(err)
+	}
+
 	s := &storage{
-		data:      make(map[string]any),
-		callbacks: make(map[string]StorageCallbackFunc),
+		data:        make(map[string]any),
+		callbacks:   make(map[string]StorageCallbackFunc),
+		jsonStorage: configuration.NewJsonStorage(file),
```
ShawkyZ, line 81: It might be ugly to call WithStorageFile(file)() here
but maybe we can extract the logic inside in smth like "initStorageFile" and use it here

internal/storage/storage_test.go
```diff
@@ -17,27 +17,76 @@
...
+		for i := range parallelism {
+			go func() {
+				lockErr := cut.Lock(context.Background(), time.Millisecond*100)
+				require.NoError(t, lockErr)
+				defer func(cut StorageWithCallbacks) {
+					unlockErr := cut.Unlock()
+					require.NoError(t, unlockErr)
+				}(cut)
+
+				err = cut.Set(fmt.Sprintf("test-%d", i), i)
```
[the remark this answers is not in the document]
cmars: (Maybe that's OK because the GAF tests this)

application/config/config.go
```diff
@@ -253,7 +253,11 @@ func initWorkFlowEngine(c *Config) {
 	conf.Set(configuration.FF_OAUTH_AUTH_FLOW_ENABLED, enableOAuth)
 
 	c.engine = app.CreateAppEngineWithOptions(app.WithConfiguration(conf), app.WithZeroLogger(c.logger))
-	c.storage = storage.NewStorage()
+	var storageCreationErr error
+	c.storage, storageCreationErr = storage.NewStorageWithCallbacks()
+	if storageCreationErr != nil {
+		c.logger.Err(storageCreationErr).Msg("unable to create a configuration storage object")
+	}
 	conf.SetStorage(c.storage)
```
cmars, line 261: This raised an eyebrow (wondered about a possible nil deref?) but looking at the `extendedViper` configuration.Configuration implementation in the GAF, I think this should be fine.

A comment that we're assuming the config implementation will tolerate this might be a good idea (the GAF might change in the future, and the toleration along with it).

internal/storage/storage_test.go
```diff
@@ -17,27 +17,76 @@
...
+		err = os.WriteFile(file, []byte("{}"), 0644)
+		require.NoError(t, err)
+
+		cut := NewStorage(WithStorageFile(file))
+
+		// we should not get concurrent writes to the backing map here
+		parallelism := 10
+		for i := range parallelism {
+			go func() {
+				lockErr := cut.Lock(context.Background(), time.Millisecond*100)
```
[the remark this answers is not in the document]
cmars: Does this 

```suggestion
				lockErr := cut.Lock(context.Background(), time.Millisecond)
```

make a significant impact on this test time

```
ok  	github.com/snyk/snyk-ls/application/config	4.593s	coverage: 54.4% of statements
```

?

(most recent test run in CI at time of writing)

cmars (approved):
LGTM

ShawkyZ (approved):

PR #626, opened 2024-08-16
chore(autofix): more visible logging

agronskiy:
### Description

We noticed, that while for Analysis, the `requestId` is visible at the `info` level, we don't have the same for autofix -- which would be useful when the users want to report the issues (without going through the complex procedure of switching the debug level on).

### Checklist

- [x] Linted
- [ ] ~~README.md updated, if user-facing~~
- [ ] ~~License file updated, if new 3rd-party dependency is introduced~~

### DIFF
--- a/infrastructure/code/autofix.go
+++ b/infrastructure/code/autofix.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/snyk/snyk-ls/application/config"
 	"github.com/snyk/snyk-ls/domain/snyk"
+	performance2 "github.com/snyk/snyk-ls/internal/observability/performance"
 )
 
 // AutofixUnifiedDiffSuggestion represents the diff between the original and the fixed source code.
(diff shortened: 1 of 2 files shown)


infrastructure/code/snyk_code_http_client.go
```diff
@@ -499,8 +499,8 @@ func (s *SnykCodeHTTPClient) RunAutofix(ctx context.Context, options AutofixOpti
 		logger.Err(err).Msg(failedToObtainRequestIdString + err.Error())
 		return AutofixResponse{}, err
 	}
-	logger.Debug().Msg("API: Retrieving autofix for bundle")
-	defer logger.Debug().Msg("API: Retrieving autofix done")
+	logger.Info().Msg("API: Retrieving autofix for bundle")
```
bastiandoetsch, line 502: This is executed in a loop, thus would spam the logs. Could you maybe add a log message that is displayed only once when on Info level?

agronskiy: @bastiandoetsch Basti, I'd need your advice on it. First idea is to put the single log outside of the ticking loop here https://github.com/snyk/snyk-ls/blob/c47764c304734a01cc5e2798c4525b190839c716/infrastructure/code/autofix.go#L85 but then I'd need to calculate `requestId` also outside, and pass `requestId` to https://github.com/snyk/snyk-ls/blob/c47764c304734a01cc5e2798c4525b190839c716/infrastructure/code/autofix.go#L99, is that fine with you?

infrastructure/code/autofix.go
```diff
@@ -46,6 +47,15 @@ func (s *SnykCodeHTTPClient) GetAutoFixDiffs(ctx context.Context, baseDir string
 	defer s.instrumentor.Finish(span)
 
 	var response AutofixResponse
+	requestId, err := performance2.GetTraceId(ctx)
+	if err != nil {
+		logger.Err(err).Msg(failedToObtainRequestIdString + err.Error())
```
bastiandoetsch, line 52: nitpick: `Err(err)` already writes the error message, right?

bastiandoetsch (approved):

PR #632, opened 2024-08-21
fix: do not execute code action commands twice [IDE-506][HEAD-229] 

bastiandoetsch:
### Description

Eclipse and VSCode have slightly different behaviours. To cater for Eclipse's behaviour (HEAD-229) we added execution of commands during resolution of code actions. This is not correct and only worked, because we returned an empty code action afterwards to Eclipse which overwrote the command earlier contained.

In VSCode, the command of the code action is not overwritten by the resolved code action, thus `executeCommand` was called both in resolution and normal command execution.

This implementation difference abides by protocol, but our returning changed fields after executing the command does not. The PR fixes this incorrect behaviour.

### Checklist

- [x] Linted

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ LDFLAGS_DEV := "-X 'github.com/snyk/snyk-ls/application/config.Development=true'
 
 TOOLS_BIN := $(shell pwd)/.bin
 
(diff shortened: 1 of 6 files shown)


acke (approved):
LGTM!

PR #659, opened 2024-09-11
fix: move learn service cache init to after ls startup & init

bastiandoetsch:
### Description

Moves the learn service initialization to after language server is initialized. Logs truncated refresh tokens for error debugging.

### Checklist

- [x] Linted

### DIFF
--- a/infrastructure/authentication/auth_service_impl.go
+++ b/infrastructure/authentication/auth_service_impl.go
@@ -19,6 +19,7 @@ package authentication
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io"
 	"net"
 	"net/http"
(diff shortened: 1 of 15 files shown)


acke (approved):
LGTM!

infrastructure/authentication/auth_service_impl.go
```diff
@@ -135,32 +144,99 @@ func (a *AuthenticationServiceImpl) IsAuthenticated() bool {
...
+		return false
+	}
+
+	// Check for EOF error
+	if errors.Is(err, io.EOF) {
+		return false
+	}
+
+	// as we can't enforce correct error reporting, let's do a final check on the internet connection, whether it's there
+	_, err = http.Get("https://www.google.com")
```
ShawkyZ, line 236: I'm not sure if this will be ok. It will depend on customer's infra setup. 
I don't suppose they would block google.com but it could happen.

Maybe we could instead try to HTTP GET config.SnykUI() ?

bastiandoetsch: Good idea!

ShawkyZ (approved):

PR #665, opened 2024-09-14
chore: output panic also on logger

bastiandoetsch:
### Description

### Checklist

- [ ] Linted


application/entrypoint/helpers.go
```diff
@@ -12,10 +12,19 @@ import (
...
 func OnPanicRecover() {
 	if err := recover(); err != nil {
-		fmt.Println("🚨 Panicking 🚨")
+		panickingMsg := "🚨 Panicking 🚨"
+		fmt.Println(panickingMsg)
 		fmt.Println(err)
 		debug.PrintStack()
-		er := sentry.NewSentryErrorReporter(config.CurrentConfig(), nil)
+
+		c := config.CurrentConfig()
```
ShawkyZ, line 20: my concern here is that if at some point config.CurrentConfig causes a panic.

ShawkyZ (approved):

application/entrypoint/helpers.go
bastiandoetsch: Hm. What would you recommend then?

application/server/server.go
```diff
@@ -436,6 +438,50 @@ func initializedHandler(srv *jrpc2.Server) handler.Func {
 	})
 }
 
+func startOfflineDetection(c *config.Config) {
```
bastiandoetsch, line 441: this is not called during initialization, yet.

PR #668, opened 2024-09-17
fix: Autofix timeout bug

BBerabi:
### Description

Autofix Diffs flow (web view flow) was implemented slightly differently than the code action flow. The diff flow treated an empty list of suggestions as an error and invalid response. This is not the correct behavior because the AI Fix pipeline might run successfully and there can be no valid fixes found. 

The code was polling until the suggestion's length was not 0. This causes the returned successful response with empty suggestions to be ignored and the code aggressively keeps polling until the timeout hits. The timeout takes 2 minutes whereas the result is available after several seconds. The timeout also hides the real issue from the user, making the user think something unexpected happened whereas we could not compute any valid fix. 

This PR corrects the flow for autofix diffs by

- adding status into the return value similar to how it is done for code actions
- corrects the logic in the polling function

### Checklist

- [x] Linted

I uploaded a video to my drive showing the incorrect behavior on the IDE.
https://drive.google.com/file/d/1BlHmbkW5WEiSdPQKzmjoU-5IspK20PoZ/vi

and here the corrected behaviour which immediately shows user the result of 0 fixes

https://github.com/user-attachments/assets/61b3c417-cfd0-4822-90b9-4592866e9ffe


BBerabi:
I think the failing test is not related to my changes but because of an RCE vulnerability detected by Snyk Code. Is anyone already looking at it?

ShawkyZ (commented):
Thank you Berkay
It looks good to me besides some nitpicks.

Let's pair on it on Monday. I would like to revisit / understand some stuff here, like why do we have the AutofixStatus struct and if we can simplify/refactor it a bit.

domain/ide/command/code_fix_diffs.go
```diff
@@ -83,6 +83,10 @@ func (cmd *codeFixDiffs) Execute(ctx context.Context) (any, error) {
 	}
 
 	suggestions, err := cmd.codeScanner.GetAutoFixDiffs(ctx, folderPath, relPath, issue)
+	if err == nil && len(suggestions) == 0 {
+		logger.Info().Msgf("Autofix run successfully but there were no good fixes")
+		return suggestions, err
```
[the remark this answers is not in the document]
BBerabi: done

domain/ide/command/code_fix_diffs.go
```diff
@@ -83,6 +83,10 @@ func (cmd *codeFixDiffs) Execute(ctx context.Context) (any, error) {
 	}
 
 	suggestions, err := cmd.codeScanner.GetAutoFixDiffs(ctx, folderPath, relPath, issue)
+	if err == nil && len(suggestions) == 0 {
+		logger.Info().Msgf("Autofix run successfully but there were no good fixes")
```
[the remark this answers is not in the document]
BBerabi: yes, makes sense, fixed

infrastructure/code/autofix.go
```diff
@@ -51,19 +55,36 @@ func (s *SnykCodeHTTPClient) GetAutoFixDiffs(ctx context.Context, baseDir string
...
+	logger.Debug().Msgf("Status: %s", response.Status)
+
+	if response.Status == failed.message {
+		logger.Err(err).Str("responseStatus", response.Status).Msg("autofix failed")
+		return nil, failed, err
 	}
 
-	return response.toUnifiedDiffSuggestions(baseDir, options.filePath), err
+	if response.Status == "" {
+		s.c.Logger().Err(err).Str("responseStatus", response.Status).Msg("unknown response status (empty)")
```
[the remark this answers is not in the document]
BBerabi: This handles the unexpected case where the connection breaks or something unexpected happens and response.Status stays empty. This should never be the case if the call is returned. 

To be honest, I took this from the code action flow. There we do the same logic. 

I converted the s.c.Logger to logger.

infrastructure/code/autofix.go
```diff
@@ -41,29 +40,48 @@ func (a AutofixUnifiedDiffSuggestion) GetUnifiedDiffForFile(filePath string) str
 	return a.UnifiedDiffsPerFile[filePath]
 }
 
-func (s *SnykCodeHTTPClient) GetAutofixDiffs(ctx context.Context, baseDir string, options AutofixOptions) (unifiedDiffSuggestions []AutofixUnifiedDiffSuggestion, err error) {
+func (s *SnykCodeHTTPClient) GetAutofixDiffs(ctx context.Context, baseDir string, options AutofixOptions) (
+	unifiedDiffSuggestions []AutofixUnifiedDiffSuggestion,
+	status AutofixStatus,
+	err error,
+) {
```
bastiandoetsch, line 47: on github it looks like it's weirdly formatted. just nitpicking, though.

infrastructure/code/autofix.go
```diff
@@ -41,29 +40,48 @@ func (a AutofixUnifiedDiffSuggestion) GetUnifiedDiffForFile(filePath string) str
...
-		return unifiedDiffSuggestions, err
+	response, err := s.RunAutofix(span.Context(), options)
+	if err != nil {
+		return nil, failed, err
+	}
+
+	logger.Debug().Msgf("Status: %s", response.Status)
+
+	if response.Status == failed.message {
+		logger.Error().Str("responseStatus", response.Status).Msg("autofix failed")
```
bastiandoetsch, line 70: This probably should be an info level or warn level, as it's not a technical error, but a business logic outcome, right?

infrastructure/code/autofix.go
```diff
@@ -41,29 +40,48 @@ func (a AutofixUnifiedDiffSuggestion) GetUnifiedDiffForFile(filePath string) str
...
+		logger.Error().Str("responseStatus", response.Status).Msg("unknown response status (empty)")
+		return nil, failed, errors.New("Unknown response status (empty)")
 	}
 
-	return response.toUnifiedDiffSuggestions(baseDir, options.filePath), err
+	if response.Status != completeStatus {
+		return nil, status, nil
+	}
+
+	logger.Debug().Msgf("Running toUnifiedDiffSuggestions")
```
bastiandoetsch, line 83: do we need this?

infrastructure/code/autofix.go
```diff
@@ -109,13 +127,17 @@ func (sc *Scanner) GetAutofixDiffs(
...
-				logger.Err(err).Msg("Error getting autofix suggestions")
-				return nil, err
-			}
-			if len(suggestions) > 0 {
-				return suggestions, nil
+			suggestions, fixStatus, autofixErr := codeClient.GetAutofixDiffs(span.Context(), baseDir, options)
+			if autofixErr != nil {
+				logger.Err(autofixErr).Msg("Error getting autofix suggestions")
+				return nil, autofixErr
+			} else if fixStatus.message == completeStatus {
```
bastiandoetsch, line 134: Isn't there the case missing when `autofixErr == nil` and `fixStatus.message != completeStatus`? I'd at least leave a comment for that case that it's gonna loop - which it probably should not do if the status is already `FAILED`.

bastiandoetsch (approved):
Some nitpicks - please decide yourself what to update :). Thanks for the PR, Berkay!

infrastructure/code/autofix.go
BBerabi: no, we do not, good catch, removed

infrastructure/code/autofix.go
BBerabi: Good point, Antoine in our team is currently working on providing more granular error propagation on the backend side. I took this from another place in the codebase. I keep it as ERROR for now and we will revisit this in the downstream services once we add granular errors to the backend and sast-autofix-api.

infrastructure/code/autofix.go
BBerabi: I have added a comment for that case :+1: 

if the status is already failed, then I made sure that we return an Error. So, in that case it is guaranteed to go into the branch `autofixErr != nil`

PR #676, opened 2024-09-26
fix: severity filters when delta is enabled

ShawkyZ:
### Description

* Severity filter when delta is enabled.
* Ensure Fingerprint.Identity is set.
* Added tests to ensure that Fingerprint.Identity won't be overwritten by delta finder if it exists.

### Checklist

- [ ] Linted

### DIFF
--- a/domain/ide/workspace/folder.go
+++ b/domain/ide/workspace/folder.go
@@ -315,7 +315,7 @@ func (f *Folder) processResults(scanData snyk.ScanData) {
 	go sendAnalytics(&scanData)
 
 	// Filter and publish cached diagnostics
-	f.FilterAndPublishDiagnostics(&scanData.Product)
+	f.FilterAndPublishDiagnostics(scanData.Product)
 }
 
 func (f *Folder) sendScanError(product product.Product, err error) {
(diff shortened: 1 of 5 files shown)


domain/ide/workspace/folder.go
```diff
@@ -479,46 +479,42 @@ func appendTestResults(sic snyk.SeverityIssueCounts, results []json_schemas.Test
 	return results
 }
 
-func (f *Folder) FilterAndPublishDiagnostics(p *product.Product) {
-	productIssuesByFile, err := f.getDelta(f.IssuesByProduct(), p)
+func (f *Folder) FilterAndPublishDiagnostics(p product.Product) {
+	issueByProduct := f.IssuesByProduct()
+	if len(issueByProduct[p]) == 0 {
+		return
+	}
```
bastiandoetsch, line 486: we need to send publishDiagnostics with an empty diagnostics array for files that don't have diagnostics anymore.

domain/ide/workspace/folder.go
```diff
@@ -479,46 +479,42 @@ func appendTestResults(sic snyk.SeverityIssueCounts, results []json_schemas.Test
...
-	if p != nil {
-		filteredIssues := f.filterDiagnostics(productIssuesByFile[*p])
-		f.publishDiagnostics(*p, filteredIssues)
-	} else {
-		for p, pIssues := range productIssuesByFile {
-			filteredIssues := f.filterDiagnostics(pIssues)
-			f.publishDiagnostics(p, filteredIssues)
-		}
-	}
+	filteredIssues := f.filterDiagnostics(productIssuesByFile[p])
```
bastiandoetsch, line 495: we need to send publishDiagnostics with an empty diagnostics array for files that don't have diagnostics anymore. so if after filtering, no issues are left for a file that has issues, we need to send publish diagnostics.

bastiandoetsch (approved):

PR #680, opened 2024-09-30
fix: delete git empty section if exists

ShawkyZ:
### Description
* If loading git Config fails because of an empty subsection error, attempt to delete the subsection.
* Before persisting a subsection ensure that the values are not empty.

### Checklist

- [ ] Linted

### DIFF
--- a/internal/git_config/git_config.go
+++ b/internal/git_config/git_config.go
@@ -17,16 +17,16 @@
 package gitconfig
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
-	"path/filepath"
-	"strings"
-	"time"
-
 	"github.com/go-git/go-git/v5"
 	"github.com/go-git/go-git/v5/storage/filesystem"
 	"github.com/rs/zerolog"
 	"gopkg.in/ini.v1"
+	"path/filepath"
+	"strings"
+	"sync"
 
 	config2 "github.com/go-git/go-git/v5/config"
 	"github.com/go-git/go-git/v5/plumbing"
@@ -39,6 +39,10 @@ import (
 const (
 	mainSection   = "snyk"
 	baseBranchKey = "baseBranch"
+)
+
+var (
+	mutex sync.RWMutex
 )
 
 // GetOrCreateFolderConfig queries git for the folder config of the given path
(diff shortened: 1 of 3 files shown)


internal/git_config/git_config.go
```diff
@@ -117,12 +129,18 @@ func getLocalBranches(repository *git.Repository) ([]string, error) {
 }
 
 func SetBaseBranch(logger *zerolog.Logger, config []types.FolderConfig) {
+	time.Sleep(5 * time.Second)
```
ShawkyZ, line 132: delete time.sleep before merging

internal/git_config/git_config.go
```diff
@@ -134,3 +152,50 @@ func SetOption(logger *zerolog.Logger, folderPath, key string, value string) {
 		logger.Error().Err(err).Msgf("could not store %s=%s configuration for folder %s", key, value, folderPath)
 	}
 }
+
+func GitRepoFolderPath(folderPath string) (string, error) {
```
ShawkyZ, line 156: moved this here because of circular dependency

internal/git_config/git_config.go
```diff
@@ -134,3 +152,50 @@ func SetOption(logger *zerolog.Logger, folderPath, key string, value string) {
...
+	if !strings.HasSuffix(repoPath, ".git") {
+		repoPath = filepath.Join(repoPath, ".git")
+	}
+
+	return repoPath, nil
+}
+
+// DeleteEmptySnykSubsection This is a migration function to be executed if empty subsections exists
+func DeleteEmptySnykSubsection() error {
+	gitFolderPath, err := GitRepoFolderPath("")
```
ShawkyZ, line 180: go-git treats empty folder as the current folder. which is annoying.

here I'm loading folderPath="" to reproduce the same behavior that we might get.

internal/git_config/git_config.go
bastiandoetsch: The name could be a bit more expressive... Maybe `GitRepoPath`?

internal/git_config/git_config.go
```diff
@@ -134,3 +152,50 @@ func SetOption(logger *zerolog.Logger, folderPath, key string, value string) {
...
+	if err != nil {
+		return err
+	}
+	configPath := filepath.Join(gitFolderPath, "config")
+	cfg, err := ini.Load(configPath)
+	if err != nil {
+		return err
+	}
+	// Construct the section name with the empty subsection
+	sectionToDelete := fmt.Sprintf("%s \"\"", mainSection)
```
bastiandoetsch, line 190: using backticks might help readibility here

internal/vcs/git_utils.go
bastiandoetsch, line 1: Maybe we should just merge git-config package with vcs?

bastiandoetsch (approved):

PR #690, opened 2024-10-07
fix/autofix: fix html encoding and add test for toUnifiedDiffSuggestions

BBerabi:
### Description

The file reading process for DeepCode AI Fix was wrong because it was not using the `DecodePath` function as it is done for other places in the codebase. This leads to AI Fix not working on files with spaces or other characters which are HTML encoded. 

There were also no tests for the conversion function `toUnifiedDiffSuggestions`. 

This PR:

- adds a test for `toUnifiedDiffSuggestions`.
- fixed the bug around HTML encoding in filepath by using `DecodePath`
- adds another test suit to test `toUnifiedDiffSuggestions` under HTML Encoding
### Checklist

- [x] Linted

Before the fix

[ai_fix_space_bug.webm](https://github.com/user-attachments/assets/e104c8a0-5be6-494d-b443-38366de4786d)

After the fix

[ai_fix_spaces_fixed.webm](https://github.com/user-attachments/assets/6f64335a-7d2c-4195-a51f-3725ea0a0ec8)

https://snyksec.atlassian.net/browse/ML-888


ShawkyZ (approved):
🚀✨

PR #696, opened 2024-10-10
fix: autofix unify code action and web panel view flows

BBerabi:
### Description

This PR unifies the two autofix flows, namely the code action flow (`GetAutofixSuggestions`) and the web panel view flow (`GetAutofixDiffs`). We unify the two flows by creating a common function `GetAutofixResponse` which is responsible for calling the backend and getting the result. Both of the flows use this function and at the very end convert the response to the desired format. By that, we avoid code duplication across the two flows.

https://snyksec.atlassian.net/browse/ML-880

### Checklist

- [x] Linted


BBerabi:
@ShawkyZ All the checks have passed upon re-try, fyi

infrastructure/code/autofix.go
```diff
@@ -47,41 +47,60 @@ func (s *SnykCodeHTTPClient) GetAutofixDiffs(ctx context.Context, baseDir string
...
+	defer logger.Info().Msg("Finished obtaining autofix diffs")
+
+	autofixResponse, status, err := s.GetAutofixResponse(ctx, baseDir, options)
+	if err != nil {
+		return nil, status, err
+	}
+	return autofixResponse.toUnifiedDiffSuggestions(baseDir, options.filePath), status, err
+}
+
+func (s *SnykCodeHTTPClient) GetAutofixResponse(ctx context.Context, baseDir string, options AutofixOptions) (
```
ShawkyZ, line 63: baseDir arg is not used

ShawkyZ (approved):

PR #700, opened 2024-10-11
chore: add logging to snyk code bundles for troubleshooting

bastiandoetsch:
### Description

### Checklist

- [ ] Linted


infrastructure/code/snyk_code_http_client.go
```diff
@@ -134,50 +134,47 @@ func (s *SnykCodeHTTPClient) CreateBundle(
...
 	s.c.Logger().Debug().Str("method", method).Msg("API: Create done")
 	return bundle.BundleHash, bundle.MissingFiles, nil
 }
 
-func (s *SnykCodeHTTPClient) doCall(ctx context.Context,
-	method string,
-	path string,
-	requestBody []byte,
-) (responseBody []byte, _ error) {
+func (s *SnykCodeHTTPClient) doCall(ctx context.Context, method string, path string, requestBody []byte) ([]byte, error, *http.Response) {
```
cmars, line 151: Questions about this, based on my limited understanding and lack of context:
- Why was the `*http.Response` return value added? I don't see any of the `doCall` callers using it.
- Where is this response's `.Body` getting closed? I'm concerned this is going to leak.

ISTM that the returned response should be used by all the callers to `defer resp.Body.Close()` to avoid the latter.

infrastructure/code/snyk_code_http_client.go
```diff
@@ -313,11 +310,13 @@ func (s *SnykCodeHTTPClient) ExtendBundle(
 	removedFiles []string,
 ) (string, []string, error) {
 	method := "code.ExtendBundle"
-	s.c.Logger().Debug().Str("method", method).Msg("API: Extending bundle for " + strconv.Itoa(len(files)) + " files")
-	defer s.c.Logger().Debug().Str("method", method).Msg("API: Extend done")
-
 	span := s.instrumentor.StartSpan(ctx, method)
 	defer s.instrumentor.Finish(span)
+	requestId, _ := performance2.GetTraceId(span.Context())
```
cmars, line 315: This error should be checked. If it shouldn't fail the method, it should at least be logged for diagnostic / debugging purposes.

There's a golangci-lint rule we should turn on that would catch this (for a followup!)

cmars (commented):
The logging changes seem reasonable but there are other drive-by changes here I have some concerns about, see below.

infrastructure/code/snyk_code_http_client.go
bastiandoetsch: 1. response is used in line 330
2. body is always closed in doCall()

infrastructure/code/snyk_code_http_client.go
bastiandoetsch: We can log it - but it should not fail anything - if this causes an error, it's nothing we can or should do about. I'd rather remove the error return value for `GetTraceId`

infrastructure/code/snyk_code_http_client.go
```diff
@@ -434,11 +446,11 @@ func (s *SnykCodeHTTPClient) analysisRequestBody(options *AnalysisOptions) ([]by
 	return requestBody, err
 }
 
-func (s *SnykCodeHTTPClient) checkResponseCode(r *http.Response) error {
-	if r.StatusCode >= 200 && r.StatusCode <= 299 {
+func (s *SnykCodeHTTPClient) checkResponseCode(statusCode int) error {
+	if statusCode >= 200 && statusCode <= 400 {
```
cmars, line 450: This has consequences for redirects. If we're not currently equipped to follow them, maybe we should treat them as "unexpected"?

cmars (commented):
Logging changes LGTM but question about the drive-by change on expected status code.

infrastructure/code/snyk_code_http_client.go
bastiandoetsch: What do you mean with `unexpected`? Why can't we follow the redirects? Doesn't the http client follow them unless we deactivate it? I definitely see an error though that it's `400` and not `399`.

cmars: I forgot net/http.Client automatically follows them (if you use the default Client settings). Nevermind.

Good catch on the 400, I'd completely missed that. What caused me to comment on this originally, an drive-by change like this can introduce other side-effects, which are offtopic from the original PR (add logging). So maybe this is best left for a separate PR?

bastiandoetsch: It was in the spirit of handling the HTTP 400, but I agree, this is something we should separate out.

application/di/test_init.go
```diff
@@ -53,6 +54,8 @@ func TestInit(t *testing.T) {
 	defer initMutex.Unlock()
 	t.Helper()
 	c := config.CurrentConfig()
+	// we want to isolate CLI fake installs
+	c.CliSettings().SetPath(filepath.Join(t.TempDir(), "fake-cli"))
```
cmars, line 58: Does this (and/or the server test changes below) fix the test flakes I've seen around the CLI downloads? :excited-conan-obrien-emoji:

cmars (approved):
Question about some test changes below. I'd still like to see those landed as separate commits (if not separate PRs) for better isolation & bisect-ability but I'm not going to block this any further.

application/di/test_init.go
bastiandoetsch: I hope so - had to add it at a few places.

PR #704, opened 2024-10-17
fix: golangci config, linter errors

bastiandoetsch:
### Description

Activated linters and fixed issues.

### Checklist

- [x] Linted


.golangci.yaml
```diff
@@ -69,7 +59,6 @@ linters-settings:
       - name: error-strings
       - name: error-naming
       - name: exported
-        disable-stuttering-check: true
```
cmars, line 72: I'd prefer to keep this rule, it's common best practice for Go style. See https://google.github.io/styleguide/go/decisions#repetition

.golangci.yaml
```diff
@@ -40,7 +31,6 @@ linters-settings:
   gosimple:
     checks: ['all']
   govet:
-    check-shadowing: true
```
cmars, line 43: While this check can be annoying at times (`err` variable tends to get tangled up in it more often than I'd like) I've seen it as a net win for keeping less experienced Go developers on the rails and avoiding some of Go's rougher edges.

cmars (commented):
Mostly looks amazing! Absolutely love that we're turning on more of these rules.

A few comments & questions below.

.golangci.yaml
bastiandoetsch: It's included in enable-all :)

.golangci.yaml
bastiandoetsch: It does not seem to be a valid option no more (got a warning in goland). I'll look it up at the source, though.

cmars: Indeed, `golangci-lint` says:

```
WARN [config_reader] The configuration option `linters.govet.check-shadowing` is deprecated. Please enable `shadow` instead, if you are not using `enable-all`. 
```

So it's valid, but could be renamed.

bastiandoetsch: We are using `enable-all`, though, so it's redundant.

infrastructure/oss/code_actions.go
```diff
@@ -169,7 +169,11 @@ func (i *ossIssue) getQuickfixEdit(affectedFilePath string) string {
 }
 
 func (i *ossIssue) getUpgradedPathParts() (string, string) {
-	rootDependencyUpgrade := strings.Split(i.UpgradePath[1].(string), "@")
+	s, ok := i.UpgradePath[1].(string)
+	if !ok {
+		return "", ""
+	}
```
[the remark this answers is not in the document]
bastiandoetsch: This is used to construct code actions - and if the cast fails, it would just not add a code action, but not pop up an error to the client. I can make this explicit though, by updating the signature. I don't like to panicking without cause...

application/server/server_test.go
```diff
@@ -156,14 +155,6 @@ func startServer(callBackFn onCallbackFn, jsonRPCRecorder *testutil.JsonRPCRecor
 	return loc
 }
 
-func Test_serverShouldStart(t *testing.T) {
-	loc, _ := setupServer(t)
-
-	si := loc.Server.ServerInfo()
-
-	fmt.Println(strings.Join(si.Methods, "\n"))
```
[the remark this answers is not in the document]
bastiandoetsch: Nah, there is no code of us in there. And setupServer(t) is tested used(=tested) everywhere.

.golangci.yaml
bastiandoetsch: Revive is not enabled (see L141), and having a stuttering check would be good, right? So this should be fine.

infrastructure/code/issue_enhancer.go
```diff
@@ -307,9 +308,11 @@ func getShardKey(folderPath string, authToken string) string {
 }
 
 func issueTitle(issue snyk.Issue) string {
-	if issue.AdditionalData != nil && issue.AdditionalData.(snyk.CodeIssueData).Title != "" {
-		return issue.AdditionalData.(snyk.CodeIssueData).Title
+	codeIssueData, ok := issue.AdditionalData.(snyk.CodeIssueData)
+	if ok {
+		if issue.AdditionalData != nil && codeIssueData.Title != "" {
```
ShawkyZ, line 313: if AdditionalData is nil it will panic

ShawkyZ (approved):

PR #705, opened 2024-10-17
fix: unc file path to URI conversion

ShawkyZ:
### Description

Fix path to URI conversion in case of UNC path.

### Checklist

- [x] Linted


internal/uri/uri_util.go
```diff
@@ -48,15 +51,58 @@ func FolderContains(folderPath string, path string) bool {
...
+}
+
 // PathToUri converts a path to a DocumentURI
 func PathToUri(path string) sglsp.DocumentURI {
-	return sglsp.DocumentURI(uri.File(path))
+	// in case of UNC file path. uri.File returns file://// which IDEs can't interpret correctly
+	// file://// is still a valid UNC, but we have to replace with file:// for the IDEs to interpret it correctly
+	parsedUri := uri.File(path)
+	uriAsString := string(parsedUri)
+	if strings.HasPrefix(uriAsString, UncFileScheme) {
```
ShawkyZ, line 102: am thinking if we should check for OS here.
but I think it makes sense to correct it anyway

internal/uri/uri_util.go
```diff
@@ -48,15 +51,58 @@ func FolderContains(folderPath string, path string) bool {
...
+	}
+
 	if !strings.HasPrefix(u, fileScheme) && strings.HasPrefix(u, eclipseWorkspaceFolderScheme) {
 		u = strings.Replace(u, eclipseWorkspaceFolderScheme, fileScheme, 1)
 	}
+
 	return uri.New(u).Filename()
 }
 
+// isUNCFileURI checks if the provided file URI represents a UNC path.
```
cmars, line 68: Please update the godoc comment to match the function

internal/uri/uri_util.go
```diff
@@ -19,16 +19,19 @@ package uri
...
 	"regexp"
+	"runtime"
 	"strings"
 
 	sglsp "github.com/sourcegraph/go-lsp"
 	"go.lsp.dev/uri"
 )
 
 const fileScheme = "file://"
+const UncFileScheme = "file:////"
```
cmars, line 34: Does this need to be exported?

```suggestion
const uncFileScheme = "file:////"
```

internal/uri/uri_util.go
```diff
@@ -48,15 +51,58 @@ func FolderContains(folderPath string, path string) bool {
...
+		return ""
+	}
+
+	if parsedURI.Scheme != "file" {
+		return ""
+	}
+
+	// A non-empty Host indicates a UNC path
+	if len(parsedURI.Host) > 0 {
+		uncPath := fmt.Sprintf(`\\%s%s`, parsedURI.Host, parsedURI.Path)
```
cmars, line 88: Is `parsedURI.Path` guaranteed to be absolute? Will this mix forward and back-slashes?

internal/uri/uri_util_test.go
```diff
@@ -17,6 +17,7 @@
 package uri
 
 import (
+	"github.com/snyk/snyk-ls/internal/testutil"
```
cmars, line 20: This import should be grouped separately below (first-party imports). I think there should be a linter rule that checks for this (gci?)

cmars:
Some of the review items above seem like they should have been caught by the linter. I wonder if #704 will improve catching them?

application/server/server_smoke_test.go
```diff
@@ -796,6 +796,52 @@ func Test_SmokeSnykCodeFileScan(t *testing.T) {
...
+	c := testutil.SmokeTest(t, false)
+	c.SetSnykCodeEnabled(true)
+	cleanupChannels()
+	di.Init()
+
+	var cloneTargetDir, err = testutil.SetupCustomTestRepo(t, t.TempDir(), nodejsGoof, "0336589", c.Logger())
+	if err != nil {
+		t.Fatal(err, "Couldn't setup test repo")
+	}
+	defer func(path string) { _ = os.RemoveAll(path) }(cloneTargetDir)
```
bastiandoetsch, line 811: Not needed, `t.Tempdir` is taking care of its cleanup

application/server/server_smoke_test.go
```diff
@@ -796,6 +796,52 @@ func Test_SmokeSnykCodeFileScan(t *testing.T) {
...
+
+	uncPath := "\\\\localhost\\" + strings.Replace(cloneTargetDir, ":", "$", 1)
```
bastiandoetsch, line 813: such an ugly syntax ;)

application/server/server_smoke_test.go
```diff
@@ -796,6 +796,52 @@ func Test_SmokeSnykCodeFileScan(t *testing.T) {
 	assert.Eventually(t, checkForPublishedDiagnostics(t, testPath, 6, jsonRPCRecorder), maxIntegTestDuration, 10*time.Millisecond)
 }
 
+func Test_SmokeUncFilePath(t *testing.T) {
+	testutil.OnlyOnWindows(t, "testing windows UNC file paths")
```
bastiandoetsch, line 800: This should be marked as Integration Test (Integration Tests run on WIndows)

internal/uri/uri_util.go
bastiandoetsch: I now know why the Java people introduced the `Path` type. Mixing of slashes should be tackled in line 90. Maybe I'd do a `filepath.Clean` before returning.

application/server/server_smoke_test.go
ShawkyZ: hahahahaha. That's my cheap workaround

internal/uri/uri_util.go
bastiandoetsch: I think this needs to be validated in all IDEs we support (IntelliJ, VS, VSC, Eclipse).

internal/uri/uri_util_test.go
bastiandoetsch: I would refrain from using testutil at all in internal packages to avoid circular dependencies.

internal/uri/uri_util_test.go
```diff
@@ -36,6 +37,19 @@ func TestPathFromUri(t *testing.T) {
 	assert.Equal(t, filepath.Clean(dir+"/asdf"), PathFromUri(u)) // Eclipse case
 }
 
+func TestPathFromUri_UNC(t *testing.T) {
+	testutil.OnlyOnWindows(t, "testing windows UNC file paths")
```
bastiandoetsch, line 41: please consider checking this manually instead of using testutil.

application/server/server_smoke_test.go
ShawkyZ: deleted it from other tests too.

internal/uri/uri_util.go
ShawkyZ: > // On Windows, Clean does not modify the volume name other than to replace
// occurrences of "/" with `\`.
// For example, Clean("//host/share/../x") returns `\\host\share\x`.

maybe we just need to use clean instead of normalizing the the slashes

bastiandoetsch (approved):

PR #722, opened 2024-11-12
fix: add ls part for iac ignore button [IDE-683]

DariusZdroba:
### Description

This is the logic needed for adding the IaC ignore button.
It has the HTML which has the button, hidden by default, enabled in **IntelliJ**.

A `FilePath` was also added to the TemplateData which is needed to trigger the IgnoreService for the specific issue, and the before named 'Path' variable was renamed to `ResourcePath` which is a more accurate name for what it represents.

A file path was also added to the template for enabling to execute the ignore issue comand which also needs file path. 
### Checklist

- [x] Linted

### DIFF
--- a/infrastructure/iac/iac_html.go
+++ b/infrastructure/iac/iac_html.go
@@ -39,7 +39,8 @@ type TemplateData struct {
 	SeverityIcon template.HTML
 	Description  template.HTML
 	Remediation  template.HTML
-	Path         template.HTML
+	ResourcePath template.HTML
+	FilePath     template.HTML
 	Nonce        template.HTML
 }
 
@@ -88,7 +89,8 @@ func (service *HtmlRenderer) GetDetailsHtml(issue snyk.Issue) string {
 		SeverityIcon: html.SeverityIcon(issue),
 		Description:  html.MarkdownToHTML(issue.Message),
 		Remediation:  html.MarkdownToHTML(issueData.Resolve),
-		Path:         formatPath(issueData.Path),
+		ResourcePath: formatPath(issueData.Path),
+		FilePath:     template.HTML(issue.Path()),
 		Nonce:        template.HTML(nonce),
 	}
 
--- a/infrastructure/iac/template/index.html
+++ b/infrastructure/iac/template/index.html
@@ -83,12 +83,12 @@
       {{end}}
     </div>
   </section>
-  <!-- Ignore In Line Actions -->
+  <!-- Ignore Issue in File -->
   {{if not .Issue.IsIgnored}}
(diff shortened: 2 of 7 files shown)


infrastructure/iac/iac_html.go
```diff
@@ -39,7 +39,8 @@ type TemplateData struct {
 	SeverityIcon template.HTML
 	Description  template.HTML
 	Remediation  template.HTML
-	Path         template.HTML
+	ResourcePath template.HTML
```
DariusZdroba, line 42: The "Path" name is a bit ambiguous, in this case the Path refers to where the issue can be found, example: 
"[DocId: 0] > input > spec > template > spec > hostNetwork" .

infrastructure/iac/template/index.html
```diff
@@ -82,7 +83,17 @@ <h2>References</h2>
       {{end}}
     </div>
   </section>
+  <!-- Ignore In Line Actions -->
```
bastiandoetsch, line 86: nitpick: this comment is probably not accurate anymore

infrastructure/iac/template/index.html
```diff
@@ -82,7 +83,17 @@ <h2>References</h2>
...
+  {{if not .Issue.IsIgnored}}
```
bastiandoetsch, line 87: comment: strictly speaking this is not necessary, but nice for later, when ignores are provided from API/Findings model.

infrastructure/iac/template/index.html
```diff
@@ -82,7 +83,17 @@ <h2>References</h2>
...
+  <footer id="ignore-container" class="ignore-action-container hidden">
+    <div id="ignore-actions">
+      <div class="actions row">
+        <button id="ignore-file-issue" issue="{{.Issue.ID}}" path="{{.ResourcePath}}" filePath="{{.FilePath}}" class="ignore-button secondary">Ignore in this file</button>
```
bastiandoetsch, line 91: nitpick: why not rename `path` here, too?

infrastructure/iac/template/index.html
```diff
@@ -82,7 +83,17 @@ <h2>References</h2>
...
+      </div>
+    </div>
+  </footer>
+  {{end}}
```
bastiandoetsch, line 95: question: what is `end`ed here?

bastiandoetsch (approved):

bastiandoetsch:
Consider testing that the template is filled correctly for the ignores stuff ;)

infrastructure/iac/template/index.html
DariusZdroba: That is the end of the {{if not .Issue.IsIgnored}} statement.

PR #725, opened 2024-11-25
refactor: remove unnecessary css

DariusZdroba:
### Description

The CSS was really cluttered and difficult to understand being in the HTML templates in `<style>` tags, I separated them into their own file, now we inject them the same way we did so far with **IaC**, through the TemplateData that is used to dynamically render variables into the HTML.

In addition to that, some minor styling changes were pushed down from **VSCode** in order to remove them from being plugin-specific and having a more centralized styling possibility in **Language Server**.

### Checklist

- [x] Linted


infrastructure/iac/template/styles.css
```diff
@@ -20,7 +21,7 @@
 }
 
 ::-webkit-scrollbar-thumb:hover {
-  background: var(--scroll-bar-hover-background-color);
+  background: var();
```
ShawkyZ, line 24: don't we need the var --scroll-bar-hover-background-color here

infrastructure/iac/template/styles.css
```diff
@@ -30,8 +31,9 @@
 }
 
 body {
+  padding: 0;
   background-color: var(--background-color);
-  font-family: var(--default-font);
+  font-family: var(--default-font), sans-serif;
```
ShawkyZ, line 36: I think this sans-serif is not needed?

infrastructure/oss/template/styles.css
```diff
@@ -0,0 +1,164 @@
...
+  color: var(--link-color);
+  cursor: pointer;
+  text-decoration: none;
+}
+
+body {
+  padding: 0;
+  background-color: var(--background-color);
+  color: var(--text-color);
+  font-family: var(--default-font), sans-serif;
```
ShawkyZ, line 41: same here, shouldn't --default-font be enough?

ShawkyZ (approved):

infrastructure/iac/template/styles.css
DariusZdroba: We don't replace it anywhere in IntelliJ or VSCode I'm not sure it's actually used, maybe I'm missing something? Or maybe we use it in VS or Eclipse ?

Missjones2 (commented):
Merge

PR #730, opened 2024-12-09
fix: data flow order for code [IDE-763]

DariusZdroba:
### Description

This PR fixes DataFlowItems not being displayed in order, previously we used a map, which doesn't guarantee order, according to Go docs **_"The iteration order over maps is not specified and is not guaranteed to be the same from one iteration to the next"_** 

To solve this while also being able to have a map and access the DataFlowItems by the `fileName` as key, a slice is created that uniquely holds the fileNames (keys) of the map.

In the HTML template we iterate over the keys slice ( which is indexed in order ), and access the map by the key from current iteration.
###  Checklist

- [x] Linted

### DIFF
--- a/infrastructure/code/template/details.html
+++ b/infrastructure/code/template/details.html
@@ -234,14 +234,14 @@
                 Data Flow - {{len .DataFlow}} {{if gt (len .DataFlow) 1}}steps{{else}}step{{end}}
               </h2>
               <div class="data-flow-body">
-                {{range $fileName, $dataFlowItems := .DataFlowTable}}
+                {{range $fileName := .DataFlowKeys}}
                 <div class="data-flow-file">
                   {{$.FileIcon}}
                   <span class="data-flow-file-name">{{$fileName}}</span>
                 </div>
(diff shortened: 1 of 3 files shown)


infrastructure/code/template/details.html
```diff
@@ -234,14 +234,14 @@ <h2 class="data-flow-header">
...
-                {{range $fileName, $dataFlowItems := .DataFlowTable}}
+                {{range $fileName := .DataFlowKeys}}
                 <div class="data-flow-file">
                   {{$.FileIcon}}
                   <span class="data-flow-file-name">{{$fileName}}</span>
                 </div>
                 <table class="data-flow-table">
                   <tbody>
-                    {{range $dataFlowItems}}
+                    {{range index $.DataFlowTable $fileName}}
```
DariusZdroba, line 244: the `$fileName` is taken from the `.DataFlowKeys` slice, which is indexed so we know the order, then the DataFlowTable map is accessed by said `$fileName` key

ShawkyZ (approved):

PR #738, opened 2024-12-12
fix: update endpoint url

andrewrobinsonhodges-snyk:
### Description

This fix ensures the API URL gets correctly updated if it changes due to redirect in the authentication process. 

The updateCredentials function gets called multiple times; when the auth token refreshes and when the authentication process completes (which is when we get the new URL).

The old behaviour was to only notify of changed configuration when the token changes, however we now also need to notify when the URL has changed. This fix changes the logic of the updateCredentials method to make the API URL an optional parameter on the credentials, and ensures that we always send the notification when the URL changes, even if the token has stayed the same.

### Checklist

- [x] Linted
- [n/a] README.md updated, if user-facing
- [n/a] License file updated, if new 3rd-party dependency is introduced


infrastructure/authentication/auth_service_impl.go
```diff
@@ -151,6 +151,10 @@ func getPrioritizedApiUrl(customUrl string, engineUrl string) string {
 	defaultUrl := config.DefaultSnykApiUrl
 	customUrl = strings.TrimRight(customUrl, "/ ")
 
+	if customUrl == "https://api.pre-prod.snyk.io" && engineUrl != "" {
+		return engineUrl
+	}
+
```
ShawkyZ, line 157: this was only used for testing, right?
In that case, please delete it.

andrewrobinsonhodges-snyk: Yes, now removed.

ShawkyZ (approved):

PR #752, opened 2025-01-13
feat: Sending a user event when fixing inside the editor

antoine-snyk:
### Description

When the user triggers the issue action in code, no user event was sent for analytics. This addresses it by sending a FIX_APPLIED event.

EDIT: Using the existing event type FIX_APPLIED for simplicity.

### Checklist

- [x] Linted

### DIFF
--- a/infrastructure/code/issue_enhancer.go
+++ b/infrastructure/code/issue_enhancer.go
@@ -223,6 +223,8 @@ func (b *IssueEnhancer) autofixFunc(ctx context.Context, issue snyk.Issue,
 
 				// send feedback asynchronously, so people can actually see the changes done by the fix
 				go func() {
+					b.SnykCode.SubmitAutofixFeedback(ctx, fix.FixId, FixAppliedUserEvent)
+
 					actionCommandMap, err := b.autofixFeedbackActions(fix.FixId)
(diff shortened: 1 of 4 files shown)


infrastructure/code/issue_enhancer.go
```diff
@@ -221,6 +221,9 @@ func (b *IssueEnhancer) autofixFunc(ctx context.Context, issue snyk.Issue,
 					return nil
 				}
 
+				// send user event for analytics
+				b.SnykCode.SubmitAutofixFeedback(ctx, fix.FixId, FixBlindlyAppliedUserEvent)
+
```
bastiandoetsch, line 226: This should be done in the background in a go routine - there's no case for bubbling up errors and blocking user-flows on the call. Maybe just move it into the go func in line 228?

bastiandoetsch: Also, please write a test for this functionality.

antoine-snyk: Yes! Good point indeed.

bastiandoetsch (approved):

bastiandoetsch:
Merged. Thx!

PR #764, opened 2025-01-30
Add Option for Pre-Scan command, fix auth race [IDE-908][IDE-939]

bastiandoetsch:
### Description

- add command that is called before a scan
- refactor: move dependency-less test helpers to package testsupport
- adds a helper target to the `Makefile`: `lint-fix` fixes findings the linter can fix
- fixes the following auth race:

A potential race condition related to authentication has been identified in IntelliJ. 

1. IDE starts up CLI with initializationOptions incl. OAuth2 token
2. LS starts WHOAMI workflow with (expired) credentials
3. WHOAMI workflow refreshes OAuth2 token
4. LS saves token and triggers $/hasAuthenticated with new token
5. At the same time, IDE sends the current configuration with DidChangeConfiguration, with the old credentials
6. LS updates the credentials to the old credentials
7. Next call to API returns not authenticated and creds are deleted in LS
8. LS forwards the logout to the IDE.

Solution: when updating the token, do not accept super-seded tokens (check Expiry)

**Note**
The high line change count is due to the license update, which pulled in a few files into the license folder.

### Checklist

- [x] Linted

### DIFF
--- a/domain/snyk/scanner/scanner.go
+++ b/domain/snyk/scanner/scanner.go
@@ -21,19 +21,18 @@ import (
 	"sync"
 	"time"
 
+	"github.com/snyk/snyk-ls/application/config"
+	"github.com/snyk/snyk-ls/domain/ide/initialize"
 	"github.com/snyk/snyk-ls/domain/scanstates"
 	"github.com/snyk/snyk-ls/domain/snyk"
 	"github.com/snyk/snyk-ls/domain/snyk/persistence"
-	"github.com/snyk/snyk-ls/internal/vcs"
-
-	"github.com/snyk/snyk-ls/application/config"
-	"github.com/snyk/snyk-ls/domain/ide/initialize"
 	"github.com/snyk/snyk-ls/infrastructure/authentication"
 	"github.com/snyk/snyk-ls/infrastructure/snyk_api"
 	"github.com/snyk/snyk-ls/internal/notification"
 	"github.com/snyk/snyk-ls/internal/observability/performance"
 	"github.com/snyk/snyk-ls/internal/product"
 	"github.com/snyk/snyk-ls/internal/types"
+	"github.com/snyk/snyk-ls/internal/vcs"
 )
 
 var (
@@ -265,6 +264,7 @@ func (sc *DelegatingConcurrentScanner) Scan(
 	}
 
 	sc.scanNotifier.SendInProgress(folderPath)
+	folderConfig := sc.c.FolderConfig(folderPath)
 	gitCheckoutHandler := vcs.NewCheckoutHandler(sc.c.Engine().GetConfiguration())
 
 	waitGroup := &sync.WaitGroup{}
@@ -282,10 +282,19 @@ func (sc *DelegatingConcurrentScanner) Scan(
 
 				scanSpan := sc.instrumentor.StartSpan(span.Context(), "scan")
 
+				err := sc.executePreScanCommand(ctx, sc.c, s.Product(), folderConfig, folderPath, true)
+				if err != nil {
+					logger.Err(err).Str("path", folderPath).Send()
+					sc.scanNotifier.SendError(scanner.Product(), folderPath, err.Error())
+					sc.scanStateAggregator.SetScanDone(folderPath, scanner.Product(), false, err)
+					return
+				}
+
 				// TODO change interface of scan to pass a func (processResults), which would enable products to stream
 				foundIssues, scanError := sc.internalScan(scanSpan.Context(), s, path, folderPath)
+
+				// this span allows differentiation between processing time and scan time
 				sc.instrumentor.Finish(scanSpan)
-				sc.scanStateAggregator.SetScanDone(folderPath, scanner.Product(), false, scanError)
 
 				// now process
 				data := snyk.ScanData{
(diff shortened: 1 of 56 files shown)


domain/snyk/scanner/scanner.go
```diff
@@ -303,8 +304,14 @@ func (sc *DelegatingConcurrentScanner) Scan(
...
 				go func() {
 					defer referenceBranchScanWaitGroup.Done()
-					isReferenceScanNeeded := path == folderPath
-					if isReferenceScanNeeded {
+					notReferenceScan := path == folderPath
+
+					err := sc.executePreScanCommand(ctx, sc.c, s.Product(), folderPath, notReferenceScan)
+					if err != nil {
+						logger.Err(err).Str("path", folderPath).Send()
+					}
```
ShawkyZ, line 312: I would trigger it only in case of fullscan. Ref scan shouldn't be triggered with partial scans

domain/snyk/scanner/scanner.go
```diff
@@ -303,8 +304,14 @@ func (sc *DelegatingConcurrentScanner) Scan(
...
+
+					if notReferenceScan {
```
ShawkyZ, line 314: why? It's a reference scan. 
reason for this condition is to make sure we are running in a full scan and not a partial one. So only in this case a reference scan would be needed

domain/snyk/scanner/scanner.go
```diff
@@ -349,6 +356,24 @@ func (sc *DelegatingConcurrentScanner) Scan(
...
+	ctx context.Context,
+	c *config.Config,
+	p product.Product,
+	workDir string,
+	notReferenceScan bool,
+) error {
+	folderConfig := c.FolderConfig(workDir)
+	commandConfig := folderConfig.ScanCommandConfig
+
+	if commandConfig == nil || commandConfig[p].PreScanCommand == "" || commandConfig[p].PreScanOnlyReferenceFolder && notReferenceScan {
```
ShawkyZ, line 369: can we abstract this with a func that gets a perscan command and takes the product as an input?

domain/snyk/scanner/scanner_test.go
```diff
@@ -124,3 +128,28 @@ func TestScan_whenProductScannerEnabled_SendsInProgress(t *testing.T) {
 
 	assert.NotEmpty(t, mockScanNotifier.InProgressCalls())
 }
+
+func TestDelegatingConcurrentScanner_executePreScanCommand(t *testing.T) {
+	c := testutil.UnitTest(t)
```
ShawkyZ, line 133: nice

infrastructure/oss/cli_scanner.go
```diff
@@ -234,17 +231,22 @@ func (cliScanner *CLIScanner) scanInternal(ctx context.Context, path string, com
...
+		additionalParameters = append(additionalParameters, folderConfigArgs...)
+	}
 
 	if len(additionalParameters) > 0 {
 		for _, parameter := range additionalParameters {
 			// if the sdk needs additional parameters, add them (Python plugin, I look at you. Yes, you)
 			// the given parameters take precedence, meaning, a given python configuration will overrule
 			// the automatically determined config
-			isDuplicateParam := storedConfig.SliceContainsParam(commandLineArgs, parameter) || storedConfig.SliceContainsParam(folderConfigArgs, parameter)
+			isDuplicateParam := storedConfig.SliceContainsParam(commandLineArgs, parameter)
```
ShawkyZ, line 249: yup that's cleaner

internal/testsupport/environment.go
```diff
@@ -1,5 +1,5 @@
 /*
- * © 2022 Snyk Limited All rights reserved.
+ * © 2022-2025 Snyk Limited
```
ShawkyZ, line 2: fix copyright date

internal/testsupport/files.go
```diff
@@ -1,5 +1,5 @@
 /*
- * © 2022 Snyk Limited All rights reserved.
+ * © 2022-2025 Snyk Limited
```
ShawkyZ, line 2: same

internal/testsupport/json_rpc_recorder.go
```diff
@@ -1,5 +1,5 @@
 /*
- * © 2022 Snyk Limited All rights reserved.
+ * © 2022-2025 Snyk Limited
```
ShawkyZ, line 2: same

internal/types/lsp.go
```diff
@@ -528,15 +528,28 @@ type WorkspaceFoldersChangeEvent struct {
 	Removed []WorkspaceFolder `json:"Removed,omitempty"`
 }
 
+// ScanCommandConfig allows to define a command that is run before (PreScanCommand)
+// or after (PostScanCommand) a scan. It will only be run for the
+// referenceFolder / referenceScan (in case of delta) if the corresponding
+// parameter PreScanOnlyReferenceFolder / PostScanOnlyReferenceFolder is set.
+// Else it will run for all scans.
+type ScanCommandConfig struct {
+	PreScanCommand              string `json:"command,omitempty"`
```
ShawkyZ, line 537: curious if we will ever have the case that the command should run in both cases.
I can't think of one though.

ShawkyZ: I also find this much more readable than the `Pair` type that we discussed earlier.

domain/snyk/scanner/scanner.go
ShawkyZ: I noticed a problem here while reviewing this. The reference scan is waiting on wd scan to succeed before triggering the go routine.

I created this PR 
#766

acke (approved):
LGTM

PR #767, opened 2025-02-03
hotfix: fix fileFilter and reduce memory usage [IDE-926] [IDE-662]

ShawkyZ:
### Description

* VS sends folderPath with a trailing slash which breaks fileFilter.
* Improve memory usage

### Checklist

- [x] Linted

### DIFF
--- a/infrastructure/filesystem/filesystem.go
+++ b/infrastructure/filesystem/filesystem.go
@@ -18,8 +18,8 @@
 package filesystem
 
 import (
+	"bufio"
 	"os"
-	"strings"
 
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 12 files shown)


infrastructure/filesystem/filesystem.go
```diff
@@ -35,20 +35,27 @@ func (f *Filesystem) GetLineOfCode(filePath string, line int) (string, error) {
 	if line <= 0 {
 		return "", errors.Errorf("invalid line number %d", line)
```
ShawkyZ, line 36: this might look unnecessary.
But the idle state memory usage was very high because probably GC didn't clean the file content var hear yet.

bastiandoetsch (approved):

PR #770, opened 2025-02-07
chore: migrate AI Fix JS logic from IDEs to LS [IDE-655]

DariusZdroba:
### Description

The logic between IDE's for handling AI Fix functionality is mostly the same, this means having separate scripts in each IDE for this feature also means a lot of duplication and difficulty in maintenance due to different nuances in the implementations.

This PR migrates AI Fix Logic, centralizes it in LS and unifies the logic, allowing injecting only the IDE Specific logic that is absolute necessary.
### Checklist

- [ ] Linted

### DIFF
--- a/infrastructure/code/template/details.html
+++ b/infrastructure/code/template/details.html
@@ -23,14 +23,16 @@
   <meta charset="utf-8" />
   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
   <meta http-equiv="Content-Security-Policy"
-    content="style-src 'self' 'nonce-${nonce}' 'nonce-ideNonce' https://fonts.googleapis.com;
-             script-src 'nonce-${nonce}' https://fonts.googleapis.com;
+    content="style-src 'self' 'nonce-{{.Nonce}}' 'nonce-ideNonce' https://fonts.googleapis.com;
+             script-src 'nonce-{{.Nonce}}' https://fonts.googleapis.com;
              font-src 'self' https://fonts.gstatic.com;" />
 
-  <style nonce="${nonce}">
+  <style nonce="{{.Nonce}}">
     {{.Styles}}
   </style>
-  ${ideStyle}
+  <style nonce="{{.Nonce}}">
+    ${ideStyle}
+  </style>
 </head>
 
 <body>
@@ -345,7 +347,7 @@
   {{end}}
 
   <!-- Scripts -->
-  <script nonce=${nonce}>
+  <script nonce="{{.Nonce}}">
     /**
      * Represents a single line change in a commit.
      * @typedef {Object} ExampleLine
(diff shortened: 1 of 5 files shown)


infrastructure/code/template/details.html
```diff
@@ -497,11 +483,17 @@ <h2 class="ai-fix-header">
       if (exampleNext) {
         exampleNext.onclick = nextExample;
       }
-
       showCurrentExample();
     });
   </script>
-  ${ideScript}
+  <!-- Embedded script from LS  -->
+  <script nonce=${nonce}>
```
bastiandoetsch, line 490: wondering if this shouldn't be injected too with .Nonce or such.

infrastructure/code/template/details.html
```diff
@@ -497,11 +483,17 @@ <h2 class="ai-fix-header">
...
+    {{.Scripts}}
+  </script>
+  <!-- Custom IDE specific scripts  -->
+  <script nonce=${nonce}>
```
bastiandoetsch, line 494: same here

infrastructure/code/template/details.html
```diff
@@ -345,7 +345,7 @@ <h2 class="ai-fix-header">
   {{end}}
 
   <!-- Scripts -->
-  <script nonce="${nonce}">
+  <script nonce=${nonce}>
```
bastiandoetsch, line 348: shouldn't this be determined by LS?

infrastructure/code/template/scripts.js
```diff
@@ -0,0 +1,186 @@
...
+  toggleElement(diffElem, 'show');
+  toggleElement(applyFixButton, 'show');
+  diffNumElem.innerText = showSuggestion.length.toString();
+  diffNum2Elem.innerText = showSuggestion.length.toString();
+
+  diffSelectedIndexElem.innerText = (diffSelectedIndex + 1).toString();
+
+  const diffSuggestion = showSuggestion[diffSelectedIndex];
+  // IntelliJ way of getting file ? TODO: Investigate
+  const filePath = showSuggestion.filePath ? showSuggestion.filePath : getFilePathFromFix(diffSuggestion);
```
bastiandoetsch, line 170: There's a todo here: is intellij doing it differently than other IDEs?

bastiandoetsch (approved):

infrastructure/code/template/scripts.js
DariusZdroba: Yes, I added it, I will need to remove the comment, 
the `suggestions` coming from IntelliJ have direct access to `filePath` and other attributes we are accessing in the AI Fix logic, the  `suggestions` coming from VSCode need `suggestion.diffs` to access those attributes, so far we never directly use the `suggestion` from VSCode only the `suggestion.diffs` so it might be a matter of how we deserialize the response from LS in VSCode, maybe we should change that.

DariusZdroba: For now I added a function which checks if we should use the `suggestion` or `suggestion.diffs` but we should probably tackle the root cause.

infrastructure/code/template/details.html
DariusZdroba: Good point, I will investigate this

infrastructure/code/template/details.html
DariusZdroba: Previously all `${nonce}` was replaced in IDEs but for some reason there was a typo here and the `${nonce}` had quotes, we never replaced this in IDE's so this was never even loaded, should probably add the `.Nonce` from LS as you suggested above anyway.

PR #782, opened 2025-02-21
fix: add correct lesson url for license issues [IDE-971]

DariusZdroba:
### Description

Learn links in the Issue Description Panel for License type Issues were always wrong, this was fixed with adding a link to URL: https://learn.snyk.io/lesson/license-policy-management/  when a License type issue is encountered.
### Checklist

- [ ] Linted


infrastructure/learn/service.go
```diff
@@ -237,7 +237,9 @@ func (s *serviceImpl) GetLesson(ecosystem string, rule string, cwes []string, cv
 
 	if len(lessons) >= 1 {
 		lesson = &lessons[0]
-		lesson.Url += "?loc=ide"
+		if !strings.Contains(lesson.Url, "?loc=ide") {
+			lesson.Url += "?loc=ide"
+		}
```
bastiandoetsch, line 242: this should be quite an easy thing to test :)

bastiandoetsch (approved):

infrastructure/learn/service.go
DariusZdroba: Good point, the test for the learn already included this but I will also add a test for lessons which are not coming from `license` issues

PR #784, opened 2025-02-25
feat: enable calling mcp commands via ls commands [IDE-923]

bastiandoetsch:
### Description

This enables calling of MCP commands via LSP

### Checklist

- [x] Linted


andrewrobinsonhodges-snyk (approved):
We went through this on a call. All good.

PR #787, opened 2025-03-03
fix: cleanup inconsistent auth configuration when needed [IDE-916]

bastiandoetsch:
### Description

- Reset authentication configuration if authentication method does not match the provider, and the API reports a credential problem.
- Retrieve SDK configuration always with the path of the working directory

### Checklist

- [x] Linted

### DIFF
--- a/infrastructure/oss/cli_scanner.go
+++ b/infrastructure/oss/cli_scanner.go
@@ -238,7 +238,7 @@ func (cliScanner *CLIScanner) updateArgs(workDir types.FilePath, commandLineArgs
 	folderConfigArgs := folderConfig.AdditionalParameters
 
 	// this asks the client for the current SDK and blocks on it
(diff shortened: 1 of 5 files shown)


infrastructure/oss/cli_scanner.go
```diff
@@ -238,7 +238,7 @@ func (cliScanner *CLIScanner) updateArgs(workDir types.FilePath, commandLineArgs
 	folderConfigArgs := folderConfig.AdditionalParameters
 
 	// this asks the client for the current SDK and blocks on it
-	additionalParameters := cliScanner.updateSDKs(workDir)
+	additionalParameters := cliScanner.updateSDKs(folderConfig.FolderPath)
```
bastiandoetsch, line 241: SDK config should always be taken from the actual IDE path, not the reference path.

PR #792, opened 2025-03-07
fix: lazy instantiation deepcode LLM binding 

DariusZdroba:
The DeepCode LLM Binding initializing was moved from `init.go` to avoid further complication with uninitialized components, as of know it is only used in the `ai_fix_handler.go` so it is initialized as it's used.

It will be initialized `WithEndpoint` using the current API Endpoint configured.
### Description

### Checklist

- [x] Linted

### DIFF
--- a/application/di/init.go
+++ b/application/di/init.go
@@ -179,19 +179,11 @@ func initInfrastructure(c *config.Config) {
 }
 
 func initApplication(c *config.Config) {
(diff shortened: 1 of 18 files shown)


ShawkyZ (approved):

bastiandoetsch (approved):
some tests would be nice :)

PR #795, opened 2025-03-10
fix: move issue view option filtering to the LS to not display ignored diagnostics in editor [IDE-899]

rrama:
### Description

Move the filtering of Issue View Options onto the language server.

### Checklist

- [x] Linted


application/config/config.go
```diff
@@ -457,7 +459,8 @@ func (c *Config) IntegrationName() string {
 func (c *Config) IntegrationVersion() string {
 	return c.engine.GetConfiguration().GetString(configuration.INTEGRATION_VERSION)
 }
-func (c *Config) FilterSeverity() types.SeverityFilter { return c.filterSeverity }
+func (c *Config) FilterSeverity() types.SeverityFilter     { return c.filterSeverity }
+func (c *Config) IssueViewOptions() types.IssueViewOptions { return c.issueViewOptions }
```
bastiandoetsch, line 463: reading should probably set a read lock on the mutex, e.g. `c.m.RLock()`

rrama: Done.

bastiandoetsch (approved):

PR #796, opened 2025-03-10
fix: render AI fix elements based on show-hidden class [IDE-1020]

DariusZdroba:
### Description
GO Template conditional statements are used to choose when to render certain elements based on different states for AI Fix (**Error**, **Success**, **Scan In Progress**...)

When Generating an AI Fix, we require values of certain attributes from the HTML Elements which are injected through the 
Go Templating mechanism. Eg. `issue-id` -  if the element holding the attribute is not rendered, the logic will not work.

We rely on `toggleElement` function to toggle the `hidden` and `show` classes to hide/display elements so a better way of working with this mechanism is to use the conditional statements for the `hidden` and `show` class name instead of the element/section itself, this way the element is still available if needed and can be toggled based on scripts.

### Checklist

- [ ] Linted

### DIFF
--- a/infrastructure/code/template/details.html
+++ b/infrastructure/code/template/details.html
@@ -301,8 +301,7 @@
                 ⚠️ Failed to generate the fix. Please try again.
               </div>
               <div class="sn-fix-wrapper">
-                <button id="retry-generate-fix" folder-path="{{.FolderPath}}" file-path="{{.FilePath}}"
-                        issue-id="{{.IssueId}}" class="button generate-ai-fix">✨ Retry AI fix</button>
+                <button id="retry-generate-fix" class="button generate-ai-fix">✨ Retry AI fix</button>
               </div>
             </section>
           </div>
--- a/infrastructure/code/template/details.html
+++ b/infrastructure/code/template/details.html
@@ -210,20 +210,16 @@
           <!-- AI Fix -->
           <div id="ai-fix-wrapper" class="{{if not .HasAIFix}}hidden{{end}}">
             <!-- AI fix buttons -->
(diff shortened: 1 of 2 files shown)


ShawkyZ (approved):

infrastructure/code/template/details.html
```diff
@@ -274,47 +266,44 @@ <h3 class="ai-fix-header">No AI Fix Available:</h3>
...
-            {{if eq .AiFixDiffStatus "ERROR"}}
-            <section id="fixes-error-section" class="sn-ai-fix-error delimiter-top">
+            <section id="fixes-error-section" class="sn-ai-fix-error delimiter-top {{if eq .AiFixDiffStatus "ERROR"}} show {{else}} hidden {{end}}">
               <div id="info-no-diffs" class="font-light">
                 ⚠️ Failed to generate the fix. Please try again.
               </div>
               <div class="sn-fix-wrapper">
-                <button id="retry-generate-fix" folder-path="{{.FolderPath}}" file-path="{{.FilePath}}"
-                        issue-id="{{.IssueId}}" class="button generate-ai-fix">✨ Retry AI fix</button>
+                <button id="retry-generate-fix" class="button generate-ai-fix">✨ Retry AI fix</button>
```
ShawkyZ, line 304: why did you remove the additional attributes ? are they not needed in case of retry ?

DariusZdroba: No, the retry function just calls `generateAiFix()` function which takes the attributes from the GenerateAiFix Button not from the RetryButton, we never use these attributes.

PR #803, opened 2025-03-12
fix: catch nilpointer in sarif response from UploadAndAnalyzeWithIgnores

acke:
### Description

 catch nilpointer in sarif response from UploadAndAnalyzeWithIgnores

### Checklist

- [ ] Linted

### DIFF
--- a/infrastructure/code/code.go
+++ b/infrastructure/code/code.go
@@ -451,6 +451,12 @@ func (sc *Scanner) UploadAndAnalyzeWithIgnores(ctx context.Context, path types.F
 	if err != nil || ctx.Err() != nil {
 		return []types.Issue{}, err
 	}
+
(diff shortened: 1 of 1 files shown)


infrastructure/code/code.go
```diff
@@ -451,6 +451,12 @@ func (sc *Scanner) UploadAndAnalyzeWithIgnores(ctx context.Context, path types.F
 	if err != nil || ctx.Err() != nil {
 		return []types.Issue{}, err
 	}
+
+	if sarif == nil {
```
ShawkyZ, line 455: interesting .. so we didn't get any error but the sarif was nil?

acke: We got a nilpointer.

PR #805, opened 2025-03-14
fix: use affectedFilePath as targetFilePath for OSS issues

ShawkyZ:
### Description
Issue.DisplayTargetFile had the wrong value if the displayTargetFile exists in a subfolder.
Which broke our cache deduplication logic. Since issues that were included in subfolders had basically DisplayTargetFile value of= root + fileName
This resulted in issues having the same displayTargetFile when they exist in other folders.

This fix will use the pre-calculated affectedFilePAth and will set it's value to issue.DisplayTargetFile.

### Checklist

- [x] Linted

### DIFF
--- a/infrastructure/oss/types.go
+++ b/infrastructure/oss/types.go
@@ -19,7 +19,6 @@ package oss
 import (
 	"fmt"
 	"net/url"
-	"path/filepath"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


infrastructure/oss/types.go
```diff
@@ -110,15 +109,10 @@ type Annotation struct {
...
+func (i *ossIssue) toAdditionalData(scanResult *scanResult, matchingIssues []snyk.OssIssueData, affectedFilePath types.FilePath) snyk.OssIssueData {
 	var additionalData snyk.OssIssueData
 
-	targetFilePath := filepath.Base(scanResult.DisplayTargetFile)
-	if scanResult.Path != "" {
-		targetFilePath = filepath.Join(scanResult.Path, targetFilePath)
-	}
-
-	additionalData.Key = util.GetIssueKey(i.Id, targetFilePath, i.LineNumber, i.LineNumber, 0, 0)
+	additionalData.Key = util.GetIssueKey(i.Id, string(affectedFilePath), i.LineNumber, i.LineNumber, 0, 0)
```
bastiandoetsch, line 115: could we have a test for that that checks the different combinations?

bastiandoetsch (approved):

PR #806, opened 2025-03-18
fix: increase resilience of authentication [IDE-974]

bastiandoetsch:
### Context
CLI has default functions to determine configuration values. [One of these is for the OAuth2 feature flag](https://github.com/snyk/cli/blob/f6fc5f7ed643fbb78826cb95104057c9662664b3/cliv2/cmd/cliv2/configuration.go#L13), which governs the availability of the OAuth2 authentication method. This default function disables the feature flag when a Snyk API token is detected in the environment variables.

At the same time, [when initializing the CLI, Viper loads environment variables](https://github.com/snyk/cli/blob/22177c253b5401807dd344533cca83af6edb1f64/cliv2/cmd/cliv2/main.go#L493), thus a `SNYK_TOKEN` environment variable is loaded into the CLI global engine configuration. 

Also, the [OAuth2 authenticator](https://github.com/snyk/go-application-framework/blob/af80929f69be397dae0f1a600fa3ed4da0478296/pkg/auth/authenticator.go#L27) in the go-application-framework is [governed by](https://github.com/snyk/go-application-framework/blob/a2347b6eee3a85b38ce432e0ce6a78893c0bb4e7/pkg/auth/oauth2authenticator.go#L251) the feature flag.

The OAuth2 authenticator is created for each API call, when the custom roundtripper [is added](https://github.com/snyk/go-application-framework/blob/fed20c61fcebab2b8ae48354c17f42582de71b4d/pkg/networking/networking.go#L237).

### Problem
This combination of defaults and implementation leads to OAuth2 being disabled, regardless of the selected authentication method if the `SNYK_TOKEN` or `SNYK_OAUTH_TOKEN` variables are set. When the roundtripper is created for the API call, the (valid) OAuth2 Token cannot be used, as the authenticator reports `IsSupported == false`. Instead, an API Token authenticator is added. As no API token is available in the configuration, the authentication header is not added, and the API call fails. When this is happening for the `whoami` API call that is used by `IsAuthenticated` in Language Server, it leads to an automatic logout. Authentication with OAuth2 is not possible.

### Solution
We [overwrite the default function for the feature flag](https://github.com/snyk/snyk-ls/blob/6f99e64ee1e3a17974736324e1933d23cc5ed3c0/application/config/config.go#L268) and [always set the feature flag to true.](https://github.com/snyk/snyk-ls/blob/6f99e64ee1e3a17974736324e1933d23cc5ed3c0/application/config/config.go#L271).

### Checklist

- [x] Linted

### DIFF
--- a/infrastructure/authentication/auth_configuration.go
+++ b/infrastructure/authentication/auth_configuration.go
@@ -37,7 +37,6 @@ func Token(c *config.Config, errorReporter error_reporting.ErrorReporter) Authen
 		c.Storage().UnRegisterCallback(auth.CONFIG_KEY_OAUTH_TOKEN)
 	}
 
-	conf.Set(configuration.FF_OAUTH_AUTH_FLOW_ENABLED, false)
 	conf.Unset(configuration.AUTHENTICATION_BEARER_TOKEN)
 	conf.Unset(auth.CONFIG_KEY_OAUTH_TOKEN)
 	return NewCliAuthenticationProvider(c, errorReporter)
@@ -47,7 +46,6 @@ func Token(c *config.Config, errorReporter error_reporting.ErrorReporter) Authen
 // the auth service parameter is needed, as the oauth2 provider needs a callback function
 func Default(c *config.Config, authenticationService AuthenticationService) AuthenticationProvider {
 	conf := c.Engine().GetConfiguration()
-	conf.Set(configuration.FF_OAUTH_AUTH_FLOW_ENABLED, true)
 	conf.Unset(configuration.AUTHENTICATION_TOKEN)
 	credentialsUpdateCallback := func(_ string, value any) {
 		// an empty struct marks an empty token, so we stay with empty string if the cast fails
--- a/internal/storage/storage_test.go
+++ b/internal/storage/storage_test.go
@@ -49,6 +49,41 @@ func Test_StorageCallsRegisterCallbacksForKeys(t *testing.T) {
 	}, 5*time.Second, time.Millisecond, "callback was not called")
 }
 
+func Test_StorageCallsEmptyValueShouldCleanValueFromFile(t *testing.T) {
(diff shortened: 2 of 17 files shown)


internal/storage/storage_test.go
```diff
@@ -49,6 +49,41 @@ func Test_StorageCallsRegisterCallbacksForKeys(t *testing.T) {
...
 }
 
+func Test_StorageCallsEmptyValueShouldCleanValueFromFile(t *testing.T) {
+	called := make(chan bool, 1)
+	callbacks := map[string]StorageCallbackFunc{}
+	file := filepath.Join(t.TempDir(), t.Name())
+	myCallback := func(_ string, _ any) { called <- true }
+
+	key := "test"
+	value := "test"
```
andrewrobinsonhodges-snyk, line 59: It would be better to make these uniquely identifiable strings "test key" and "test value", for instance, so we can easily spot if the wrong field is being used.

andrewrobinsonhodges-snyk (approved):
Looks good. One comment on your test, but not critical to address.

infrastructure/authentication/auth_configuration.go
```diff
@@ -37,7 +37,6 @@ func Token(c *config.Config, errorReporter error_reporting.ErrorReporter) Authen
 		c.Storage().UnRegisterCallback(auth.CONFIG_KEY_OAUTH_TOKEN)
 	}
 
-	conf.Set(configuration.FF_OAUTH_AUTH_FLOW_ENABLED, false)
```
ShawkyZ, line 40: could this cause any side effects when using token auth?

ShawkyZ (approved):

infrastructure/authentication/auth_configuration.go
bastiandoetsch: Nope :)

PR #810, opened 2025-03-24
fix: delta improve matching on dir name and similar path names [IDE-1051]

ShawkyZ:
### Description
Changes:
* Added ContentRoot to Issue Type.
* Improve Fuzzy matcher dir check to use issue relative paths.
* Fixed an issue caused by automatically setting GlobalIdentity when Fingerprint is set.
* Bumped file schema version to force persisted cache deletion.

### Checklist

- [x] Linted


internal/delta/fuzzy_matcher.go
```diff
@@ -251,17 +252,38 @@ func matchDistance(baseIssue Identifiable, currentIssue Identifiable) (float64,
...
+
+	longestPossiblePath := math.Max(math.Max(baseRelativePathSeparatorCount, currentRelativePathSeparatorCount), relativePathDistanceSeparatorCount)
+
+	return 1 - relativePathDistanceSeparatorCount/longestPossiblePath
+}
 
-	return 1 - relativePathDistance/longestPossiblePath
+func normalizePath(path types.FilePath) types.FilePath {
+	pathSeparator := string(os.PathSeparator)
+	normalizedPath := strings.ReplaceAll(strings.ToLower(string(path)), pathSeparator, "/")
```
bastiandoetsch, line 285: please don't normalize posix filesystems like this, only do it for windows

bastiandoetsch (approved):

PR #811, opened 2025-03-25
fix: use pointers for optional LSP settings [IDE-899]

rrama:
### Description

Previously we had no way to distinguish between the settings object being omitted and being set to "empty" value for the struct. As such we were ignoring settings objects that were set to the same as the "empty" struct value, as it would have broken / overridden the default LSP settings for clients that did not support the setting. With this change to using pointers, we can distinguish between a client not sending the settings object (because they don't support it yet) and a client sending the settings object but deliberately setting the values to be the same as the empty struct.

### Checklist

- [x] Linted
 - N/A


internal/util/ptr_of.go
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+
+package util
+
+// PtrOf creates a pointer on the heap for the value provided.
+// Because in Go you can't do something like `takesPtr(&(returnsStruct()))`.
+// So instead do `takesPtr(PtrOf(returnsStruct()))`.
+func PtrOf[T any](value T) *T {
+	pointerToValue := new(T) // Heap may be safer than `&value`
```
ShawkyZ, line 23: I love golang lol

application/config/config.go
```diff
@@ -581,29 +581,27 @@ func (c *Config) SetSnykAdvisorEnabled(enabled bool) {
...
-	emptySeverityFilter := types.SeverityFilter{}
-	if severityFilter == emptySeverityFilter {
+	if severityFilter == nil {
 		return false
 	}
-	filterModified := c.filterSeverity != severityFilter
+	filterModified := c.filterSeverity != *severityFilter
 	c.logger.Debug().Str("method", "SetSeverityFilter").Interface("severityFilter", severityFilter).Msg("Setting severity filter:")
-	c.filterSeverity = severityFilter
+	c.filterSeverity = *severityFilter
```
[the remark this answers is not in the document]
rrama: After discussing with @ShawkyZ in person, we decided that there is no need for the `config` values to be pointers, as it would just add unnecessary complications.

internal/util/ptr_of.go
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package util
+
+// PtrOf creates a pointer on the heap for the value provided.
+// Because in Go you can't do something like `takesPtr(&(returnsStruct()))`.
+// So instead do `takesPtr(PtrOf(returnsStruct()))`.
+func PtrOf[T any](value T) *T {
```
acke, line 22: We could possibly consider using this helper function instead. 

// ptr returns a pointer to the input value.
func ptr[T any](v T) *T {
	return &v
}

acke (approved):
Nitpick on the helper function. And pr looks good to me.

PR #814, opened 2025-03-28
fix: improve folder configuration handling and storage [IDE-668]

bastiandoetsch:
### Description

- Add test for GetOrCreateFolderConfig to verify existing folder configs
- Move folder config notification from command handler to server initialization
- Improve storage handling by properly refreshing config from storage
- Fix GetStoredConfig to return errors instead of creating new config on failure
- Update token handling with EmptyToken method for better readability
- Fix storage persistence by ensuring tokens are properly persisted

### Checklist

- [ ] Linted

### DIFF
--- a/application/server/notification.go
+++ b/application/server/notification.go
@@ -85,7 +85,7 @@ func registerNotifier(c *config.Config, srv types.Server) {
 	logger := c.Logger().With().Str("method", "registerNotifier").Logger()
 	callbackFunction := func(params any) {
 		for !c.IsLSPInitialized() {
-			logger.Trace().Msg("waiting for lsp initialization to be finished...")
+			logger.Debug().Msg("waiting for lsp initialization to be finished...")
 			time.Sleep(300 * time.Millisecond)
 		}
 		switch params := params.(type) {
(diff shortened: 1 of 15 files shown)


acke (approved):

domain/ide/command/folder_handler.go
```diff
@@ -34,7 +34,6 @@ const DoTrust = "Trust folders and continue"
 const DontTrust = "Don't trust folders"
 
 func HandleFolders(c *config.Config, ctx context.Context, srv types.Server, notifier noti.Notifier, persister persistence.ScanSnapshotPersister, agg scanstates.Aggregator) {
-	go sendFolderConfigsNotification(c, notifier)
```
ShawkyZ, line 37: we need to sendFolderConfig after a folder is trusted as well.

I would actually prefer the existing behavior

application/server/notification.go
```diff
@@ -85,7 +85,7 @@ func registerNotifier(c *config.Config, srv types.Server) {
 	logger := c.Logger().With().Str("method", "registerNotifier").Logger()
 	callbackFunction := func(params any) {
 		for !c.IsLSPInitialized() {
-			logger.Debug().Msg("waiting for lsp initialization to be finished...")
+			logger.Trace().Msg("waiting for lsp initialization to be finished...")
```
ShawkyZ, line 88: I would leave it to be debug. It might spam but it's always useful to know that LS didn't start yet.
IIRC we also have other issues when trace is enabled

ShawkyZ (approved):

PR #816, opened 2025-03-31
fix: update code-client-go version and LS usage [IDE-1098]

DariusZdroba:
### Checklist

- [ ] Linted


bastiandoetsch (approved):

ShawkyZ (changes requested):
I'm blocking merging here because service isn't up yet and we can't test if it works e2e. 
Also we need to make sure that the service is deployed to all clusters before merging this change

ShawkyZ (approved):

PR #820, opened 2025-04-03
fix: WIP ignore Issue View Options when CCI disabled [IDE-899]

rrama:
### Description

Before the IDE's IVOs would still be applied even if Code Consistent Ignores was not enabled in the org.
Now all issues will be returned if CCI is not enabled regardless of the IDE's settings.

Also this now uses the new GAF API for fetching feature flags.
We can figure out the configuration caching another day, for now it calls the API 5 times during the scan to fetch the feature flag.

### Checklist

- [x] Linted
 - N/A


domain/ide/command/get_feature_flag_status.go
```diff
@@ -57,16 +55,8 @@ func (cmd *featureFlagStatus) Execute(_ context.Context) (any, error) {
 		return nil, errors.New("invalid feature flag name argument")
 	}
 
-	ff := snyk_api.FeatureFlagType(ffStr)
-	ffResponse, err := cmd.apiClient.FeatureFlagStatus(ff)
+	// No need for language server to enforce which feature flags exist or not.
+	enabled := config.CurrentConfig().Engine().GetConfiguration().GetBool(ffStr)
```
rrama, line 59: Doesn't work as it isn't the GAF internal name for the feature flag...

rrama:
Splitting this PR into two PRs:
- https://github.com/snyk/snyk-ls/pull/823
- https://github.com/snyk/snyk-ls/pull/824

PR #821, opened 2025-04-03
fix: filter code actions by Issue View Options [IDE-899]

rrama:
### Description

This uses the new GAF API for fetching feature flags.
We can figure out the configuration caching another day, for now it calls the API 4 times during the scan to fetch the feature flag.

### Checklist

 - No tests added, I am a bad person.
- [x] Linted
 - N/A
 - N/A


acke (approved):
LGTM!

PR #823, opened 2025-04-08
fix: ignore Issue View Options when CCI disabled [IDE-899]

rrama:
### Description

Before the IDE's IVOs would still be applied even if Code Consistent Ignores was not enabled in the org.
Now all issues will be returned if CCI is not enabled regardless of the IDE's settings.

Also this now uses the new GAF API for fetching feature flags.
We can figure out the configuration caching another day, for now it calls the API 5 times during the scan to fetch the feature flag.

### Checklist

 - None added, just edited existing to pass.
- [x] Linted
 - N/A
 - N/A


infrastructure/code/code_test.go
```diff
@@ -522,7 +523,6 @@ func Test_Scan(t *testing.T) {
 		assert.Nil(t, params)
 	})
 
-	//nolint:dupl // test cases differ by a boolean
 	t.Run("Should run existing flow if feature flag is disabled", func(t *testing.T) {
 		c := testutil.UnitTest(t)
 		snykCodeMock := &FakeSnykCodeClient{C: c}
```
acke, line 528: should we update this test case as we do with the "Should run new flow if feature flag is enabled"?

acke (approved):
Appoved with the nitpick comments below.

PR #835, opened 2025-04-14
fix: git config migration precedence

ShawkyZ:
### Description
When merging git config set precedence of dynamic fields to the config loaded from Config if it exists.

### Checklist

- [ ] Linted


acke (approved):
LGTM

PR #848, opened 2025-04-28
fix: make JS IE compatible [IDE-1079] [IDE-1075]

rrama:
### Description

Needs to work for Visual Studio.
- Added a polyfill for a datepicker.
  - Couldn't think of a better way to prevent the license being deleted by the GH action, so I had to do a bodge solution.
- Scroll IAW form into view when opened.
- Also fixed a null pointer exception.

### Checklist

 - Untested.
- [ ] Linted
 - N/A
 - N/A

### Screenshots

No difference for VS Code and JetBrains.

Eclipse:
Screenshot 2025-04-29 at 11 45 34
Visual Studio:
Screenshot 2025-04-29 at 11 47 38
Note: The date format is always American, even though I have my locale as the GB/UK.


infrastructure/code/template/scripts.js
```diff
@@ -1,20 +1,22 @@
...
 function _unsupportedIterableToArray(r, a) { if (r) { if ("string" == typeof r) return _arrayLikeToArray(r, a); var t = {}.toString.call(r).slice(8, -1); return "Object" === t && r.constructor && (t = r.constructor.name), "Map" === t || "Set" === t ? Array.from(r) : "Arguments" === t || /^(?:Ui|I)nt(?:8|16|32)(?:Clamped)?Array$/.test(t) ? _arrayLikeToArray(r, a) : void 0; } }
 function _arrayLikeToArray(r, a) { (null == a || a > r.length) && (a = r.length); for (var e = 0, n = Array(a); e < a; e++) n[e] = r[e]; return n; }
-function toggleElement(element, toggle) {
+function dispatchEvent(element, dispatchEventName) {
```
ShawkyZ, line 4: I would rather leave it as toggleElement. It's easier to use
also we can always use the es5 impl.

rrama: `classList` is working and is already in use on [L110](https://github.com/snyk/snyk-ls/pull/848/files#diff-96f094e12b3f3f4466e90c9a406bf9a26d4e016d09c29d7295a5bc9dd14e0e80L110), so I assumed it was a remnant of supporting a prior version, but if that is not the case then I am happy to make it all use old-school manipulation.

rrama: Reverted.

acke (approved):

PR #849, opened 2025-04-28
feat: send analytic for create ignore

acke:
### Description

Added analytics for create ignore request.

### Checklist

- [ ] Linted

### DIFF
--- a/domain/ide/command/ignores_request.go
+++ b/domain/ide/command/ignores_request.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/snyk/snyk-ls/application/config"
 	"github.com/snyk/snyk-ls/domain/snyk"
+	"github.com/snyk/snyk-ls/infrastructure/analytics"
 	"github.com/snyk/snyk-ls/infrastructure/code"
 	"github.com/snyk/snyk-ls/internal/notification"
 	"github.com/snyk/snyk-ls/internal/types"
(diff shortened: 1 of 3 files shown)


domain/ide/command/ignores_request.go
```diff
@@ -287,3 +290,8 @@ func (cmd *submitIgnoreRequest) executeIgnoreWorkflow(engine workflow.Engine, wo
 	}
 	return nil
 }
+
+func (cmd *submitIgnoreRequest) sendIgnoreRequestAnalytics() {
+	event := analytics.NewAnalyticsEventParam("create ignores")
```
rrama, line 295: Do we not want to call it like `ignores.create`, then later have `ignores.edit`, so then you could filter on `ignores.*`?
I don't know how the analytics currently work, just a thought. If nothing else uses this format, then ignore me.

infrastructure/analytics/analytics.go
```diff
@@ -35,6 +35,14 @@ import (
 
 var analyticsMu = sync.RWMutex{}
 
+func NewAnalyticsEventParam(interactionType string) types.AnalyticsEventParam {
```
rrama, line 38: Consider `NewAnalyticsEventSuccessParam`?

rrama (approved):
LGTM!
Just made a few comments to consider.

PR #850, opened 2025-04-29
feat: add instrumentation on config [IDE-1070]

acke:
### Description

Added analytics for delta config change.
Added a timestamp bug in authentication analytics, where timestamp is nil when it should have a value.

### Checklist

- [ ] Linted


rrama (approved):
LGTM!

PR #852, opened 2025-04-29
fix: oldValue delta should be string in analytics

acke:
### Description

### Checklist

- [ ] Linted

### DIFF
--- a/application/server/configuration.go
+++ b/application/server/configuration.go
@@ -25,9 +25,6 @@ import (
 
 	"github.com/rs/zerolog"
 
-	"github.com/snyk/snyk-ls/infrastructure/analytics"
-	"github.com/snyk/snyk-ls/internal/storedconfig"
-
 	"github.com/creachadair/jrpc2"
 	"github.com/creachadair/jrpc2/handler"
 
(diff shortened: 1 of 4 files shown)


application/server/configuration.go
```diff
@@ -256,7 +256,7 @@ func updateDeltaFindings(c *config.Config, settings types.Settings) {
 
 	modified := c.SetDeltaFindingsEnabled(enable)
 	if modified {
-		sendWorkspaceConfigChanged(c, "enableDeltaFindings", oldValue, strconv.FormatBool(enable))
+		sendWorkspaceConfigChanged(c, "enableDeltaFindings", strconv.FormatBool(oldValue), strconv.FormatBool(enable))
```
rrama, line 259: Oops 🙈 At this point should we not just send them both as an actual boolean or is it easier in Snowflake as a string?

infrastructure/analytics/analytics.go
```diff
@@ -35,16 +35,19 @@ import (
 
 var analyticsMu = sync.RWMutex{}
 
-func NewAnalyticsEventParam(interactionType string, err error) types.AnalyticsEventParam {
+func NewAnalyticsEventParam(interactionType string, err error, path types.FilePath) types.AnalyticsEventParam {
 	status := string(analytics.Success)
 	if err != nil {
 		status = string(analytics.Failure)
 	}
+	targetId, _ := instrumentation.GetTargetId(string(path), instrumentation.AutoDetectedTargetId)
```
rrama, line 43: Do we just not care if it can't be found? I suppose it will be nil, which is fine.

rrama (approved):
Swell. Made 2 comments, but they are up to you to decide.

application/server/configuration.go
acke: Yes, will change!

PR #856, opened 2025-05-05
fix: mcp host check [IDE-1119]

ShawkyZ:
### Description

### Checklist

- [ ] Linted


internal/mcp/llm_binding.go
```diff
@@ -162,31 +163,43 @@ var allowedHostnames = map[string]bool{
 	"localhost": true,
 	"127.0.0.1": true,
 	"::1":       true,
+	"":          true,
```
bastiandoetsch, line 166: why is this necessary?

ShawkyZ: clients can send empty host header, which is the case with Cline atm

PR #861, opened 2025-05-06
fix: display inline ignores, if feature flag is set

bastiandoetsch:
### Description

### Checklist

- [ ] Linted

### DIFF
--- a/infrastructure/code/code_html.go
+++ b/infrastructure/code/code_html.go
@@ -109,7 +109,7 @@ func GetHTMLRenderer(c *config.Config, apiClient snyk_api.SnykApiClient) (*HtmlR
 
 	status, err := apiClient.FeatureFlagStatus("snykCodeInlineIgnore")
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("Failed to retrieve feature flag status: %w", err)
 	}
 	codeRenderer.inlineIgnoresEnabled = status.Ok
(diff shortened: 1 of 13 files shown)


infrastructure/code/code_html.go
```diff
@@ -216,6 +230,20 @@ func (renderer *HtmlRenderer) GetDetailsHtml(issue types.Issue) string {
 	return result
 }
 
+func (renderer *HtmlRenderer) updateFeatureFlags() {
+	conf := renderer.c.Engine().GetConfiguration()
+	logger := renderer.c.Logger().With().Str("method", "updateFeatureFlags").Logger()
+	renderer.iawEnabled = conf.GetBool(configuration.FF_IAW_ENABLED)
+	ffInlineIgnores := "snykCodeInlineIgnore"
+	status, err := renderer.snykApiClient.FeatureFlagStatus(snyk_api.FeatureFlagType(ffInlineIgnores))
```
rrama, line 238: Any reason we can't have this FF be fetched using the GAF API?
I know it would require registering it somewhere (like LS init), as the ability to fetch feature flags directly without registering them is still a WIP.

rrama (approved):
I would prefer if we got the FF from GAF, as the LS API client is going to be removed. But if you don't fancy doing the change now, this is fine and isn't the worst to rip out later.

PR #863, opened 2025-05-07
fix: AI fix applying to the wrong lines [IDE-1096]

rrama:
### Description

Also added more error state checking.
Also added more tests for some more complex use cases.
Added support for edits at the end of the file, including files without a LF at the EOF.

### Checklist

- [x] Linted
 - Had to split the function to stay under the complexity limit.
 - N/A
 - N/A


domain/ide/command/code_fix_apply_edit.go
```diff
@@ -95,14 +95,17 @@ func (cmd *applyAiFixEditCommand) Execute(ctx context.Context) (any, error) {
 	return nil, nil
 }
 
-func (cmd *applyAiFixEditCommand) getWorkspaceEdit(htmlRenderer *code.HtmlRenderer, fixId string) (types.WorkspaceEdit, error) {
+func (cmd *applyAiFixEditCommand) getWorkspaceEdit(htmlRenderer *code.HtmlRenderer, fixId string) (*types.WorkspaceEdit, error) {
 	path, diff, err := htmlRenderer.AiFixHandler.GetResults(fixId)
 	if err != nil {
 		cmd.logger.Debug().Str("method", "applyAiFixEditCommand.Execute").Msgf("Unable to get the fix fix for %s", fixId)
```
bastiandoetsch, line 101: why is this a debug level message?

infrastructure/code/convert.go
```diff
@@ -562,23 +562,45 @@ func (s *SarifConverter) getMarkers(r codeClientSarif.Result, baseDir types.File
...
 
-	var textEdits []types.TextEdit
+	// Diffs will always use \n instead of \r\n, so no need to sanitize (see getUnifiedDiff).
+	diffLines := strings.Split(diff, "\n")
+	// Remove new line at EOF, if it exists.
+	if n := len(diffLines); n > 0 && diffLines[n-1] == "" {
+		diffLines = diffLines[:n-1]
+	}
+	if len(diffLines) == 0 {
+		return nil, fmt.Errorf("diff is empty")
```
bastiandoetsch, line 582: Why is this an error? IMHO, it's ok to log a warning and just not send an applyEdit. The added error handling should not cause errors, when not needed.

infrastructure/code/convert.go
```diff
@@ -562,23 +562,45 @@ func (s *SarifConverter) getMarkers(r codeClientSarif.Result, baseDir types.File
...
-	for _, line := range strings.Split(diff, "\n") {
-		// If we are being asked to make changes outside the original file, abort and return an empty edit.
-		if hunkLine-hunkOffset > len(fileContentLineStrings) {
-			return edit
+func processLines(diffLines []string, fileContentLineStrings []string) ([]types.TextEdit, error) {
+	var textEdits []types.TextEdit
+	var currentSourceFileLine = 0 // 0-indexed line number counter for our current position in the original file.
+	for _, line := range diffLines {
+		if currentSourceFileLine > len(fileContentLineStrings) {
+			return nil, fmt.Errorf("diff line trying to insert outside of the original file bounds (%d) to line %d: %s", len(fileContentLineStrings), currentSourceFileLine, line)
```
bastiandoetsch, line 603: Why would that be an error? Code could be added to the end of the file, right?

infrastructure/code/convert.go
bastiandoetsch, line 1: This manual parsing/converting of unified diffs still feels wrong. Is there no lib, e.g. from gopls, that converts diffs to text edits?

infrastructure/code/convert.go
```diff
@@ -587,38 +609,49 @@ func CreateWorkspaceEditFromDiff(absoluteFilePath string, diff string) (edit typ
...
+				if err != nil {
+					return nil, err
+				}
+				textEdits = append(textEdits, *textEdit)
+			}
+			// A new insertion does not exist in the original file, so don't increment the counter.
+		} else if strings.HasPrefix(line, " ") { // Context line
+			currentSourceFileLine += 1 // Still exists in the original file.
+		} else {
+			return nil, fmt.Errorf("diff line badly formatted: %s", line)
```
bastiandoetsch, line 640: I'd change that message to `unexpected prefix in line %s`

infrastructure/code/convert.go
rrama: Ah, the error message isn't clear...
If you add to the end of a file, it would correctly not hit this error, as it would still be in an edit in the bounds of the original file, as you would be telling the IDE the following "On the last line of the file, insert the text: `"  a line\n  another line\n  yet another line"`".
I should probably add a test case for this, as I imagine this is now a bug that I have inadvertently fixed. I will also check for inserting at the end with no new line ending at the end of file as well, as that may still be broken.

infrastructure/code/convert.go
rrama: There are a few libraries to convert diffs to nice structs in Go. E.g. [sourcegraph/go-diff](https://github.com/sourcegraph/go-diff)
There is a library to convert from its internal representation of a diff to a LSP `TextEdit`s. E.g. [gopls's `EditsFromDiffEdits`](https://pkg.go.dev/golang.org/x/tools/gopls/internal/protocol#EditsFromDiffEdits)
There is no library that provides both. (gopls requires [giving it a before and after file to generate a diff](https://pkg.go.dev/golang.org/x/tools/internal/diff#Strings))
So we would have to write conversion logic from another libraries representation of the diff to gopls's, or apply the diff to the file in memory in the LS and then use gopls fully.

domain/ide/command/code_fix_apply_edit.go
```diff
@@ -95,14 +95,17 @@ func (cmd *applyAiFixEditCommand) Execute(ctx context.Context) (any, error) {
...
-		return types.WorkspaceEdit{}, err
+		return nil, err
 	}
 
-	workspaceEdit := code.CreateWorkspaceEditFromDiff(path, diff)
+	workspaceEdit, err := code.CreateWorkspaceEditFromDiff(path, diff)
+	if err != nil {
+		return nil, fmt.Errorf("unable to create WorkspaceEdit for %s: %w", path, err)
```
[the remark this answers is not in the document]
rrama: It is just returned all the way up the stack to fail the `snyk.code.fixApplyEdit` command. So it will be shown to the user and I imagine put in the IDE logs.

infrastructure/code/convert.go
rrama: Without this error, from the user's perspective, they would press "Apply fix" and then nothing would happen. We never expect the diff to be empty, but better safe than sorry.

bastiandoetsch: In this case, instead of showing an error, I'd rather show a warning. I don't think it should error out. It may be unexpected, but it's not necessarily an error.

infrastructure/code/convert.go
bastiandoetsch: Hm. Yeah, sounds right, that that would apply to the last line of the file.

domain/ide/command/code_fix_apply_edit.go
rrama: Done.

infrastructure/code/convert.go
rrama: I have moved this check to just be when creating a TextEdit. It is less strict checking, but it makes the error message a lot nicer and the logic more logical.

bastiandoetsch (approved):

infrastructure/code/convert.go
rrama: Discussed with Basti, and raised IDE-1170 for this.

PR #865, opened 2025-05-08
fix: fix unmanaged scan

acke:
### Description

Make sure parsing of issues does not end prematurely for unmanaged scans. We should still parse scan result even if `targetfilepath` is empty, since unmanaged scans does not have a target file.

Smoketest added for unmanaged scans for repository:
https://github.com/snyk-fixtures/cpp-goof

### Checklist

- [x] Linted


application/server/server_smoke_test.go
```diff
@@ -1023,6 +1023,36 @@ func Test_SmokeSnykCodeDelta_NoNewIssuesFound_JavaGoof(t *testing.T) {
 	assert.Equal(t, 0, len(issueList), "no issues expected, as delta and no new change")
 }
 
+func Test_SmokeScanUnmanaged(t *testing.T) {
+	c := testutil.IntegTest(t)
```
bastiandoetsch, line 1027: Why not tagged as Smoke Test?

application/server/server_smoke_test.go
```diff
@@ -1023,6 +1023,36 @@ func Test_SmokeSnykCodeDelta_NoNewIssuesFound_JavaGoof(t *testing.T) {
...
+	testsupport.NotOnWindows(t, "git clone does not work here. dunno why. ") // FIXME
+	loc, jsonRPCRecorder := setupServer(t, c)
+	c.SetSnykOssEnabled(true)
+	c.SetSnykIacEnabled(false)
+	c.EnableSnykCodeQuality(false)
```
bastiandoetsch, line 1032: Hm. Isn't that the default anyways?

application/server/server_smoke_test.go
```diff
@@ -1023,6 +1023,36 @@ func Test_SmokeSnykCodeDelta_NoNewIssuesFound_JavaGoof(t *testing.T) {
...
+
+	cloneTargetDir, err := storedconfig.SetupCustomTestRepo(t, types.FilePath(t.TempDir()), testsupport.CppGoof, "259ea516a4ec", c.Logger())
+	cloneTargetDirString := string(cloneTargetDir)
+	if err != nil {
+		t.Fatal(err, "Couldn't setup test repo")
+	}
+
+	initParams := prepareInitParams(t, cloneTargetDir, c)
+	folderConfig := c.FolderConfig(cloneTargetDir)
+	folderConfig.AdditionalParameters = []string{"--unmanaged"}
```
bastiandoetsch, line 1044: this may be flaky. the right way would be to add it to the stored config in the target directory, I guess.

application/server/server_smoke_test.go
```diff
@@ -1023,6 +1023,36 @@ func Test_SmokeSnykCodeDelta_NoNewIssuesFound_JavaGoof(t *testing.T) {
 	assert.Equal(t, 0, len(issueList), "no issues expected, as delta and no new change")
 }
 
+func Test_SmokeScanUnmanaged(t *testing.T) {
+	c := testutil.IntegTest(t)
+	testsupport.NotOnWindows(t, "git clone does not work here. dunno why. ") // FIXME
```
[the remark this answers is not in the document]
acke: The smoke test will fail when running for windows. We have a few other smoke tests with the same clone git issue I think.

application/server/server_smoke_test.go
acke: Removed these, but kept the Iac, since I need to disable that one specifically.

infrastructure/oss/cli_scanner.go
```diff
@@ -337,12 +337,15 @@ func (cliScanner *CLIScanner) unmarshallAndRetrieveAnalysis(ctx context.Context,
...
 	for _, scanResult := range scanResults {
 		targetFilePath := getAbsTargetFilePath(cliScanner.config, scanResult, workDir, path)
-		fileContent, err := os.ReadFile(string(targetFilePath))
-		if err != nil {
-			reportedErr := fmt.Errorf("skipping scanResult for path: %s displayTargetFile: %s in workDir: %s as we can't determine the absolute filesystem path. %w", scanResult.Path, scanResult.DisplayTargetFile, workDir, err)
-			cliScanner.errorReporter.CaptureErrorAndReportAsIssue(targetFilePath, reportedErr)
-			logger.Error().Err(reportedErr).Send()
-			continue
+		var fileContent []byte
+		if targetFilePath != "" && scanResult.PackageManager != packageManagerUnmanaged {
```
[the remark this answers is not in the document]
acke: The `getAbsTargetFilePath` returns the `path` if `DisplayTargetFile` is empty. When we run smoketests, the `targetFilePath` will never be empty. 

```
	if scanResult.DisplayTargetFile == "" && path != "" {
		return path
	}
```

Also `updateGlobalCacheAndSeverityCounts` assumes the results are connected to a path and panics if not.

```
		if !f.Contains(issue.GetAffectedFilePath()) {
			msg := "issue found in scanData " + issue.GetAffectedFilePath() + " that does not pertain to folder: " + f.path
			panic(msg)
		}

```
The code flow works is we allow `getAbsTargetFilePath` to return path, but check for packageManager in cli_scanner.

Do we want to rewrite the scan flow, or do a check for packageManager?

PR #868, opened 2025-05-12
fix: test for unmanaged scans as smoke test [IDE-850]

acke:
### Description

Cleanup smoke test for unmanaged scans and make it run as a smoke test.

### Checklist

- [x] Linted

### DIFF
--- a/application/server/server_smoke_test.go
+++ b/application/server/server_smoke_test.go
@@ -1024,12 +1024,10 @@ func Test_SmokeSnykCodeDelta_NoNewIssuesFound_JavaGoof(t *testing.T) {
 }
 
 func Test_SmokeScanUnmanaged(t *testing.T) {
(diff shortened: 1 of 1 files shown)


application/server/server_smoke_test.go
```diff
@@ -1024,12 +1024,10 @@ func Test_SmokeSnykCodeDelta_NoNewIssuesFound_JavaGoof(t *testing.T) {
 }
 
 func Test_SmokeScanUnmanaged(t *testing.T) {
-	c := testutil.IntegTest(t)
 	testsupport.NotOnWindows(t, "git clone does not work here. dunno why. ") // FIXME
```
ShawkyZ, line 1027: is it failing on Windows ?

bastiandoetsch (approved):

PR #869, opened 2025-05-13
fix: remove packageManager check [IDE-1174]

acke:
### Description

The pull request removes the packageManager check in cli_scanner.go and the packageManagerUnmanaged constant in types.go. 

### Checklist

- [x] Linted

### DIFF
--- a/infrastructure/oss/cli_scanner.go
+++ b/infrastructure/oss/cli_scanner.go
@@ -343,7 +343,7 @@ func (cliScanner *CLIScanner) unmarshallAndRetrieveAnalysis(ctx context.Context,
 		targetFilePath := getAbsTargetFilePath(cliScanner.config, scanResult, workDir, path)
 		var fileContent []byte
 
-		if targetFilePath != "" && uri.IsReadableFile(targetFilePath) {
+		if targetFilePath != "" && uri.IsRegularFile(targetFilePath) {
 			fileContent, err = os.ReadFile(string(targetFilePath))
 			if err != nil {
 				reportedErr := fmt.Errorf("skipping scanResult for path: %s displayTargetFile: %s in workDir: %s as we can't determine the absolute filesystem path. %w", scanResult.Path, scanResult.DisplayTargetFile, workDir, err)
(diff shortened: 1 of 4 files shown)


infrastructure/oss/cli_scanner.go
```diff
@@ -338,7 +338,8 @@ func (cliScanner *CLIScanner) unmarshallAndRetrieveAnalysis(ctx context.Context,
 	for _, scanResult := range scanResults {
 		targetFilePath := getAbsTargetFilePath(cliScanner.config, scanResult, workDir, path)
 		var fileContent []byte
-		if targetFilePath != "" && scanResult.PackageManager != packageManagerUnmanaged {
+
+		if targetFilePath != "" && !uri.IsDirectory(targetFilePath) {
```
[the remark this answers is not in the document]
ShawkyZ: I think uri.IsDirectrory would return false if path doesnt exist (not sure though)
I think using `.Stat()` and checking file access mode would be more accurate

internal/uri/uri_util.go
```diff
@@ -136,6 +136,11 @@ func IsDirectory(path types.FilePath) bool {
 	return stat.IsDir()
 }
 
+func IsReadableFile(path types.FilePath) bool {
```
rrama, line 139: Function doesn't actually check if the file is readable, just that it exists. If we don't need to actually check if it is readable or not, then just call this `IsRegularFile`, otherwise you will need to update the function to check if the file is readable.
I don't know the answer, but I imagine without using a Go library it's difficult to determine if a file is readable without attempting to open it, so maybe we should let the caller deal with that.

internal/uri/uri_util_test.go
```diff
@@ -264,3 +264,32 @@ func getTestRange() Range {
 		EndChar:   10,
 	}
 }
+
+func TestIsReadableFile(t *testing.T) {
```
rrama, line 268: Misnamed.

rrama (approved):
I think I get it. For unmanaged we get back a path pointing to the directory of the project from `getAbsTargetFilePath()`, which is useless to us, so we need to distinguish if this is the case.
My only concern was `targetFilePath` being passed into `cliScanner.retrieveIssues()`, but it seems like we do some checks it that it never matches and then don't use it. (Technically if the user named their directory "blah.html", "blah.htm" or "pom.xml", it could go down the wrong path, but that is a risk I am willing to take)

PR #871, opened 2025-05-14
fix(uri): implement case-insensitive path checking with in-memory cache [IDE-1198]

bastiandoetsch:
### Description

This PR addresses an issue where workspace folder trust validation
fails on case-insensitive filesystems (Windows and macOS) when comparing paths
with different letter casing.

Key changes:
- Added robust case sensitivity detection logic for different filesystems:
    - Windows is always treated as case-insensitive
    - macOS is tested by creating temporary files with different case and
      comparing inode numbers
    - Linux and other systems default to case-sensitive behavior
- Implemented an in-memory cache for filesystem case sensitivity results to
  avoid redundant filesystem operations
- Used filesystem root/volume as cache key to correctly handle different
  mounted volumes with potentially different case sensitivity settings
- Fixed several edge cases with proper error handling and fallbacks
- Added comprehensive test cases to verify behavior across different OSes
- Ensured compatibility with IDE file scheme handling

This improvement ensures that workspace folder trust validation works
correctly across all major operating systems, regardless of filesystem
case sensitivity configuration.

### Checklist

- [x] Linted

### DIFF
--- a/internal/uri/uri_util_test.go
+++ b/internal/uri/uri_util_test.go
@@ -17,6 +17,7 @@
 package uri
 
 import (
+	"errors"
 	"os"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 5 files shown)


internal/uri/uri_util_test.go
```diff
@@ -77,12 +76,13 @@ func TestFolderContains(t *testing.T) {
 		assert.True(t, FolderContains("C:\\folder\\", "C:\\folder"))
 		assert.True(t, FolderContains("C:\\folder", "C:\\folder\\"))
 		assert.True(t, FolderContains("C:\\folder\\", "C:\\folder\\"))
+		assert.True(t, FolderContains("C:\\Folder\\", "C:\\folder\\file"))
+		assert.True(t, FolderContains("C:\\folder\\", "C:\\FOLDER\\file"))
+		assert.True(t, FolderContains("C:\\FOLDER\\", "C:\\folder\\SUBFOLDER\\file"))
```
acke, line 81: Do we want to add test cases for alternative supported windows paths, such as
\\Server\Share\Folder\File.txt
\\?\C:\Very\Long\Path\File.txt
\Folder\File.txt

https://www.perplexity.ai/search/what-types-of-file-path-does-w-IEmziHcKQ.2zpJE431bfcA

acke (approved):
Small nitpick, looks good!

internal/uri/uri_util_test.go
bastiandoetsch: Hm. Not sure if UNC paths should be supported here. But good point.

PR #872, opened 2025-05-15
fix: html encode ResourcePath

acke:
### Description

### Checklist

- [ ] Linted


infrastructure/iac/iac_html.go
```diff
@@ -41,7 +42,7 @@ type TemplateData struct {
 	SeverityIcon template.HTML
 	Description  template.HTML
 	Remediation  template.HTML
-	ResourcePath template.HTML
+	ResourcePath string
```
bastiandoetsch, line 45: Why is this necessary?

infrastructure/iac/iac_html_test.go
```diff
@@ -35,8 +35,8 @@ func Test_IaC_Html_getIacHtml(t *testing.T) {
 		assert.Contains(t, iacPanelHtml, `/Users/cata/git/playground/dex/examples/k8s/dex.yaml`, "HTML should contain file path for the ignore functionality")
 	}
 
-	// ResourcePath is correctly formatted
-	assert.Contains(t, iacPanelHtml, "[DocId: 5] > rules[0] > verbs", "HTML should contain the path to the affected file")
+	// ResourcePath is correctly HTML encoded
```
bastiandoetsch, line 38: Could you maybe add a test case that demonstrates that JavaScript injections are prevented?

acke: Done!

infrastructure/iac/iac_html.go
acke: Perplexity stated that

`Do not use template.HTML on unescaped or untrusted input, as it disables Go's auto-escaping and can reintroduce XSS vulnerabilities.`

But it might be okay to use `template.HTML` after our call to `html.EscapeString`.

Ill add it back, to make it more consistent with the other items in the data struct.

infrastructure/iac/iac_html.go
```diff
@@ -112,6 +113,6 @@ func (service *HtmlRenderer) GetDetailsHtml(issue types.Issue) string {
 	return htmlTemplate.String()
 }
 
-func formatPath(path []string) template.HTML {
-	return template.HTML(strings.Join(path, " > "))
+func formatPath(path []string) string {
+	return gohtml.EscapeString(strings.Join(path, " > ")) // Strict HTML encoding, prevents XSS attacks
```
rrama, line 117: Why do we need to escape it if you are no longer returning it as `template.HTML`?

### Old code breakdown:
1. `formatPath()` is called and the `path` segments are joined together.
2. `formatPath()` casts these joined up segments from a `string` to `template.HTML`.
3. The `template.HTML` is stored in `ResourcePath` (still as `template.HTML`) in `data`.
4. `service.globalTemplate.Execute(&htmlTemplate, data)` sees that the value of `ResourcePath` is `template.HTML` and injects it into the HTML template unmodified.
   - `template.HTML` is effectively saying "Don't touch this, it's already safe HTML, trust me, I promise."

N.b. There has been zero actual escaping of the user data (`path` > `ResourcePath`).

### Your new code breakdown:
1. `formatPath()` is called, and the `path` segments are joined together into a regular string.
2. `formatPath()` then explicitly escapes this joined string using `gohtml.EscapeString()`. This escapes the user data.
3. This HTML-escaped string is stored in `ResourcePath` (still as `string`) in `data`.
4. `service.globalTemplate.Execute(&htmlTemplate, data)` sees that the value of `ResourcePath` is a standard `string` and performs HTML escaping on the content before injecting it. This is the second escaping.
   - `string` is effectively saying "I don't know what's in this content, so please make sure it is safe."
N.b. The user data (`path` > `ResourcePath`) was escaped twice.

N.b. There has been zero actual escaping of the user data (`path` > `ResourcePath`).

infrastructure/iac/iac_html_test.go
```diff
@@ -35,8 +36,8 @@ func Test_IaC_Html_getIacHtml(t *testing.T) {
...
+	assert.Contains(t, iacPanelHtml, "[DocId: 5] &amp;gt; rules[0] &amp;gt; verbs", "HTML should contain the path to the affected file")
```
rrama, line 40: As mentioned above, this is double escaped. I would expect the function to only do a single escape.
I.e. instead of `&amp;gt;` it would return it as `&gt;`

infrastructure/iac/iac_html_test.go
```diff
@@ -91,3 +92,31 @@ func createIacIssueSample() snyk.Issue {
...
+		AdditionalData: snyk.IaCIssueData{
+			Path: maliciousPath,
+		},
+	}
+
+	// Get the HTML details
+	htmlDetails := renderer.GetDetailsHtml(&issue)
+
+	// Assert that the malicious script is HTML encoded in the ResourcePath
+	expectedEncodedPath := "&amp;lt;script nonce=&amp;#34;${nonce${headerEnd}}&amp;#34;&amp;gt;alert(1)&amp;lt;/script&amp;gt;"
```
rrama, line 117: Again, this is double escaped.

rrama (changes requested):
Your code ends up double escaping the user data.

infrastructure/iac/iac_html.go
rrama: I am going to disagree with the AI here. We end up with double escaping if you don't. Although I think the better to just make `ResourcePath`a `string` and not use `html.EscapeString`.

rrama (approved):
LGTM!
For the security issue this is addressing, we should verify with Cristina-Elena that it is fixed after this change.
We should also take this opportunity having just done this one, to raise and work on a task to check all other data template variables across the various HTML generator files for similar issues.

PR #886, opened 2025-06-02
feat: add pat support

andrewrobinsonhodges-snyk:
### Description

### Checklist

- [ ] Linted

### DIFF
--- a/application/config/config.go
+++ b/application/config/config.go
@@ -1107,6 +1107,24 @@ func (c *Config) Logger() *zerolog.Logger {
 	return c.logger
 }
 
+func (c *Config) AuthenticationMethodForCurrentToken() types.AuthenticationMethod {
+	token := c.token
+
+	if len(token) == 0 {
+		return types.EmptyAuthenticationMethod
+	}
+
+	if auth.IsAuthTypePAT(token) {
+		return types.PatAuthentication
+	}
+
+	if auth.IsAuthTypeToken(token) {
+		return types.TokenAuthentication
+	}
+
+	return types.OAuthAuthentication
+}
+
 func (c *Config) TokenAsOAuthToken() (oauth2.Token, error) {
 	token := c.Token()
 
--- a/infrastructure/authentication/auth_configuration.go
+++ b/infrastructure/authentication/auth_configuration.go
@@ -32,6 +32,22 @@ import (
 
 // Token authentication configures token only authentication
 func Token(c *config.Config, errorReporter error_reporting.ErrorReporter) AuthenticationProvider {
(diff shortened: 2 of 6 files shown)


infrastructure/authentication/auth_configuration.go
```diff
@@ -32,14 +32,29 @@ import (
...
+	unsetOauthTokenConfig(c)
+	return NewCliAuthenticationProvider(c, errorReporter)
+}
+
+func Pat(c *config.Config, authenticationService AuthenticationService) AuthenticationProvider {
+	unsetOauthTokenConfig(c)
+
+	openBrowserFunc := func(url string) {
+		authenticationService.provider().setAuthUrl(url)
+		types.DefaultOpenBrowserFunc(url)
```
ShawkyZ, line 44: just to double check, will this pipe StdOut to StdErr ?

application/config/config.go
```diff
@@ -1077,6 +1074,28 @@ func (c *Config) Logger() *zerolog.Logger {
...
+	}
+
+	var derivedMethod types.AuthenticationMethod
+	if len(token) == 0 {
+		derivedMethod = types.EmptyAuthenticationMethod
+	} else if auth.IsAuthTypePAT(token) {
+		derivedMethod = types.PatAuthentication
+	} else if auth.IsAuthTypeToken(token) {
+		derivedMethod = types.TokenAuthentication
+	} else {
```
ShawkyZ, line 1092: nitpick: we can also detect OAuth tokens, easiest would be to try parse JSON.

infrastructure/authentication/pat_provider.go
```diff
@@ -0,0 +1,78 @@
...
+	p.logger.Debug().Msg("clearing authentication")
+	p.config.Unset(auth.CONFIG_KEY_OAUTH_TOKEN)
+	p.config.Unset(configuration.AUTHENTICATION_TOKEN)
+	p.config.Unset(configuration.AUTHENTICATION_BEARER_TOKEN)
+	return nil
+}
+
+func (p *PatAuthenticationProvider) AuthURL(_ context.Context) string {
+	// no lock should be used here, as this is usually called during authentication flow, which write-locks the mutex
+	return p.authURL
```
ShawkyZ, line 77: this noop, or maybe line 53 should use this url ?

rrama (changes requested):
Must revert the breaking change, the rest is all super minor.

It seems blanking of the other auth provider's tokens is already quite a messy process, so if it can't be cleaned up now easily (timeboxed), we should probably raise a follow on task to do it while auth is still fresh in our minds.

PR #890, opened 2025-06-04
fix: add snyk-ls- prefix to x-snyk-ide header [IDE-1244]

ShawkyZ:
### Description

Add `snyk-ls-` prefix to `x-snyk-ide` header. 
Backend requires this prefix 

Also add X-Snyk-CLI-Version header back, since current version checking in registry will break for LS  and it will prioritize checking for the CLI header, if it exists first

This is a temp workaround until the backend logic is fixed. Also this change basically reverts the headers change to how it was before.

### Checklist

- [ ] Linted

### DIFF
--- a/application/server/server.go
+++ b/application/server/server.go
@@ -223,7 +223,7 @@ func initNetworkAccessHeaders() {
 	gafConfig := engine.GetConfiguration()
 	ua := util.GetUserAgent(gafConfig, config.Version)
 	engine.GetNetworkAccess().RemoveHeaderField("x-snyk-cli-version")
-	engine.GetNetworkAccess().AddHeaderField("x-snyk-ide", ua.AppVersion)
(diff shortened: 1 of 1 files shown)


bastiandoetsch (approved):

bastiandoetsch:
Please also double-check against the backend verification logic if this is enough.

application/server/server.go
```diff
@@ -222,8 +222,8 @@ func initNetworkAccessHeaders() {
 	engine := config.CurrentConfig().Engine()
```
bastiandoetsch, line 222: could we maybe improve this to inject the config via parameter?

bastiandoetsch (approved):

application/server/server.go
ShawkyZ: was done in another PR

PR #903, opened 2025-06-18
chore: improve MCP tools description

ShawkyZ:
### Description
* Improve MCP tools description
* Log err only if CLI exit code > 1

### Checklist

- [ ] Linted


mcp_extension/scan_tool.go
```diff
@@ -125,7 +126,15 @@ func (m *McpLLMBinding) runSnyk(ctx context.Context, invocationCtx workflow.Invo
...
 	if err != nil {
-		logger.Err(err).Msg("Received error running command")
+		var errorType *exec.ExitError
+		if errors.As(err, &errorType) {
+			if errorType.ExitCode() > 1 {
+				// Exit code > 1 means CLI run didn't work
+				logger.Err(err).Msg("Received error running command")
+			}
+		} else {
+			logger.Err(err).Msg("Received error running command")
```
rrama, line 136: But the log line is the same...

mcp_extension/snyk_tools.json
```diff
@@ -1032,7 +1032,7 @@
...
+      "description": "Modifies the project's `.snyk` policy file to ignore specific issues (vulnerabilities, license issues via `id`) or to exclude files/directories from analysis (primarily for Snyk Code and Open Source `unmanaged` scans via `file_path`). \nWhen to use: For risk acceptance (documenting why a vulnerability is acceptable). For Snyk Code scope management (excluding test/generated code). \nHow to use: Ignore npm vuln with expiry/reason: `<snyk_ignore> `id`=`'npm:tough-cookie:20160722'` `expiry`=`'2024-12-31'` `reason`=`'Not exploitable'`. Ignore by path: <snyk_ignore> `id`=`'SNYK-JS-LIB-XXX'` `vulnerable_path`=`'app>pkg>lib'`. Exclude test files from Snyk Code: <snyk_ignore> `file_path`=`'./**/test/**'` `file_path-group`=`'code'`. Ignore IaC misconfig: <snyk_ignore> `id`=`SNYK-CC-TF-118` `path`=`\"resource > type > name\"`. \nSecure SDLC Integration: Vulnerability Management/Triage: Allows teams to formally document and manage accepted risks or refine scan scope as 'policy as code' within the `.snyk` file, which is version controlled with the project.",
```
rrama, line 1035: Q: Does this tool work with CCIs enabled? I know we wouldn't update the description until CCI is GA though.
I really hope we don't need a tool for checking feature flags...

mcp_extension/snyk_tools.json
```diff
@@ -1152,7 +1152,7 @@
     },
     {
       "name": "snyk_trust",
-      "description": "Trust a given folder to allow Snyk to scan it",
+      "description": "Trust a given folder to allow Snyk to scan it.",
```
rrama, line 1155: Qs:
1. Do we need to add like "This tool will need to be run if a scan fails because the folder isn't trusted." or is the LLM smart enough to figure it out?
1. Does the LLM try to run this every time or just if the scan command fails?

mcp_extension/snyk_tools.json
ShawkyZ: It will write to the .snyk file no FF check is needed.
In the future if a cmd is behind a FF it will just fail, the error need to be then clear enough to instruct the LLM on what to do

mcp_extension/snyk_tools.json
ShawkyZ: If a scan fails because of trust, the error that is returned explicitly says `run snyk_trust`

PR #909, opened 2025-06-24
WIP fix: use Git's branches instead of cached list

rrama:
### Description

The cached list is going to be out of date most of the time.
Was broken in this PR: https://github.com/snyk/snyk-ls/pull/835/files#diff-ffb72e6e3be88f30be0024e4c8fbf8870f3cd6f4ccfdfbb2337f2bbddb437bd3L75-R75

### Checklist

- [x] Linted
 - N/A
 - N/A

### DIFF
--- a/internal/storedconfig/stored_config.go
+++ b/internal/storedconfig/stored_config.go
@@ -38,7 +38,7 @@ func GetOrCreateFolderConfig(conf configuration.Configuration, path types.FilePa
 	}
 
 	localBranches, err := getLocalBranches(repository)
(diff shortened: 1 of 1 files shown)


rrama:
Shawky pointed out that it would be better to:
A) Scrap the config merging, which is out of scope for this fix.
B) Rename the function `getFromGit` to `getFromGitConfig` and pull out the local branches functions so I will do that.

rrama:
Not going to spend any longer on this until this customer support case is raised, as it is not a priority for this release.

rrama:
Replaced by #942

PR #911, opened 2025-06-25
fix: distinguish between provider and token changes when showing errors to users

andrewrobinsonhodges-snyk:
### Checklist

- [ ] Linted


rrama (approved):
LGTM!

PR #918, opened 2025-07-08
fix: send auth analytic when token updated [IDE-1292]

andrewrobinsonhodges-snyk:
### Description

Add analytic when token is updated in LS config.

Note - I've not verified this yet; analytics are only sent when the client is authenticated, and we might be sending this one too early.

### Checklist


acke (approved):
Approved with nitpick.

PR #928, opened 2025-07-21
feat: send analytics on org change [IDE-880]

andrewrobinsonhodges-snyk:
### Description

Simplified Analytics Event Creation - Event creation now supports creating a target ID for a particular file, or for an instance of the Language server. This means that places in the code that send analytics just need to call the helper method to create a new event, then add their specific extension field.

Added analytics for Org changes: Uses the new mechanism to create an analytics event, then populates the extension with the old and new org values. Note that we use the Org UUIDs rather than the slug names.

### Checklist

### DIFF
--- a/infrastructure/analytics/analytics.go
+++ b/infrastructure/analytics/analytics.go
@@ -43,6 +43,8 @@ func NewAnalyticsEventParam(interactionType string, err error, path types.FilePa
 	}
 
 	var targetId string
+	// If we have a file path, we use that to determine the target ID. For analytics such as authentication events, which
+	// are not associated with a file, we use the binary name as the target ID (with a fallback if that fails).
 	if path != "" {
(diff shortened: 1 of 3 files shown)


infrastructure/analytics/analytics.go
```diff
@@ -40,7 +41,18 @@ func NewAnalyticsEventParam(interactionType string, err error, path types.FilePa
 	if err != nil {
 		status = string(analytics.Failure)
 	}
-	targetId, _ := instrumentation.GetTargetId(string(path), instrumentation.AutoDetectedTargetId)
+
+	var targetId string
+	// If we have a file path, we use that to determine the target ID. For analytics such as authentication events, which
+	// are not associated with a file, we use the binary name as the target ID (with a fallback if that fails).
+	if path != "" {
+		targetId, _ = instrumentation.GetTargetId(string(path), instrumentation.AutoDetectedTargetId)
```
bastiandoetsch, line 49: shouldn't the error be handled?

PR #942, opened 2025-07-28
fix: remove Git-stored folder config migration [IDE-1291]

rrama:
### Description

The stored config migration code (from local Git repo config to the new file) had existed for over 6 months, so could be removed.
A bug with the migration code meant we were using the Git branches stored in the config file as opposed to getting them freshly from Git each time, this has been fixed.

### Checklist

 - N/A
 - N/A
 - Removed a dependency.


rrama:
> Ensured that Git branch information (local branches and base branch) is always fetched freshly from the Git repository

Gemini is wrong, base branch is not and should not be freshly fetched from Git unless we don't have the base branch already stored.

internal/storedconfig/git.go
```diff
@@ -117,121 +72,36 @@ func getLocalBranches(repository *git.Repository) ([]string, error) {
...
-}
+func enrichFromGit(logger *zerolog.Logger, folderConfig *types.FolderConfig) {
+	l := logger.With().Str("method", "enrichFromGit").Logger()
 
-func getFromGit(path types.FilePath) (*types.FolderConfig, error) {
-	repository, repoConfig, _, folderSection, err := getConfigSection(path)
+	repository, err := git.PlainOpen(string(folderConfig.FolderPath))
 	if err != nil {
-		return nil, nil
+		return // Probably not a git repo and that's okay
```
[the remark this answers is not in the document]
rrama: Good idea, but scope creep.

internal/storedconfig/stored_config_test.go
```diff
@@ -96,213 +100,97 @@ func Test_GetOrCreateFolderConfig_shouldReturnExistingFolderConfig(t *testing.T)
...
-	require.NoError(t, err)
+	// Base branch should be empty since we couldn't determine it
+	assert.Empty(t, folderConfig.BaseBranch)
+}
 
-	// Create branches that are neither main nor master
-	cmd = exec.Command("git", "checkout", "-b", "feature-branch")
-	cmd.Dir = tempDir
-	err = cmd.Run()
+func Test_GetOrCreateFolderConfig_GitLocalBranchesTakePriorityOverStoredConfig(t *testing.T) {
```
rrama, line 139: New tests below, but the diff is apparently too similar, sorry.

internal/storedconfig/stored_config.go
```diff
@@ -20,73 +20,31 @@ import (
...
 	if err != nil {
-		return folderConfig, nil
+		return nil, err
 	}
 
-	if folderConfig != nil && gitFolderConfig != nil {
-		// the previous git configuration takes precedence
-		folderConfig = mergeFolderConfigs(gitFolderConfig, folderConfig)
-	}
+	enrichFromGit(&l, folderConfig)
```
bastiandoetsch, line 38: Hm. shouldn't that be assigned back? Or is it updating the pointer parameter? In any case, it would be cleaner, to return the enriched value and assign it to `folderConfig`.

internal/vcs/checkout_handler.go
```diff
@@ -65,7 +65,7 @@ func (ch *CheckoutHandler) CheckoutBaseBranch(logger *zerolog.Logger, folderConf
 		return nil
 	}
 
-	baseBranchName := GetBaseBranchName(ch.conf, folderPath)
+	baseBranchName := GetBaseBranchName(ch.conf, folderPath, logger)
```
bastiandoetsch, line 68: I wonder, can't we get the logger from the conf struct? I hate passing loggers around ;)

bastiandoetsch (approved):

bastiandoetsch (approved):
LGTM, good job.

internal/vcs/checkout_handler.go
rrama: I don't think you can get it from the GAF configuration. I imagine you would need to be passing engine about.

bastiandoetsch (approved):

PR #957, opened 2025-08-06
fix: update GAF and Go [IDE-1351] [CLI-965]

rrama:
### Checklist


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/snyk/snyk-ls
 
-go 1.23.8
+go 1.24
```
ShawkyZ, line 3: we need to update GAF to pin the go version.. but after this PR

PR #958, opened 2025-08-06
fix: wip determine env and PATH

ShawkyZ:
### Description

GAF change as well:
```go
// LoadConfiguredEnvironment updates the environment with local configuration. Precedence as follows:
//  1. std folder-based config files
//  2. given command-line parameter config file
//  3. std config file in home directory
//  4. global shell configuration
//
// VVV NEW COMMENT HERE VVV
// Any PATH will be appended to the current PATH.
// ^^^ NEW COMMENT HERE ^^^
func LoadConfiguredEnvironment(customConfigFiles []string, workingDirectory string) {
	bashOutput := getEnvFromShell("bash")

	// this is applied at the end always, as it does not overwrite existing variables
	defer func() { _ = gotenv.Apply(strings.NewReader(bashOutput)) }() //nolint:errcheck // we can't do anything with the error

	env := gotenv.Parse(strings.NewReader(bashOutput))
	specificShell, ok := env["SHELL"]
	if ok {
		fromSpecificShell := getEnvFromShell(specificShell)
		_ = gotenv.Apply(strings.NewReader(fromSpecificShell)) //nolint:errcheck // we can't do anything with the error

		// VVV NEW CODE HERE VVV
		env = gotenv.Parse(strings.NewReader(fromSpecificShell))
		if val, ok := env["PATH"]; ok {
			UpdatePath(val, false)
		}
		// ^^^ NEW CODE HERE ^^^
	}

	// process config files
	for _, file := range customConfigFiles {
		if !filepath.IsAbs(file) {
			file = filepath.Join(workingDirectory, file)
		}
		loadFile(file)
	}
}
```

### Checklist

### DIFF
--- a/application/config/config.go
+++ b/application/config/config.go
@@ -176,6 +176,7 @@ type Config struct {
 	defaultDirs                      []string
 	automaticAuthentication          bool
 	tokenChangeChannels              []chan string
+	prepareDefaultEnvChannel         chan bool
 	filterSeverity                   types.SeverityFilter
 	issueViewOptions                 types.IssueViewOptions
 	trustedFolders                   []types.FilePath
@@ -335,12 +336,8 @@ func getNewScrubbingLogger(c *Config) *zerolog.Logger {
 	return &logger
 }
 
-func (c *Config) AddBinaryLocationsToPath(searchDirectories []string) {
-	c.m.Lock()
+func (c *Config) addBinaryLocationsToPath(searchDirectories []string) {
 	c.defaultDirs = searchDirectories
-	c.m.Unlock()
-	c.determineJavaHome()
-	c.mavenDefaults()
 }
 
 func (c *Config) determineDeviceId() string {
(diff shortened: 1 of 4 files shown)


application/config/config.go
```diff
@@ -890,14 +891,33 @@ func (c *Config) SetAutomaticScanning(value bool) {
...
-	if //goland:noinspection GoBoolExpressions
-	runtime.GOOS != windows {
-		envvars.UpdatePath("/usr/local/bin", false)
-		envvars.UpdatePath("/bin", false)
-		envvars.UpdatePath(xdg.Home+"/bin", false)
-	}
-	c.determineJavaHome()
-	c.mavenDefaults()
+	go func() {
+		envvars.LoadConfiguredEnvironment(nil, "")
```
ShawkyZ, line 895: this still needs a change in GAF to always append PATH

rrama:
Closing in favour of: https://github.com/snyk/snyk-ls/pull/969

PR #962, opened 2025-08-08
chore(ci): check mocks were generated

rrama:
### Description

A combined step for lint and generate, to ensure `make generate` isn't missed.
Regen the missed mock.
Stop using global `mockgen`, now use the `go tool` command to have a project specified version of `mockgen`.
Update `mockgen` commands to ensure the mocks are formatted properly to linter standards.

### Expected FAQs

Why change how we run `mockgen`?

- Because we previously used whatever version of `mockgen` that happened to be on the `PATH`. We have other projects which were using different versions, which were also installing `mockgen` to the exact same place, so people were regenerating the mocks using outdated or newer versions of the command.

Why this approach to tools?

- It is the new Go standard as of 1.24 https://go.dev/doc/go1.24#tools

Why change the command format?

- I could have just ignored the mocks from linting, but where is the fun in that?

### Checklist

 - N/A
 - Yes and now it is enforced :)
 - N/A
 - N/A
 - N/A


.github/workflows/build.yaml
```diff
@@ -18,8 +18,8 @@ on:
   pull_request:
 
 jobs:
-  lint:
-    name: lint
+  lint-and-generate:
```
bastiandoetsch, line 21: no need to rename. this is just another type of linting.

Makefile
```diff
@@ -103,7 +99,7 @@ instance-test:
 
 ## generate: Regenerate generated files (e.g. mocks).
 .PHONY: generate
-generate: tools
```
bastiandoetsch, line 106: why?

rrama:
I have now seen that in Go 1.24 they have added a new way to do tools in the go.mod without a tools.go. https://go.dev/doc/go1.24#tools So I will use that instead.

bastiandoetsch (approved):

PR #969, opened 2025-08-19
fix: rework environment fetching [IDE-1314]

rrama:
### Description

Move initial environment setup to one place that runs in a Goroutine to not block, then wait on the env before scanning to prevent a race condition.
Fix a race condition where the LS settings `Path` was causing a cache of the `PATH` before the environment was fully set up, then later restoring to this old value.

Note: Splitting the reading of user shell env and reading project env files to different places was scrapped, so we're staying with `LoadConfiguredEnvironment`, but still adding calls to `WaitForDefaultEnv` beforehand.
Note: Environment snapshotting for each workspace before reading the env files was scrapped last minute, as I realised it would prevent parallel project scanning. Plus, the ordering of whether the IDE SDK paths are prepended before or after the env files are read is a bit random. This being said, it is no worse than it was before the changes (probably), I had just hoped to make it a lot better all at once.

Added an option to `config.New()` to allow tests to create a config without having to search all the directories for binaries.
Also, removed `Config.Path()` as it was only used by tests and `os.Getenv("PATH")` is more reliable.

### Checklist

 - N/A
 - N/A
 - N/A

### DIFF
--- a/application/config/config.go
+++ b/application/config/config.go
@@ -17,6 +17,7 @@
 package config
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -245,6 +246,7 @@ func newConfig(engine workflow.Engine) *Config {
 	c := &Config{}
 	c.logger = getNewScrubbingLogger(c)
 	c.cliSettings = NewCliSettings(c)
+	c.prepareDefaultEnvChannel = make(chan bool, 1)
 	c.automaticAuthentication = true
 	c.configFile = ""
 	c.format = FormatMd
@@ -498,8 +500,15 @@ func (c *Config) TokenChangesChannel() <-chan string {
 	return channel
 }
 
-func (c *Config) PrepareDefaultEnvChannel() <-chan bool {
-	return c.prepareDefaultEnvChannel
+// WaitForDefaultEnv blocks until the default environment has been prepared
+// or until the provided context is canceled. Returns the context error if it is done.
+func (c *Config) WaitForDefaultEnv(ctx context.Context) error {
+	select {
+	case <-c.prepareDefaultEnvChannel:
+		return nil
+	case <-ctx.Done():
+		return ctx.Err()
+	}
 }
 
 func (c *Config) SetCliSettings(settings *CliSettings) {
@@ -892,7 +901,7 @@ func (c *Config) SetAutomaticScanning(value bool) {
 
 func (c *Config) addDefaults() {
 	go func() {
-		envvars.LoadConfiguredEnvironment(nil, "")
+		envvars.LoadShellEnvironment()
 		//goland:noinspection GoBoolExpressions
 		if runtime.GOOS == "windows" {
 			c.defaultDirs = []string{
@@ -916,7 +925,7 @@ func (c *Config) addDefaults() {
 		}
 		c.determineJavaHome()
 		c.mavenDefaults()
-		c.prepareDefaultEnvChannel <- true
+		close(c.prepareDefaultEnvChannel)
 	}()
 }
 
--- a/application/config/config.go
+++ b/application/config/config.go
@@ -898,6 +898,7 @@ func (c *Config) SetAutomaticScanning(value bool) {
 
 func (c *Config) addDefaults() {
 	go func() {
+		defer func() { close(c.prepareDefaultEnvChannel) }()
 		envvars.LoadShellEnvironment()
 		//goland:noinspection GoBoolExpressions
 		if runtime.GOOS != "windows" {
@@ -907,7 +908,6 @@ func (c *Config) addDefaults() {
 		}
 		c.determineJavaHome()
 		c.mavenDefaults()
-		close(c.prepareDefaultEnvChannel)
 	}()
 }
 
--- a/application/config/config.go
+++ b/application/config/config.go
@@ -205,6 +205,7 @@ type Config struct {
 	isLSPInitialized                 bool
 	snykAgentFixEnabled              bool
 	cachedOriginalPath               string
+	userSettingsPath                 string
 }
 
 func CurrentConfig() *Config {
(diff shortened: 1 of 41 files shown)


infrastructure/oss/cli_scanner.go
```diff
@@ -207,6 +207,7 @@ func (cliScanner *CLIScanner) scanInternal(ctx context.Context, path types.FileP
 	cliScanner.mutex.Unlock()
 
 	cmd := commandFunc([]string{string(workDir)}, map[string]bool{"": true}, workDir, folderConfig)
+
```
rrama, line 210: Remnants of an undo, will revert.

internal/testutil/test_setup.go
```diff
@@ -128,6 +132,10 @@ func prepareTestHelper(t *testing.T, envVar string, useConsistentIgnores bool) *
 	}
 
 	c := config.New()
```
rrama, line 134: Consider setting an empty binary search path for integration tests. Otherwise we will be slowing down the Windows test massively. Will need to ensure it doesn't break anything if we do though though...

application/config/config.go
```diff
@@ -231,19 +232,28 @@ func IsDevelopment() bool {
...
+func newConfig(engine workflow.Engine, opts ...ConfigOption) *Config {
 	c := &Config{}
+
+	for _, opt := range opts {
+		opt(c)
+	}
+
 	c.logger = getNewScrubbingLogger(c)
 	c.cliSettings = NewCliSettings(c)
+	c.prepareDefaultEnvChannel = make(chan bool, 1)
```
andrewrobinsonhodges-snyk, line 253: Should probably add a close as a deferred func in case of errors

infrastructure/cli/cli.go
```diff
@@ -93,13 +93,15 @@ func (c *SnykCli) doExecute(ctx context.Context, cmd []string, workingDir types.
 }
 
 func (c *SnykCli) getCommand(cmd []string, workingDir types.FilePath, ctx context.Context) *exec.Cmd {
+	_ = c.c.WaitForDefaultEnv(ctx) // TODO - Handle the error.
```
andrewrobinsonhodges-snyk, line 96: TODO not done

infrastructure/oss/cli_scanner_test.go
```diff
@@ -177,6 +178,65 @@ func TestCLIScanner_getAbsTargetFilePathForPackageManagers(t *testing.T) {
...
+		inlineValueMutex:  &sync.RWMutex{},
+		packageScanMutex:  &sync.Mutex{},
+		runningScans:      make(map[types.FilePath]*scans.ScanProgress),
+		supportedFiles:    make(map[string]bool),
+		packageIssueCache: make(map[string][]types.Issue),
+	}
+
+	// Start building the command in a separate goroutine; it should block waiting on readiness
+	started := make(chan bool, 1)
+	unblocked := make(chan bool, 1)
```
andrewrobinsonhodges-snyk, line 208: Best to close these so the go func and require threads get cleaned up

application/server/configuration.go
```diff
@@ -321,22 +319,21 @@ func updateSnykCodeSecurity(c *config.Config, settings types.Settings) {
...
+// TODO stop using os env, move parsing to CLI
 func updatePathFromSettings(c *config.Config, settings types.Settings) {
-	// when changing the path from settings, we cache the original path first, to be able to restore it later
-	if len(cachedOriginalPath) == 0 {
-		cachedOriginalPath = os.Getenv("PATH")
-	}
-
 	if len(settings.Path) > 0 {
+		// If the user has a path set in their settings, we must wait for the default env to load fully first to prevent race conditions.
+		_ = c.WaitForDefaultEnv(context.Background())
```
bastiandoetsch, line 326: why don't we handle an error here? Seems important.

application/server/configuration.go
```diff
@@ -321,22 +319,21 @@ func updateSnykCodeSecurity(c *config.Config, settings types.Settings) {
...
 		c.Logger().Debug().Str("method", "updatePathFromSettings").Msgf("added configured path to PATH Environment Variable '%s'", os.Getenv("PATH"))
 		if err != nil {
 			c.Logger().Err(err).Str("method", "updatePathFromSettings").Msgf("couldn't add path %s", settings.Path)
 		}
-	} else {
-		_ = os.Setenv("PATH", cachedOriginalPath)
+	} else if c.IsDefaultEnvReady() {
+		// If the default env is ready, then we know any further wipes of the PATH need to reset the PATH.
+		// If it isn't ready, then we are still initializing and the goroutine will be taking care of setting the default env still.
+		_ = os.Setenv("PATH", c.GetCachedOriginalPath())
```
bastiandoetsch, line 336: I'd probably move the `os.SetEnv` out of the if and just update a path variable that is to be set, following a pattern of prepare what to do, then apply.

application/config/config_test.go
```diff
@@ -41,30 +41,33 @@ import (
...
 		token := uuid.New().String()
-		config := New()
-		SetCurrentConfig(config)
-		config.SetToken(token)
-		assert.Equal(t, config.Token(), token)
-		assert.NotEqual(t, config.Engine().GetConfiguration().Get(auth.CONFIG_KEY_OAUTH_TOKEN), token)
-		assert.Equal(t, config.Engine().GetConfiguration().Get(configuration.AUTHENTICATION_TOKEN), token)
+		c := New(WithBinarySearchPaths([]string{}))
+		require.NoError(t, c.WaitForDefaultEnv(t.Context()))
+		SetCurrentConfig(c)
```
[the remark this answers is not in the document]
rrama: NGL, I just kept the test the exact same (other than updating the call to `New()` and waiting) without giving it much thought.
Will remove it now.

bastiandoetsch (approved):

application/server/configuration.go
rrama: Done.

application/server/configuration.go
rrama: Done.

PR #974, opened 2025-08-21
fix: use whoami wf to check auth status

ShawkyZ:
### Description

### Checklist

### DIFF
--- a/mcp_extension/snyk_tools.json
+++ b/mcp_extension/snyk_tools.json
@@ -630,19 +630,19 @@
     },
     {
       "name": "snyk_send_feedback",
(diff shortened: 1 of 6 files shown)


mcp_extension/snyk_tools.json
```diff
@@ -12,7 +12,7 @@
...
+      "description": "Before invoking this tool, snyk_auth_status should be called to check if authentication is needed. WE NEED TO USE THE ABSOLUTE PATH IN THE PATH ARGUMENT. Analyzes projects for open-source vulnerabilities and license compliance issues by inspecting manifest files (e.g., package.json, pom.xml, requirements.txt) to understand dependencies and then queries the Snyk vulnerability database. \nWhen to use: During local development by developers on their workstations before committing changes for immediate feedback. \nHow to use: Test locally: run tool with at least the path parameter. \nPrerequisites: Project's package manager (e.g., Gradle, Maven, npm) must be installed for accurate dependency resolution.",
```
bastiandoetsch, line 15: why the change? I think the exit codes help?!

ShawkyZ: we don't expose exit code anymore

PR #985, opened 2025-09-02
fix: add path validation

acke:
### Description

### Checklist


internal/util/folder_config.go
bastiandoetsch, line 1: These files are utils - we should not name them folder_config, I'd actually just merge them into path.go and path_test.go.

internal/storedconfig/cross_platform_test.go
bastiandoetsch, line 1: Where is the cross-platform aspect tested now, e.g. tests are run on Windows, Linux & macOS?

internal/util/folder_config.go
```diff
@@ -0,0 +1,50 @@
...
+	"fmt"
+
+	"github.com/snyk/snyk-ls/internal/types"
+)
+
+func ValidateReferenceFolderPath(path types.FilePath) error {
+	options := PathValidationOptions{
+		AllowEmpty:    true,
+		RequireExists: false,
+	}
```
rrama, line 13: So we don't allow relative paths with traversal up the FS. I feel like this should not be disallowed, as I may want my reference folder to be like `../reference-folders/something`.

internal/util/folder_config.go
rrama: Not sure I agree, these are the requirement rules for the 2 fields in the folder config, so aren't generic rules.

internal/storedconfig/xdg.go
```diff
@@ -117,15 +124,35 @@ func UpdateFolderConfigs(conf configuration.Configuration, folderConfigs []types
...
+	if err := util.ValidateFolderPathLenient(folderConfig.FolderPath); err != nil {
+		logger.Error().Err(err).Str("path", string(folderConfig.FolderPath)).Msg("invalid folder path")
+		return err
+	}
+
+	// Validate the reference folder path for security if provided
+	if folderConfig.ReferenceFolderPath != "" {
+		if err := util.ValidateReferenceFolderPath(folderConfig.ReferenceFolderPath); err != nil {
+			logger.Error().Err(err).Str("referencePath", string(folderConfig.ReferenceFolderPath)).Msg("invalid reference folder path")
+			return err
```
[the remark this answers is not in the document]
acke: Yes, we now have a notifier that sends the validation result to user if it fails to validate.

internal/storedconfig/cross_platform_test.go
```diff
@@ -76,12 +75,21 @@ func Test_GetOrCreateFolderConfig_CrossPlatformPaths(t *testing.T) {
...
 			require.NoError(t, err)
 			require.NotNil(t, folderConfig)
-			require.Equal(t, tt.inputPath, folderConfig.FolderPath)
+
+			// Calculate the expected normalized path for this specific input
+			expectedPath := util.PathKey(tt.inputPath)
+			require.Equal(t, expectedPath, folderConfig.FolderPath)
+
+			// For paths that should normalize to the same result as the base case, verify they do
+			if tt.name == "Path with trailing slash" || tt.name == "Path with whitespace" {
```
rrama, line 84: Not a fan of doing this based on the tc name, I'd personally add an additional flag to the tcs for whether they should match the base path or not.

internal/storedconfig/xdg.go
```diff
@@ -117,15 +123,35 @@ func UpdateFolderConfigs(conf configuration.Configuration, folderConfigs []types
...
+	// Normalize paths for consistent storage
+	normalizedFolderConfig := *folderConfig
+	normalizedFolderConfig.FolderPath = util.PathKey(folderConfig.FolderPath)
+	if folderConfig.ReferenceFolderPath != "" {
+		normalizedFolderConfig.ReferenceFolderPath = util.PathKey(folderConfig.ReferenceFolderPath)
+	}
+
 	// Generate normalized key for consistent cross-platform storage
-	normalizedPath := util.GenerateFolderConfigKey(folderConfig.FolderPath)
+	normalizedPath := util.PathKey(folderConfig.FolderPath)
```
rrama, line 152: Move the variable up, so it can be used for [this line](https://github.com/snyk/snyk-ls/pull/985/files#diff-61ccb9354e67b24f8ced5b8cb47b91bdd5b116bfe5ae597a07f4521568a5a226R146) too:
```go
normalizedFolderConfig.FolderPath = util.PathKey(folderConfig.FolderPath)
```

internal/storedconfig/xdg.go
```diff
@@ -117,15 +123,35 @@ func UpdateFolderConfigs(conf configuration.Configuration, folderConfigs []types
...
 
-	sc.FolderConfigs[normalizedPath] = folderConfig
+	sc.FolderConfigs[normalizedPath] = &normalizedFolderConfig
```
rrama, line 154: Q: Do we delete the old non-normalised path version(s) from the object?

internal/util/path.go
```diff
@@ -1,23 +1,164 @@
...
 	"strings"
 
 	"github.com/snyk/snyk-ls/internal/types"
 )
 
-// GenerateFolderConfigKey creates a normalized key for folder config storage
-// This ensures consistent cross-platform map keys while preserving original paths
-func GenerateFolderConfigKey(p types.FilePath) types.FilePath {
+// Common dangerous characters that could be used for injection attacks
+var dangerousChars = []string{";", "&", "|", "`", "$", "\"", "'", "\n", "\r", "\t"}
```
rrama, line 13: Hmm, maybe `*` too.

internal/util/path.go
```diff
@@ -1,23 +1,164 @@
...
+	}
+
+	if !info.IsDir() {
+		return fmt.Errorf("path must be a directory: '%s'", input)
+	}
+	return nil
+}
+
+// ValidatePath validates any path for security with customizable requirements
+func ValidatePath(path types.FilePath, options PathValidationOptions) error {
```
rrama, line 65: Since we enforce it being a dir, the func name should reflect that, or have it as one of the `PathValidationOptions` to check if dir or file.
Could be like: `Existence: ExistAsDirectory`, `Existence: ExistAsFileOrDirectory`, `Existence: DoesNotExist`, etc.

internal/util/path.go
```diff
@@ -1,23 +1,164 @@
...
+	for _, char := range dangerousChars {
+		if strings.Contains(input, char) {
+			return fmt.Errorf("dangerous character detected in '%s': %s", input, char)
+		}
+	}
+	return nil
+}
+
+// validatePathTraversal checks for path traversal attempts
+func validatePathTraversal(input string) error {
```
rrama, line 26: Technically should it be `validateNoPathTraversals`?
I only thought of this on my 3rd pass through, so not too fussed either way, your call.

internal/util/path_test.go
```diff
@@ -9,45 +10,48 @@ import (
...
 	tests := []struct {
 		name     string
 		input    types.FilePath
 		expected types.FilePath
 	}{
 		{
 			name:     "Unix path without trailing slash",
-			input:    "/Users/foo/project",
-			expected: "/Users/foo/project",
+			input:    types.FilePath(tempDir),
```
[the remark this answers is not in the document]
rrama: I see what you have done in the other test now, may want to do similar for this one or just rely on the os matrix and don't have the "Unix path without trailing slash" and "Windows path without trailing slash", just having "Path without trailing slash"

internal/util/path_test.go
```diff
@@ -1,53 +1,57 @@
...
-			expected: "C:/Users/foo/project",
+			input:    types.FilePath(tempDir + "/"),
+			expected: types.FilePath(tempDir),
 		},
 		{
 			name:     "Mixed separators",
-			input:    `C:\Users\foo\project`,
-			expected: "C:/Users/foo/project",
+			input:    types.FilePath(tempDir),
+			expected: types.FilePath(tempDir),
```
rrama, line 49: This isn't mixed anymore.

internal/util/path_test.go
```diff
@@ -62,19 +66,207 @@ func TestGenerateFolderConfigKey(t *testing.T) {
...
+				}
+			} else {
+				assert.NoError(t, err)
+			}
+		})
+	}
+}
+
+// createCommonTestCases creates the shared test cases for both validation functions
+func createCommonTestCases(tempDir string, errorPrefix string) []testCase {
```
rrama, line 138: `errorPrefix` now unused?

PR #989, opened 2025-09-08
chore: update golangci-lint to v2.3.0 [IDE-1377]

rrama:
### Description

Migrate stylecheck, gosimple, and varcheck into staticcheck (they are now combined).
Enable staticcheck explicitly, it was previously enabled, but was not listed.
Remove deprecated `EXC####` exclusion rules.
Disable package comment rules (ST1000).
Disable naming convention rules (ST1003).
Move the timeout to command-line arguments (v2 discourages timeouts in the config).
Fix any new lint errors.
Improve updating of golangci-lint and pact, now will be automatically updated if needed before use.
Add a target to remove the tools dir on `make clean-tools` and `make-clean`, with a new `make clean-build` target to delete just the build directory.

### Notes for the reviewer

I would recommend comparing `.golangci.yaml` with whitespace ignored.
Migration guide: https://golangci-lint.run/docs/product/migration-guide/

### Checklist

 - None added.
 - N/A
 - N/A
 - N/A

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -28,7 +28,7 @@ require (
 	github.com/rs/zerolog v1.34.0
 	github.com/sabhiram/go-gitignore v0.0.0-20210923224102-525f6e181f06
 	github.com/shirou/gopsutil v3.21.11+incompatible
-	github.com/snyk/code-client-go v1.23.4
+	github.com/snyk/code-client-go v1.23.7
 	github.com/snyk/go-application-framework v0.0.0-20250825133112-d14fa1a6b7be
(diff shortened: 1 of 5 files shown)


internal/types/lsp.go
```diff
@@ -390,11 +390,9 @@ type TextDocumentClientCapabilities struct {
 	} `json:"inlineValue,omitempty"`
 }
 
-/**
- * A parameter literal used in inline value requests.
- *
- * @since 3.17.0
- */
```
rrama, line 397: If you prefer this type of doc string for the LSP types, so it stays an exact copy from the JS then let me know and I will add an ignore for this file and revert the changes.

PR #1003, opened 2025-09-29
fix(MCP): pre-populate trusted folders from env var [AG-5]

ShawkyZ:
### Description

### Checklist

### DIFF
--- a/mcp_extension/trust/trust.go
+++ b/mcp_extension/trust/trust.go
@@ -24,6 +24,7 @@ import (
 	"html/template"
 	"net"
 	"net/http"
+	"os"
 	"path/filepath"
 	"runtime"
 	"strings"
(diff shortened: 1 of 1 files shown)


mcp_extension/trust/trust.go
```diff
@@ -53,10 +54,24 @@ type FolderTrust struct {
 var SnykTrustPage string
 
 func NewFolderTrust(logger *zerolog.Logger, config configuration.Configuration) *FolderTrust {
-	return &FolderTrust{
+	folderTrust := &FolderTrust{
 		logger: logger,
 		config: config,
 	}
+
+	// Pre-populate trusted folders from TRUSTED_FOLDER environment variable if present.
```
bastiandoetsch, line 62: Why not get them from ls-config files?

PR #1085, opened 2025-12-10
test: unified converter tests

rrama:
### Description

AI took the tests from #1042 and updated them to work with #1054 which went into main. It took 2 days of the AI spinning in the background, I then took what it did and made the tests proper. 
There is also a fix to ensure we don't panic if we receive malformed data with a shorter than expected dependency path.

### Checklist

 - N/A
 - N/A
 - N/A


infrastructure/oss/unified_converter_test.go
```diff
@@ -0,0 +1,1297 @@
...
+
+	workDir, err := filepath.Abs("testdata")
+	require.NoError(t, err)
+
+	filePath := filepath.Join(workDir, "package.json")
+	ctx = ctx2.NewContextWithWorkDirAndFilePath(ctx, types.FilePath(workDir), types.FilePath(filePath))
+
+	// Create deps with WRONG types (strings instead of proper types)
+	deps := map[string]any{
+		ctx2.DepConfig:        "not-a-config",    // Should be *config.Config
```
bastiandoetsch, line 705: [Critical] This test file does not compile against `main` — it was written against an older API and not recompiled after merge. `go vet ./infrastructure/oss/` fails, and because Go builds one test binary per package, this breaks the **entire `infrastructure/oss` test suite**: every test in the package fails to run, so the coverage this PR adds is currently zero. Confirmed undefined symbols:

- `ctx2.DepConfig` (lines 705, 854) — the context package defines `DepConfigResolver`/`DepConfiguration`/`DepEngine`/…, there is no `DepConfig`.
- `c.SetSnykOSSQuickFixCodeActionsEnabled` (line 833) — exists nowhere in the repo.
- `error_reporting.NewTestErrorReporter()` (line 846) — signature is `NewTestErrorReporter(workflow.Engine)`, called with 0 args.
- The anonymous `Relationships` struct literals (lines ~988, ~1192) no longer match `testapi.FindingData` (Id fields are `oapi-codegen/runtime/types.UUID`, and a `Project` relationship was added).

Fix: rebuild the test against current APIs — use the real DI keys (`DepEngine` + `ConfigResolverFromContext` + `FolderConfigFromContext`), the real config-flag mechanism, `NewTestErrorReporter(engine)`, and current generated types. Add `go test ./infrastructure/oss/` to CI as a build gate — a green CI here would mean the package's tests aren't being compiled.

— AI review

infrastructure/oss/unified_converter.go
```diff
@@ -400,7 +401,7 @@ func extractUpgradePackage(dependencyPath []string, finding *testapi.FindingData
 		return nil
 	}
 
-	if len(dependencyPath) == 0 {
+	if len(dependencyPath) < 2 {
 		return nil
```
bastiandoetsch, line 405: [Should Fix] Good fix here — but the identical bug class remains a few lines below at line 412:

```go
if len(path) > 0 && path[1].Name == depPathPackageName {
```

`path[1]` requires length ≥ 2, but the guard only checks `len(path) > 0`, so a single-element `UpgradePath.DependencyPath` (API data) panics with index-out-of-range — exactly what you just fixed for `dependencyPath`. Tighten to `len(path) > 1`, and add a test feeding a length-1 upgrade path (mirroring the single-element case you already added for the input side). 3 of 4 reviewers flagged this; no current test exercises it.

— AI review

bastiandoetsch:
Context for this review: the only prior review on this PR was from `gemini-code-assist` (a bot), so it hadn't had a human/auth review yet.

The **source change is correct**: `extractUpgradePackage` line 405 (`len(dependencyPath) == 0` → `< 2`) is a genuine root-cause fix for the `dependencyPath[1]` access two lines down, and the comment edits are net-positive (incl. the `vuln.`→`problem.` typo fix). Security scan: clean, no PR-attributable issues. Where it compiles, the tests assert real behaviour against real production functions (not tautologies) — the single-element cases would genuinely panic against the old guard.

Non-blocking follow-ups once the compile is fixed:
- [Suggestion] The skipped 'TODO' + active 'Wrong behavior' tests deliberately pin a known production bug (`//FIXME: don't stop at the first dependency path`) as passing. The pattern is honest (with delete-me notes), but please ensure a Jira ticket tracks the FIXME so these don't become permanent.
- [Suggestion] The 'deep transitive returns empty remediation' test proves the function returns `""` but not that empty remediation is the *intended* product behaviour — worth confirming against legacy parity (an empty remediation flows into `buildMessage` as a trailing `". "`).
- [Suggestion] The ~45-line anonymous `Relationships` struct is inlined in ~4 helpers; have `createCompleteUnifiedFinding` reuse `createBaseFindingWithRelationshipsStruct` so an upstream `testapi` regen doesn't break all copies at once.
- [Suggestion] Four near-identical npm/yarn/maven/pip cases test per-ecosystem text at the wrong layer — collapse to two, or test `buildOutdatedDependencyMessage` directly.

— AI review

infrastructure/oss/unified_converter_test.go
basti-snyk, line 705: **Critical — this test file does not compile against HEAD, so the entire `infrastructure/oss` test package fails to build.** `go test ./infrastructure/oss/` fails (this is also why CI lint is red and the unit/race/integration jobs are skipped — none of the ~1300 new lines of assertions have ever run). The file references an API that no longer exists at HEAD:

- `ctx2.DepConfig` (here at line 705, and line 854) — undefined. The context package has `DepEngine`, `DepConfigResolver`, `DepConfiguration`, `DepLearnService`, `DepErrorReporter`, but no `DepConfig`.
- `config.Config` (line ~818) — no such exported type in `application/config`.
- `c.SetSnykOSSQuickFixCodeActionsEnabled(...)` (line ~833) — `c` is the `workflow.Engine` returned by `testutil.UnitTestWithCtx`, which has no such method.
- `error_reporting.NewTestErrorReporter()` (line ~846) — now requires a `workflow.Engine` argument; see the sibling call in `vulnerability_count_test.go`.
- The hand-rolled anonymous `Relationships` struct literals (lines ~988, ~1192) no longer match the generated `testapi` type — it gained a `Project` field and uses `oapi-codegen/runtime/types.UUID` for `Id`, while the literals use `github.com/google/uuid.UUID` and omit `Project`.

This needs a rebase onto current `main` and an update to the live API: obtain the engine/reporter the way the package's existing tests do, key deps by the real `Dep*` constants, and build `Relationships` from the generated `testapi` types rather than re-declaring the anonymous struct (the duplicated struct shape is the root cause of the last breakage and will keep breaking on every API change). Confirm `go test ./infrastructure/oss/` passes before merge.

— AI review

infrastructure/oss/unified_converter.go
```diff
@@ -400,7 +401,7 @@ func extractUpgradePackage(dependencyPath []string, finding *testapi.FindingData
...
 	}
 
-	if len(dependencyPath) == 0 {
+	if len(dependencyPath) < 2 {
```
basti-snyk, line 404: This guard fix is correct — `dependencyPath[1]` at line 407 would panic on a length-1 path, and `< 2` cleanly rejects both empty and single-element inputs while preserving behavior for valid (len ≥ 2) inputs.

**Should fix — the structurally identical twin panic just below is still unguarded.** At line 412, `path[1].Name` is reached after only `len(path) > 0`, so an upgrade `DependencyPath` with exactly one package panics with index-out-of-range. That `path` slice comes from the API (`upgradeAction.UpgradePaths[…].DependencyPath`) and has no length invariant. Since this PR's theme is hardening this exact access pattern, tighten line 412 to `len(path) > 1` to match, and add a test case with a single-element upgrade path — none of the new tests currently feed one, so the "avoid panic" coverage they imply doesn't actually exist.

— AI review

PR #1131, opened 2026-01-27
feat: Org Specific Config in Language Server [IDE-1711][IDE-1637][IDE-1644] 

andrewrobinsonhodges-snyk:
## LDX-Sync Caching

- New LDXSyncConfigCache (internal/types/ldx_sync_config.go) - Caches org configs returned from LDX-Sync API
- LDXSyncOrgConfig stores per-org settings with locked/enforced field metadata
- Folder-to-org mapping - Maps workspace folders to their determined org IDs
- Cache is populated when RefreshConfigFromLdxSync runs after authentication/folder changes

## Config Resolution by Org [IDE-1711]

- New ConfigResolver (internal/types/config_resolver.go) - Resolves effective config values with priority:

  1. User overrides (stored per-folder)
  2. LDX-Sync org config (locked/enforced fields)
  3. Global IDE settings
  4. Defaults

- Folder-aware config accessors on Config: FilterSeverityForFolder(), IsSnykOssEnabledForFolder(), etc.
- LDXSyncAdapter (internal/types/ldx_sync_adapter.go) - Converts LDX-Sync API responses to internal config format

## Separation of Stored Config vs LSP Messages

- StoredFolderConfig (internal/types/stored_folder_config.go) - Persisted config with user overrides, org settings, feature flags, metadata from LDX-sync
- LspFolderConfig (internal/types/lsp.go) - Wire format for $/snyk.folderConfigs notification with PATCH semantics:
  - Omitted = don't change
  - Null = reset to default
  - Value = set override
- LspConfigurationParam - New notification ($/snyk.configuration) for global/machine-wide settings [IDE-1637]
- NullableField[T] - Generic type supporting three-state JSON semantics (omitted/null/value)

## Settings Validation [IDE-1644]

- Locked field validation - validateLockedFields() rejects changes to fields locked by org policy
- Enforced field handling - Enforced values from LDX-Sync override user settings
- clearLockedOverridesFromStoredFolderConfigs() - Clears user overrides when a field becomes locked

Key Files Changed

- internal/types/stored_folder_config.go - New persistent config type
- internal/types/config_resolver.go - Config resolution logic
- internal/types/ldx_sync_config.go - LDX-Sync cache types
- application/server/configuration.go - Processing of incoming LspFolderConfig
- domain/ide/command/folder_handler.go - Building outgoing notifications

### Checklist

[IDE-1711]: https://snyksec.atlassian.net/browse/IDE-1711?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[IDE-1637]: https://snyksec.atlassian.net/browse/IDE-1637?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nick-y-snyk (commented):
## 🔴 Critical: Race Condition in Cache Access

**File:** `application/config/config.go:1553-1557`

```go
func (c *Config) GetLdxSyncOrgConfigCache() *types.LDXSyncConfigCache {
	c.ldxSyncConfigCacheMutex.RLock()
	defer c.ldxSyncConfigCacheMutex.RUnlock()
	return c.ldxSyncConfigCache
}
```

**Critical race condition here.** We're returning a pointer to the cache while holding the read lock, but the lock gets released immediately when the function returns. This means callers can access and modify the cache without any mutex protection.

Look at how this is used in `domain/ide/command/ldx_sync_service.go:174`:

```go
cache := c.GetLdxSyncOrgConfigCache()  // RLock released here
cache.SetFolderOrg(folderPath, orgId)   // Writing to cache with NO lock!
```

This same pattern appears in:
- `application/server/configuration.go:544`
- `application/server/configuration.go:1033`
- `domain/ide/command/folder_handler.go:73`

With the parallel goroutines in `RefreshConfigFromLdxSync`, we're definitely hitting this race. Multiple goroutines are calling `cache.SetFolderOrg()` concurrently with zero synchronization.

**Suggested fix:**

**Option A** - Add internal locking to the cache methods:
```go
func (c *LDXSyncConfigCache) SetFolderOrg(folderPath FilePath, orgId string) {
	c.mu.Lock()
	defer c.mu.Unlock()
	// ... existing logic
}
```

**Option B** - Document that callers must acquire locks externally and provide explicit lock/unlock methods:
```go
func (c *Config) LockLdxSyncCache() { c.ldxSyncConfigCacheMutex.Lock() }
func (c *Config) UnlockLdxSyncCache() { c.ldxSyncConfigCacheMutex.Unlock() }
```

I'd prefer **Option A** - push the locking into the cache itself. Cleaner API and harder to misuse.

nick-y-snyk (commented):
## 🔴 Critical: Nil Pointer Risk in ConfigResolver

**File:** `internal/types/config_resolver.go:256-262`

```go
func (r *ConfigResolver) getOriginScope(settingName string, folderConfig *StoredFolderConfig) string {
	scope := GetSettingScope(settingName)

	switch scope {
	case SettingScopeMachine:
		// ...
	case SettingScopeOrg:
		if folderConfig \!= nil && r.ldxSyncCache \!= nil && r.c \!= nil {
			effectiveOrg := r.c.FolderOrganization(folderConfig.FolderPath)
```

Good defensive check here, but we're inconsistent about it. Compare this to line 479:

```go
func (r *ConfigResolver) IsLocked(settingName string, folderConfig *StoredFolderConfig) bool {
	if folderConfig == nil || r.ldxSyncCache == nil || r.c == nil {
		return false
	}

	effectiveOrg := r.c.FolderOrganization(folderConfig.FolderPath)  // Uses r.c without re-checking
}
```

And then at line 494 in `IsEnforced()` we have the same pattern.

The issue is: we check `r.c \!= nil` at the top of these functions, but `r.c` is set in the constructor and never validated. What if someone calls `NewConfigResolver(cache, settings, nil, logger)` with a nil ConfigProvider? (Which actually happens in tests - see `config_resolver_test.go:51`).

**Two options:**

1. **Require `c` to be non-nil** - panic in constructor if it's nil (except for machine-scope-only use cases)
2. **Always check before use** - but then make it consistent everywhere

I'd lean toward option 1 with a clear contract: if you're resolving org-scoped settings, you MUST provide a ConfigProvider. Document this in the constructor and make it explicit when nil is acceptable (machine-scope only scenarios).

nick-y-snyk (commented):
## 🔴 Critical: Error Results Stored in LDX-Sync Cache

**File:** `domain/ide/command/ldx_sync_service.go:138-149`

```go
// Store result in temporary map (even if there's an error)
// This allows ResolveOrg to distinguish between "never attempted" and "attempted but failed"
resultsMutex.Lock()
results[f.Path()] = &cfgResult
resultsMutex.Unlock()

if cfgResult.Error \!= nil {
	logger.Err(cfgResult.Error).
		Str("folder", string(f.Path())).
		Msg("Failed to get user config from LDX-Sync")
	return
}
```

I understand the intent here (distinguishing between "not attempted" and "failed"), but storing error results and then processing them downstream is dangerous.

In `updateOrgConfigCache` at line 179:
```go
for folderPath, result := range results {
	if result == nil || result.Config == nil {
		continue
	}
```

We check `result.Config == nil`, but what if the API returns a **partially populated** Config object before failing? Or what if the error occurred during JSON deserialization and we have garbage data in `result.Config`?

**Suggested approach:**

1. Store successes and failures separately:
```go
successResults := make(map[types.FilePath]*ldx_sync_config.LdxSyncConfigResult)
failedFolders := make(map[types.FilePath]error)
```

2. Only pass `successResults` to `updateOrgConfigCache`
3. Handle `failedFolders` explicitly - maybe send warnings to the IDE or implement retry logic

This makes the data flow clearer: we either have valid config OR an error, never both.

nick-y-snyk (commented):
## 🟡 Medium: Locked Field Validation Timing Issue

**File:** `application/server/configuration.go:318-330`

```go
normalizedPath := types.PathKey(path)
if incoming, hasIncoming := incomingMap[normalizedPath]; hasIncoming {
	// Validate locked fields before applying
	hasLockedFieldRejections := validateLockedFields(c, &folderConfig, &incoming, &logger)
	if hasLockedFieldRejections {
		projectName := filepath.Base(string(folderConfig.FolderPath))
		notifier.SendShowMessage(sglsp.MTWarning,
			fmt.Sprintf("Failed to update %s: Some settings are locked by your organization's policy", projectName))
	}

	// Apply the PATCH update
	folderConfig.ApplyLspUpdate(&incoming)
}
```

The validation approach here feels fragile. We parse the entire `LspFolderConfig` from JSON (including all the locked fields), then try to "un-set" them by marking `Present=false` in `clearLockedField()`:

```go
func clearLockedField(incoming *types.LspFolderConfig, settingName string) {
	switch settingName {
	case types.SettingEnabledSeverities:
		incoming.EnabledSeverities.Present = false
	// ...
```

**Two problems:**

1. **The value was already deserialized** - any side effects from `UnmarshalJSON` already happened
2. **Future-proofing** - if someone modifies `ApplyLspUpdate()` to ignore the `Present` flag, we silently bypass the lock

**Better approach:** Reject the update entirely if ANY locked field is present:

```go
if hasIncoming {
	if lockedFields := checkForLockedFields(c, &folderConfig, &incoming); len(lockedFields) > 0 {
		projectName := filepath.Base(string(folderConfig.FolderPath))
		notifier.SendShowMessage(sglsp.MTError,
			fmt.Sprintf("Cannot update %s: fields %v are locked by org policy", 
				projectName, lockedFields))
		return folderConfig, false  // Don't apply ANY changes
	}
	folderConfig.ApplyLspUpdate(&incoming)
}
```

This is clearer: if you try to change a locked field, we reject the whole request. User gets immediate feedback, and we don't have to rely on the `Present` flag working correctly.

nick-y-snyk (commented):
## 🟡 Medium: Global Setting Propagation Performance and Correctness

**File:** `application/server/configuration.go:1015-1073`

```go
func propagateOrgScopedGlobalSettingsToStoredFolderConfigs(c *config.Config, settingName string, value any) {
	// ...
	for folderPath, fc := range sc.StoredFolderConfigs {
		if fc == nil {
			continue
		}

		// Check if this field is locked for this folder's org
		effectiveOrg := cache.GetOrgIdForFolder(folderPath)
		// ...
		fc.SetUserOverride(settingName, value)
		modified = true
	}

	if modified {
		if err := storedconfig.Save(gafConfig, sc); err != nil {
```

This function gets called on **every global settings change** and iterates through **every single folder**. Three concerns:

**1. Performance:** For a workspace with 100 folders, we're:
- Iterating all 100 folders
- Looking up org for each folder from cache
- Checking if the field is locked (another cache lookup)
- Persisting the entire stored config to disk

If a user toggles "Snyk Code Enabled" on and off a few times, we're hammering the disk.

**2. Race condition potential:** We load the stored config, modify it, save it. But what if a scan is concurrently reading a folder's config? We're holding `ldxSyncConfigCacheMutex` during the cache lookups, but not during the file I/O to save.

**3. Timing issue:** What if:
- User adds a new folder to workspace
- Before we run LDX-sync for that folder, user changes a global setting
- `propagateOrgScopedGlobalSettingsToStoredFolderConfigs` runs
- New folder isn't in `sc.StoredFolderConfigs` yet
- Setting doesn't get propagated

Then when the folder is added later, it won't have the user's global preferences.

**Suggestions:**

1. **Batch the propagation** - don't save on every setting change, debounce it
2. **Make it lazy** - instead of eagerly propagating, have the ConfigResolver check global settings if no user override exists (which it already does at line 209 in `config_resolver.go`). Then we only need to propagate on LDX-sync updates, not on every settings change
3. **Document the timing assumption** - clarify when this propagation happens relative to folder addition/removal

Actually, looking at the ConfigResolver logic... do we even need this propagation? The resolver already falls back to global settings when there's no user override. So why are we eagerly writing user overrides for ALL folders? Just to persist user intent? This needs clarification in the code comments.

nick-y-snyk (commented):
## 🟡 Medium: Path Normalization Consistency

**Files:** Multiple
**Lines:** Various

We're using `types.PathKey()` for path normalization in some places but not others. This is going to cause cross-platform bugs.

**Examples of correct usage:**

`configuration.go:278`:
```go
normalizedPath := types.PathKey(fc.FolderPath)
incomingMap[normalizedPath] = fc
```

`ldx_sync_config.go:169`:
```go
c.FolderToOrgMapping[PathKey(folderPath)] = orgId
```

**Examples where normalization is missing:**

`ldx_sync_service.go:141`:
```go
results[f.Path()] = &cfgResult  // f.Path() not normalized
```

Then later at line 179:
```go
for folderPath, result := range results {
	// ...
	cache.SetFolderOrg(folderPath, orgId)  // Using the key from map directly
}
```

`SetFolderOrg` normalizes internally (good!), but the `results` map keys aren't normalized. So if we try to look up a folder from the results map later using a normalized path, we'll miss it.

**Suggested fix:**

1. **Normalize at the boundaries** - as soon as we receive a path from external sources (IDE, API, filesystem), normalize it
2. **Document the invariant** - `types.FilePath` values should ALWAYS be normalized. Add a comment to the type definition
3. **Add a constructor** - instead of `types.FilePath(rawPath)`, use `types.NewFilePath(rawPath)` which normalizes

Something like:
```go
func NewFilePath(path string) FilePath {
	return PathKey(FilePath(path))
}
```

Then we can grep for `FilePath(` and make sure we're using the constructor everywhere. This pattern would catch cross-platform path bugs at compile time instead of mysteriously at runtime.

nick-y-snyk (commented):
## 🟡 Medium: Inconsistent Error Handling Patterns

**Files:** Multiple
**Lines:** Various

We have three different patterns for handling errors in config updates, and it's not clear when to use which:

**Pattern 1 - Log and return:**
`ldx_sync_service.go:359`:
```go
if err := storedconfig.GetStoredConfig(gafConfig, logger, true); err != nil {
	logger.Err(err).Msg("Failed to get stored config for clearing locked overrides")
	return
}
```

**Pattern 2 - Log, send diagnostic to IDE, but continue:**
`configuration.go:339`:
```go
err := c.UpdateStoredFolderConfig(&folderConfig)
if err != nil {
	logger.Err(err).Msg("failed to update folder config")
	notifier.SendErrorDiagnostic(path, err)
}
```

**Pattern 3 - Log but don't tell user:**
`configuration.go:1067`:
```go
if err := storedconfig.Save(gafConfig, sc); err != nil {
	logger.Err(err).Msg("Failed to save stored config after propagating global setting")
} else {
	logger.Debug().Msg("Saved stored config after propagating global setting")
}
```

**The question:** When should the user be notified about config errors vs. when should we silently degrade?

**Suggested policy:**

1. **User-initiated changes** (IDE sends us a config update) → Always notify on failure via `SendErrorDiagnostic` or `SendShowMessage`
2. **System-initiated changes** (LDX-sync, propagation) → Log at WARN level but don't spam the user UNLESS it affects their current work
3. **Transient failures** (network errors during LDX-sync) → Retry with backoff, only notify if it keeps failing

Right now it feels inconsistent - we tell the user about some failures but not others, with no clear pattern.

Also: in `configuration.go:1067`, if we fail to save the propagated settings, we log it but continue. The in-memory state is now out of sync with the persisted state. Next time we restart, the user's settings will be wrong. Shouldn't we at least send a warning to the IDE?

application/config/config.go
```diff
@@ -1519,32 +1541,175 @@ func (c *Config) SetSecureAtInceptionExecutionFrequency(frequency string) {
 	c.secureAtInceptionExecutionFrequency = frequency
 }
 
-// InitLdxSyncCache initializes the LDX-Sync cache map
-func (c *Config) InitLdxSyncCache() {
-	c.ldxSyncCacheMutex.Lock()
-	defer c.ldxSyncCacheMutex.Unlock()
-	c.ldxSyncCache = make(map[types.FilePath]*ldx_sync_config.LdxSyncConfigResult)
+// InitLdxSyncOrgConfigCache initializes the LDX-Sync org config cache and ConfigResolver
+func (c *Config) InitLdxSyncOrgConfigCache() {
```
bastiandoetsch: This does not need to be exported

application/config/config.go
```diff
@@ -1310,26 +1310,48 @@ func (c *Config) SetSnykOpenBrowserActionsEnabled(enable bool) {
...
-func (c *Config) UpdateFolderConfig(folderConfig *types.FolderConfig) error {
-	return storedconfig.UpdateFolderConfig(c.engine.GetConfiguration(), folderConfig, c.logger)
+// StoredFolderConfigReadOnly returns the folder config for a path without writing to storage or enriching from Git.
+// This is suitable for read-only configuration checks. If no config exists in storage, this creates one with default
+// values (OrgMigratedFromGlobalConfig=true, OrgSetByUser=false, FeatureFlags initialized) but does not persist it.
+func (c *Config) StoredFolderConfigReadOnly(path types.FilePath) *types.StoredFolderConfig {
+	folderConfig, err := storedconfig.GetStoredFolderConfigWithOptions(c.engine.GetConfiguration(), path, c.Logger(), storedconfig.GetStoredFolderConfigOptions{
+		CreateIfNotExist: true,
+		ReadOnly:         true,
+		EnrichFromGit:    false,
```
bastiandoetsch: why?

application/server/configuration.go
```diff
@@ -272,47 +298,125 @@ func gatherAllFolderPaths(incomingMap map[types.FilePath]types.FolderConfig, wor
...
+		types.SettingIssueViewIgnoredIssues: incoming.IssueViewIgnoredIssues.Present,
+	}
+
+	for settingName, hasUpdate := range fieldsToCheck {
+		if !hasUpdate {
+			continue
+		}
+		_, source := resolver.GetValue(settingName, folderConfig)
+		if source == types.ConfigSourceLDXSyncLocked {
+			logger.Warn().
```
bastiandoetsch: should just be info. it's an expected scenario.

bastiandoetsch (commented):
started reviewing, more coming...

infrastructure/configuration/template/config.html
```diff
@@ -375,6 +390,150 @@ <h2>{{.FolderLabel}} Settings</h2>
...
+                                <select id="folder_{{$index}}_override_scan_automatic" name="folder_{{$index}}_override_scan_automatic" data-setting="scan_automatic" {{if isLocked $folder.EffectiveConfig "scan_automatic"}}disabled{{end}}>
+                                    <option value="auto" {{if isAutoScan .Value}}selected{{end}}>Auto</option>
+                                    <option value="manual" {{if not (isAutoScan .Value)}}selected{{end}}>Manual</option>
+                                </select>
+                            </div>
+                            {{end}}
+
+                            <!-- Delta Findings Override -->
+                            {{with getEffectiveValue $folder.EffectiveConfig "scan_net_new"}}
+                            <div class="form-group {{if isLocked $folder.EffectiveConfig "scan_net_new"}}field-locked{{end}}">
```
bastiandoetsch: here, too. did you try the dialog?

infrastructure/configuration/template/config.html
```diff
@@ -375,6 +390,150 @@ <h2>{{.FolderLabel}} Settings</h2>
...
+                        <span data-toggle="tooltip" title="Override organization policy settings for this folder. Locked settings cannot be changed.">Policy Overrides</span>
+                        <span class="collapse-icon">▼</span>
+                    </button>
+                    <div class="collapse ml-1" id="folder-{{$index}}-overrides">
+                        <div class="mt-2">
+                            <p class="text-muted small">These settings can be customized per-folder. Locked settings are enforced by your organization.</p>
+                            
+                            <!-- Scanning Mode Override -->
+                            {{with getEffectiveValue $folder.EffectiveConfig "scan_automatic"}}
+                            <div class="form-group {{if isLocked $folder.EffectiveConfig "scan_automatic"}}field-locked{{end}}">
```
bastiandoetsch: what about enforced?

application/config/config.go
```diff
@@ -210,8 +209,9 @@ type Config struct {
 	userSettingsPath                    string
 	autoConfigureMcpEnabled             bool
 	secureAtInceptionExecutionFrequency string
-	ldxSyncCache                        map[types.FilePath]*ldx_sync_config.LdxSyncConfigResult
-	ldxSyncCacheMutex                   sync.RWMutex
+	ldxSyncConfigCache                  *types.LDXSyncConfigCache
+	ldxSyncConfigCacheMutex             sync.RWMutex
+	configResolver                      *types.ConfigResolver
```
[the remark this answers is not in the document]
bastiandoetsch: The goal should be to get away from a custom config that lives apart from GAF configuration. GAF configuration is accessible everywhere. Config Resolver could live either session based in the context (as other DI components) or in the gaf configuration, needing just an interface at the user side.

bastiandoetsch (approved):

infrastructure/configuration/template/config.html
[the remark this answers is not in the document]
andrewrobinsonhodges-snyk: I think they're OK - it's a Go/HTML templating thing.

infrastructure/configuration/template/config.html
andrewrobinsonhodges-snyk: Ditto. I did try it. It needs some work on the styling, but it does render correctly.

PR #1173, opened 2026-03-05
fix: server lifecycle context [IDE-1791]

rrama:
### Description

Done by passing a context to the loading of the configured environment.
This will prevent interactive terminals being launched when the server shutsdown.
Currently targeting the PR branch in GAF (https://github.com/snyk/go-application-framework/pull/564). Change before merge.

### Checklist

 - N/A
 - N/A
 - N/A

### DIFF
--- a/application/config/config.go
+++ b/application/config/config.go
@@ -211,6 +211,8 @@ type Config struct {
 	offline                                bool
 	ws                                     types.Workspace
 	isLSPInitialized                       bool
+	serverLifecycleCtx                     context.Context
+	serverLifecycleCtxCancelFunc           context.CancelFunc
 	cachedOriginalPath                     string
 	userSettingsPath                       string
 	lastSetOrganization                    string // Trimmed raw org value last passed to SetOrganization
@@ -1554,6 +1556,40 @@ func (c *Config) SetLSPInitialized(initialized bool) {
 	c.m.Lock()
 	defer c.m.Unlock()
 	c.isLSPInitialized = initialized
+}
+
+func (c *Config) SetServerLifecycleContext(ctx context.Context, cancel context.CancelFunc) {
+	if ctx == nil {
+		panic("server lifecycle context must not be nil")
+	}
+	if cancel == nil {
+		panic("server lifecycle cancel func must not be nil")
+	}
+	c.m.Lock()
+	defer c.m.Unlock()
+	c.serverLifecycleCtx = ctx
+	c.serverLifecycleCtxCancelFunc = cancel
+}
+
+// ServerLifecycleContext returns the language server lifecycle context to be used by
+// background operations. It will be canceled when the language server shuts down.
+func (c *Config) ServerLifecycleContext() context.Context {
+	c.m.RLock()
+	defer c.m.RUnlock()
+	if c.serverLifecycleCtx == nil {
+		panic("server lifecycle context is not initialized")
+	}
+	return c.serverLifecycleCtx
+}
+
+func (c *Config) CancelServerLifecycleContext() {
+	c.m.RLock()
+	cancel := c.serverLifecycleCtxCancelFunc
+	c.m.RUnlock()
+	if cancel == nil {
+		panic("server lifecycle cancel func is not initialized")
+	}
+	cancel()
 }
 
 func (c *Config) EmptyToken() bool {
--- a/domain/snyk/scanner/scanner.go
+++ b/domain/snyk/scanner/scanner.go
@@ -298,7 +298,7 @@ func (sc *DelegatingConcurrentScanner) Scan(ctx context.Context, pathToScan type
 			go func() {
 				defer referenceBranchScanWaitGroup.Done()
 				isSingleFileScan := pathToScan != folderPath
-				scanTypeCtx := ctx2.NewContextWithDeltaScanType(ctx2.Clone(ctx, context.Background()), ctx2.Reference)
+				scanTypeCtx := ctx2.NewContextWithDeltaScanType(ctx, ctx2.Reference)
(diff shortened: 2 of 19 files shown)


bastiandoetsch (approved):

application/config/config.go
```diff
@@ -1556,6 +1558,40 @@ func (c *Config) SetLSPInitialized(initialized bool) {
...
+	c.serverLifecycleCtx = ctx
+	c.serverLifecycleCtxCancelFunc = cancel
+}
+
+// ServerLifecycleContext returns the language server lifecycle context to be used by
+// background operations. It will be canceled when the language server shuts down.
+func (c *Config) ServerLifecycleContext() context.Context {
+	c.m.RLock()
+	defer c.m.RUnlock()
+	if c.serverLifecycleCtx == nil {
```
bastiandoetsch, line 1579: [Should Fix] These accessors panic when the lifecycle context was never set, but `newConfig` doesn't default-initialize it — only `server.Start()` and the test helpers do. Production consumers reached deep in background paths (`internal/sdk/sdk.go`, `infrastructure/oss/cli_scanner.go:602 scheduleRefreshScan`) call `ServerLifecycleContext()` unconditionally, so any `Config` built via `config.New`/`NewFromExtension` that triggers a scan without going through `Start()` (an embedder, or a unit test bypassing the testutil helpers) hard-panics the whole server instead of degrading. Flagged by 3 of 4 reviewers.

Fix: default-initialize the field in `newConfig` with `context.WithCancel(context.Background())` (overwritten by `Start`), which lets you delete all three panic guards. (Storing a context on the long-lived shared `Config` singleton is also the `containedctx` anti-pattern — pragmatic here since Config is the universal handle, but worth a comment.)

— AI review

domain/snyk/scanner/scanner.go
```diff
@@ -298,7 +298,7 @@ func (sc *DelegatingConcurrentScanner) Scan(ctx context.Context, pathToScan type
 			go func() {
 				defer referenceBranchScanWaitGroup.Done()
 				isSingleFileScan := pathToScan != folderPath
-				scanTypeCtx := ctx2.NewContextWithDeltaScanType(ctx2.Clone(ctx, context.Background()), ctx2.Reference)
+				scanTypeCtx := ctx2.NewContextWithDeltaScanType(ctx, ctx2.Reference)
```
bastiandoetsch, line 301: [Should Fix · verify] Binding the reference-branch scan to the parent `ctx` (instead of the previous `Clone(ctx, context.Background())` decoupling) is the right call for shutdown cancellation, but it newly makes the base scan abortable by parent-context cancellation. The base scan does a git checkout (`scanBaseBranch` → `gitCheckoutHandler`); if `ctx` is cancelled mid-checkout the working tree could be left on the reference branch, which was previously impossible. Please confirm every `Scan` caller passes a lifecycle-bound ctx (not a per-message ctx that cancels on the next request), and that `scanBaseBranch` restores the branch on cancellation. (3 of 4 reviewers raised this; the WaitGroup makes the normal-completion path safe, the concern is the cancel path.)

— AI review

bastiandoetsch:
A couple more items from the verification pass (test files; not anchoring inline to keep it readable):

[Should Fix] `application/server/server_test.go` — `setupServer` registers the shutdown/exit `t.Cleanup` **before** `di.TestInit(t)`. `t.Cleanup` is LIFO, so the gomock controller's `Finish` (registered later) now runs **before** the shutdown/exit handlers, where the old `setupCustomServer` ran shutdown while mocks were still live. Any mock interaction or expectation reached on the shutdown path now risks an order-dependent 'missing call'/'controller finished' failure. Register the shutdown/exit cleanup after `di.TestInit` (or otherwise restore the original teardown order).

[Suggestion] `server_test.go` (~line 142) — `assert.NoError(t, exitCtx.Err(), "LS exit took too long")` runs right after a fire-and-forget `Notify`, which returns before the exit handler completes, so `exitCtx.Err()` is essentially always nil — the assertion can't catch a stuck exit. Consider waiting on an actual completion signal.

[Suggestion] Two files solve 'decouple but cancel on shutdown' with different idioms (`Clone(ctx, ServerLifecycleContext())` in cli_scanner.go vs using `ctx` directly in scanner.go) — worth a single shared helper.

Nice catch on the side: `infrastructure/cli/cli.go` — the refactor fixes a real latent bug, the old `Trace().Str("response", ...)` had no terminating `.Msg()`/`.Send()` and never emitted. Security scan: clean, and the GAF bump was verified to introduce zero new vulns (its go.mod hash is unchanged).

— AI review

domain/snyk/scanner/scanner.go
basti-snyk, line 301: **Worth confirming — semantic change to reference-branch scan cancellation.** Previously this cloned onto a fresh `context.Background()` (`ctx2.Clone(ctx, context.Background())`), deliberately decoupling the base/reference scan from the parent's cancellation so it always ran to completion and persisted its baseline snapshot. Deriving directly from `ctx` now places the reference scan in the parent's cancellation tree.

For the shutdown case this is the intended improvement (no detached goroutine). But there's a narrower regression to check: if the parent scan ctx is cancelled by a **token change mid-scan** (the auth-token-change cancel path), the in-flight base scan now aborts before `persistScanResults`, so no baseline is persisted and the next delta scan has no baseline — where the old decoupled clone guaranteed completion.

The goroutine lifetime itself is fine (bounded by `referenceBranchScanWaitGroup.Wait()`). Please confirm aborting the baseline on a token-change cancel is intended. Also worth a comment: `cli_scanner.go` clones its scheduled-scan base onto the lifecycle ctx, so the two scanners now use different base-ctx strategies — note it so they don't drift.

— AI review

PR #1180, opened 2026-03-16
feat: support per os settings from ldx-sync [IDE-1756]

andrewrobinsonhodges-snyk:
### Description

### Checklist


internal/types/ldx_sync_adapter.go
```diff
@@ -17,14 +17,47 @@
...
+
+func goosToSuffix(goos string) string {
+	var suffix string
+	switch goos {
+	case "windows":
+		suffix = "windows"
+	case "linux":
+		suffix = "linux"
+	default:
+		suffix = "macos"
```
bastiandoetsch: This should be Darwin

internal/types/ldx_sync_adapter.go
```diff
@@ -17,14 +17,47 @@
...
+	}
+	log.Debug().Str("goos", goos).Str("suffix", suffix).Msg("goosToSuffix - resolved OS suffix for per-OS settings")
```
bastiandoetsch: Most likely should be trace level, as this func will be called a lot :)

internal/types/ldx_sync_adapter.go
```diff
@@ -17,14 +17,47 @@
...
+	return suffix
+}
+
+// perOSSettings maps internal setting names to their base API field name.
+// The API sends these as <base>_<os>, e.g. "cli_path_macos".
+// Only the variant matching the current OS is accepted.
+var perOSSettings = map[string]string{
```
bastiandoetsch: Why make this a special case? We can just allow everything to be arch and os-specific.

andrewrobinsonhodges-snyk: Why would we add unnecessary complexity to the API and UIs?? I don't think end users will want different scan behavior on different OSs.

internal/types/ldx_sync_adapter.go
andrewrobinsonhodges-snyk: Does that help? As long as it matches the LDX-Sync value it shouldn't matter.

PR #1185, opened 2026-03-25
feat: settings HTML tabs

rrama:
### Description

Settings laid out in tabs. AI generated with some guidance, so probably removed something we cared about..

### Checklist

 - N/A
 - N/A (We should make the example page be generated though)
 - N/A
 - N/A


domain/ide/command/configuration_command.go
```diff
@@ -47,7 +47,7 @@ func (cmd *configurationCommand) Execute(ctx context.Context) (any, error) {
 	method := "configurationCommand.Execute"
 	cmd.logger.Debug().Str("method", method).Msg("executing configuration command")
 
-	settings := constructSettingsFromConfig(cmd.engine, cmd.configResolver)
+	settings := ConstructSettingsFromConfig(cmd.engine, cmd.configResolver)
```
acke, line 50: Nitpick: Exported for testing? Maybe add keep it internal and testing via Execute or a test-only helper.

infrastructure/configuration/template/config.html
```diff
@@ -21,536 +21,587 @@
...
+
+                            <div class="form-group hidden" id="token-field-group">
+                                <label for="token">Token <span class="config-scope-slot" data-config-scope-slot="true" data-setting-key="token"></span></label>
+                                <input type="password" id="token" name="token" value="{{.Settings.Token}}" placeholder="Snyk API Token">
+                                <span id="token-error" class="error-message hidden"></span>
+                            </div>
+
+                            <div class="form-group half-width">
+                                <label for="api_endpoint"><span data-toggle="tooltip" title="<p>If you're using SSO with Snyk and OAuth2, this is automatically populated.</p><p>For public regional instances, see Snyk documentation. For private instances, contact your team or account manager.</p>">Endpoint</span> <span class="config-scope-slot" data-config-scope-slot="true" data-setting-key="api_endpoint"></span></label>
+                                <input type="text" id="api_endpoint" name="api_endpoint" value="{{.Settings.Endpoint}}" placeholder="https://app.snyk.io/api">
```
acke: Is this the correct placeholder endpoint?
I would have expected https://api.snyk.io.

(I saw that we used https://app.snyk.io/api before as well.)

domain/ide/command/configuration_command.go
rrama: Exported for the `main.go` script IIRC.

infrastructure/configuration/template/config.html
```diff
@@ -21,536 +21,587 @@
...
-                                        Run only in reference folder
-                                    </label>
+                            <div class="form-group">
+                                <h3>Scanners</h3>
+                                <div class="checkbox-group">
+                                    <div class="override-indicator-wrapper {{if eq (getSource $folder.EffectiveConfig "snyk_oss_enabled") "user-override"}}has-user-override{{end}}">
+                                        <label class="checkbox-label {{if isLocked $folder.EffectiveConfig "snyk_oss_enabled"}}field-locked{{end}}">
+                                            <input type="checkbox" name="folder_{{$index}}_override_snyk_oss_enabled" data-setting="snyk_oss_enabled" {{with getEffectiveValue $folder.EffectiveConfig "snyk_oss_enabled"}}{{if .Value}}checked{{end}}{{end}} {{if isLocked $folder.EffectiveConfig "snyk_oss_enabled"}}disabled{{end}}>
```
acke, line 365: Question:

line 201 implementation for checkbox OSS scans

 <input type="checkbox" name="snyk_oss_enabled" {{if eq .Settings.ActivateSnykOpenSource "true"}}checked{{end}}>

differs from how it is implemented here for the same input type.

Why do we do the same thing differently? 

Could we implement this in the same way and reuse it?

infrastructure/configuration/template/config.html
```diff
@@ -21,536 +21,587 @@
...
-                                        Snyk Secrets
-                                        {{if isLocked $folder.EffectiveConfig "snyk_secrets_enabled"}}<span class="locked-indicator">Locked</span>{{end}}
-                                        <span class="source-badge {{getSourceClass $folder.EffectiveConfig "snyk_secrets_enabled"}}">{{getSourceLabel $folder.EffectiveConfig "snyk_secrets_enabled"}}</span>
+                            <div class="override-indicator-wrapper {{if eq (getSource $folder.EffectiveConfig "scan_net_new") "user-override"}}has-user-override{{end}}">
+                                <div class="form-group half-width {{if isLocked $folder.EffectiveConfig "scan_net_new"}}field-locked{{end}}">
+                                    <label for="folder_{{$index}}_override_scan_net_new">
+                                        <span data-toggle="tooltip" title="<p>Specifies whether to see all issues or only net new issues.</p><p>Net new issues requires a Git repository, where it compares findings with those in the base branch.</p>">Delta findings</span>
```
acke: Is this correct?

Net new issues requires a Git repository, I thought we also support folder to folder diff?

infrastructure/configuration/config_html.go
```diff
@@ -223,17 +233,40 @@ func tmplIsAutoScan(value any) bool {
...
+		return ""
+	}
+	val, exists := effectiveConfig[settingName]
+	if !exists {
+		return ""
+	}
+
+	switch val.Source {
+	case "ldx-sync-locked":
+		return template.HTML(`<span class="source-indicator" data-toggle="tooltip" title="Locked due to organization settings">🏢🔒</span>`)
```
acke, line 253: Did we move the icons from the style.css here, and if so why?

Line 585 in style.css had this
.locked-indicator::before {
    content: "🔒";
    margin-right: 4px;
}

acke (approved):
Some questions and nitpicks, nothing blocking.

infrastructure/configuration/config_html.go
rrama: I did because I wanted everything to have the same tooltip and I couldn't figure out how to do it in CSS for a particular class to all have the same HTML attribute added.

infrastructure/configuration/template/config.html
rrama: Because these can have source indicators and can be locked, whereas the global settings cannot be locked.
Although I think we are wanting to add the `has-user-override` stuff to the global settings as well, but that can be a next PR.

infrastructure/configuration/template/config.html
rrama: This was like this before my PR, but I will ping Matt to see.

rrama: Matt gave new wording, so I have updated it.

PR #1189, opened 2026-04-01
fix: pass trusted folders through didChangeConfiguration [IDE-1893]

nick-y-snyk:
### Description

The IDE-1786 refactoring moved `TrustedFolders` into `InitializationOptions` as init-only metadata, but on `main` it was updatable at runtime via `writeSettings`/`updateTrustedFolders`. This caused trusted folder changes sent via `workspace/didChangeConfiguration` to be silently dropped.

**Changes:**
- Add `TrustedFolders []string` field to `LspConfigurationParam` (matching the field on `InitializationOptions`)
- Wire `handlePushModel` and `handlePullModel` to call `applyTrustedFolders` with the new field
- Update push/pull model entry conditions to also check for `TrustedFolders`
- Add test for trusted folders via `didChangeConfiguration` push model
- Add `js-tests/fixtures/*.html` to `.gitignore` (generated test artifacts)

### Checklist


application/server/configuration.go
```diff
@@ -84,7 +84,7 @@ func workspaceDidChangeConfiguration(conf configuration.Configuration, srv *jrpc
 		engine, _ := ctx2.EngineFromContext(ctx)
 		defer logger.Info().Str("method", "WorkspaceDidChangeConfiguration").Msg("DONE")
 
-		if len(params.Settings.Settings) > 0 || len(params.Settings.FolderConfigs) > 0 {
+		if len(params.Settings.Settings) > 0 || len(params.Settings.FolderConfigs) > 0 || params.Settings.TrustedFolders != nil {
```
ShawkyZ, line 87: it's a string slice, check for len

ShawkyZ (approved):

PR #1194, opened 2026-04-07
fix: populate feature flags before auth notification [IDE-1901]

nick-y-snyk:
## Summary

- Fix race condition where scans after login fail with "Snyk Code is not enabled" because the IDE triggers a scan before SAST settings are cached
- `postCredentialUpdateHook` runs `populateAllFolderConfigs` between credential storage and auth notification, ensuring feature flags and SAST settings are available before the IDE reacts
- `sendFolderConfigs` made synchronous in login command

## Review feedback addressed

- Use unenriched folder config (no git enrichment needed for feature flag fetch)
- Loop only on trusted folders
- Remove unnecessary `Clone()` — `PopulateFolderConfig` only reads org from config
- Set `ConfigResolver` on folder config so `SetFeatureFlag` persists to GAF (was silently no-op without it)
- Case-insensitive error string matching in `shouldCauseLogout`
- Extract `isTransientNetworkError` helper to reduce cyclomatic complexity

## Test plan

- [x] `TestLoginCommand_Execute_FeatureFlagsPopulatedBeforeAuthNotification` — regression test verifying feature flags populate BEFORE `$/snyk.hasAuthenticated` notification (fails when hook is removed)
- [x] `Test_PostCredentialUpdateHook_CalledBeforeNotification` — hook runs between token storage and notification
- [x] `Test_shouldCauseLogout` — all 13 cases pass including expired refresh token and transient network errors
- [x] `TestIsAuthenticated_ConcurrentCallsSendOnlyOneNotification` — concurrent dedup still works
- [x] All existing `TestLoginCommand_*` tests pass
- [x] Lint clean, zero issues


domain/ide/command/login.go
```diff
@@ -116,7 +124,7 @@ func (cmd *loginCommand) Execute(ctx context.Context) (any, error) {
 		// Use context.Background() so this is not canceled if the LSP request context is
 		// canceled (e.g. when the IDE cancels the snyk.login request after auth completes).
 		cmd.ldxSyncService.RefreshConfigFromLdxSync(context.Background(), conf, cmd.engine, logger, config.GetWorkspace(conf).Folders(), cmd.notifier)
-		go sendFolderConfigs(conf, cmd.engine, logger, cmd.notifier, cmd.featureFlagService, cmd.configResolver)
+		sendFolderConfigs(conf, cmd.engine, logger, cmd.notifier, cmd.featureFlagService, cmd.configResolver)
```
acke: Ideally, if a user has large workspaces with 100+ projects, we want a separation between the login flow and the setup folders. With clear indications for the user that they are logged in, but snyk is not ready to scan until all configs are ready.

infrastructure/authentication/auth_service_impl.go
```diff
@@ -365,6 +374,13 @@ func shouldCauseLogout(err error, logger *zerolog.Logger) bool {
 	if errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF) {
 		return false
 	}
+	// Expired or revoked refresh tokens cause the OAuth2 token endpoint to reject
+	// the request. http.Client wraps this in url.Error, so detect it before the
+	// generic transport-error guard.
+	errMsg := err.Error()
```
acke: Could we user a re.Response.StatusCode == 401 or 400 to catch expired/invalid API credentials?

https://github.com/snyk/sweater-comb/blob/main/docs/standards/rest.md#401---unauthorized

Needs to be verified for both token an oauth.

infrastructure/authentication/auth_service_impl.go
```diff
@@ -365,6 +374,13 @@ func shouldCauseLogout(err error, logger *zerolog.Logger) bool {
 	if errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF) {
 		return false
 	}
+	// Expired or revoked refresh tokens cause the OAuth2 token endpoint to reject
```
acke: Should the change here be submitted as a separate PR for for changelog clarity?

domain/ide/command/folder_handler.go
```diff
@@ -46,12 +46,35 @@ const (
...
+	if featureFlagService == nil {
+		return
+	}
+	ws := config.GetWorkspace(conf)
+	if ws == nil {
+		return
+	}
+	log := logger.With().Str("method", "populateAllFolderConfigs").Logger()
+	for _, folder := range ws.Folders() {
+		fc, err := folderconfig.GetOrCreateFolderConfig(conf, folder.Path(), &log)
```
ShawkyZ: Use Unenriched variant

domain/ide/command/folder_handler.go
```diff
@@ -46,12 +46,35 @@ const (
...
+		if err != nil {
+			log.Err(err).Msg("unable to load folderConfig")
+			continue
+		}
+		if fc == nil {
+			continue
+		}
+		featureFlagService.PopulateFolderConfig(fc.Clone())
```
ShawkyZ: why the clone ?

domain/ide/command/folder_handler.go
```diff
@@ -46,12 +46,35 @@ const (
...
+func populateAllFolderConfigs(conf configuration.Configuration, engine workflow.Engine, logger *zerolog.Logger, featureFlagService featureflag.Service, configResolver types.ConfigResolverInterface) {
+	if featureFlagService == nil {
+		return
+	}
+	ws := config.GetWorkspace(conf)
+	if ws == nil {
+		return
+	}
+	log := logger.With().Str("method", "populateAllFolderConfigs").Logger()
+	for _, folder := range ws.Folders() {
```
ShawkyZ, line 65: loop on trusted folders

domain/ide/command/login.go
ShawkyZ: This can still run in a go routine, since it will get to this point after the postAuth hook

infrastructure/authentication/auth_service_impl.go
```diff
@@ -365,6 +374,13 @@ func shouldCauseLogout(err error, logger *zerolog.Logger) bool {
...
+	// the request. http.Client wraps this in url.Error, so detect it before the
+	// generic transport-error guard.
+	errMsg := err.Error()
+	if strings.Contains(errMsg, "oauth2/token") && strings.Contains(errMsg, "authentication failed") {
```
ShawkyZ: nitpick: make the comparison case insensitive

infrastructure/authentication/auth_service_impl.go
```diff
@@ -231,6 +238,17 @@ func (a *AuthenticationServiceImpl) updateCredentials(newToken string, sendNotif
 		a.notifDedup.Unlock()
 	}
 
+	if a.postCredentialUpdateHook != nil && newToken != "" {
+		func() {
+			defer func() {
+				if r := recover(); r != nil {
+					a.engine.GetLogger().Error().Interface("panic", r).Msg("postCredentialUpdateHook panicked")
```
ShawkyZ, line 245: why are you expecting it to panic?

nick-y-snyk: I didn't. just a double safety (ai suggestion to prevent ls crash). Do you want this removed?

nick-y-snyk: I also added an actual test which fails without other changes. But haven't tested manually yet

ShawkyZ: we have a global recovery that should catch it if it happens, I think this is not necessary here.

ShawkyZ (approved):

PR #1195, opened 2026-04-07
fix(ldx-sync): drop machine-scope dual-write + align readers/writers [IDE-1995]

andrewrobinsonhodges-snyk:
### Description

Three-step fix for machine-scope IDE-PATCH / LDX-Sync interaction:

#### Step X (commit cacd1545)
- LDX-Sync no longer dual-writes to `UserGlobalKey` — every machine value lands at `RemoteMachineKey` only.
- New `types.GetGlobalBool/String/Int` wrapper that walks the 4-phase machine chain (locked-remote → UserGlobal → unlocked-remote → default), mirroring GAF's `resolveMachine`.
- Fixes 3 documented bugs: MDM 2nd-sync drop, opposite-default never-apply, lock-removal stuck.

#### Step Y (commit 6fe22c24)
- Migrate ~12 production raw readers (`config.go`, `configuration.go`, `featureflag.go`, `sentry/init.go`, `command_service.go`) to `types.GetGlobalX`.
- Introduce `setGlobalUser` helper in `configuration.go`; rewrite IDE PATCH writers to wrap values in `*configresolver.LocalConfigField{Changed: true}`. Mirrors folder-scope `applyGenericFolderOverrides`.
- Cross-file PATCH writers migrated: `client_settings.go` (`SendErrorReports`), `config.go` (`ApiEndpoint`/`Organization`/`LastSetOrganization`), `apply_auth_config.go` (`AuthenticationMethod`).
- Add `IsLockedMachine` to `ConfigResolverInterface` + `validateLockedMachineFields` rejection in `processConfigSettings`. Mirrors folder-scope `validateLockedFields`. Without it, locked machine PATCH writes were silently shadowed by resolver phase 1 but persisted as ghost values at `UserGlobalKey`.
- Fix `GetGlobal*` wrapper to coerce `*LocalConfigField` (in-memory) and JSON-deserialized `map[string]any` (post-restart) at phase 2 via `util.CoerceToLocalConfigField`. Without this, every wrapper read of a newly-wrapped UserGlobal value would skip phase 2 and fall through to remote-unlocked or default.
- Add `Test_UpdateSettings_LockedMachineField_RejectsPATCH`.

#### Step Z (commit c9b9c823) — fills the four wrap gaps left by Y
- Wraps the four machine-scope IDE-PATCH writers that were still writing raw to `UserGlobalKey`, so every machine PATCH now lands as `*LocalConfigField{Changed: true}` (resolver phase 2 sees explicit user intent, no longer indistinguishable from a framework default):
  - `applyManageBinariesAutomatically` (`SettingAutomaticDownload`) — one of the two LDX-Sync routable settings flagged in the original Slack note.
  - `applyErrorReporting` (`SettingSendErrorReports`).
  - `applyTrustEnabledFromSettings` (`SettingTrustEnabled`).
  - `applyTrustedFolders` (`SettingTrustedFolders`) + the parallel `AddTrustedFolders` writer in `domain/ide/workspace/workspace.go`.
- Old-value diffs in those writers switched from raw `conf.GetBool(UserGlobalKey(...))` to `types.GetGlobalBool/SliceFilePath` so the comparison stays correct after the wrap.
- New `types.GetGlobalSliceFilePath` reader for the slice-typed `SettingTrustedFolders`. Globals are not persisted, so the helper only handles the in-memory shapes (raw `[]FilePath` or wrapped `*LocalConfigField`) — no JSON-deserialized form to coerce.
- Test reads of `SettingTrustedFolders` and `SettingAutomaticDownload` migrated to the wrapper helpers (raw seeds in tests left raw because the helpers accept both shapes).
- New `Test_UpdateSettings_MachineFields_PATCHWrapsAsLocalConfigField` asserts the wrap shape + `Changed=true` for all four writers.

#### Out of scope (separate ticket)

Folder-scope settings written via `UserGlobalKey` (9 sites: `SetSeverityFilterOnConfig`, `SetIssueViewOptionsOnConfig`, `applyAutoScan`, `applyDeltaFindings`, `applyProductEnablement`, `client_settings` Snyk*Enabled, `applyCliConfig` `CliAdditionalOssParameters`, `processInitMetadata` `ScanAutomatic`). Each fix needs folder-path threading through helper signatures (`SetSeverityFilterOnConfig` would gain a folder arg, etc.). Not in this PR — blast radius too wide.

`SettingToken` writes left raw — `GetToken` reads via raw `conf.GetString` and would break under wrapping.

`processInitMetadata` raw writes (`SettingDeviceId`, `SettingOsArch`, `SettingOsPlatform`, `SettingRuntimeName`, `SettingRuntimeVersion`, `SettingHoverVerbosity`, `SettingFormat`, `SettingClientProtocolVersion`) — init-only metadata; not subject to LDX-Sync, so the wrap distinction has no effect today. Cleanup deferred for symmetry.

### Checklist

- [x] Tests added and all succeed (locked-machine rejection + reader/writer migrations + Step Z wrap-shape assertion)
- [x] Regenerated mocks (`go generate ./internal/types/...` — ConfigResolverInterface added `IsLockedMachine`)
- [x] Linted (`make lint-fix` — 0 issues)


application/server/configuration.go
```diff
@@ -172,7 +172,10 @@ func processInitMetadata(conf configuration.Configuration, engine workflow.Engin
 
 	applyPathToEnv(conf, logger, opts.Path)
 
-	// Auto scan true by default unless explicitly disabled
+	// Auto scan true by default unless explicitly disabled.
+	// SettingScanAutomatic is folder-scoped per register_configurations.go but
+	// historically written here at UserGlobalKey. Tracked separately; left raw to
```
andrewrobinsonhodges-snyk: Where is this tracked? Might be good to put a link in the comment.

application/server/configuration.go
```diff
@@ -237,6 +240,41 @@ func refreshLdxSyncOnTokenChange(conf configuration.Configuration, engine workfl
 	di.LdxSyncService().RefreshConfigFromLdxSync(context.Background(), conf, engine, logger, folders, di.Notifier())
 }
 
+// Wrap is required so the resolver's phase 2 distinguishes user intent from a framework
```
andrewrobinsonhodges-snyk: Can we describe "phase 2" etc? Would make the comments more useful.

application/server/configuration.go
```diff
@@ -237,6 +240,41 @@ func refreshLdxSyncOnTokenChange(conf configuration.Configuration, engine workfl
...
+func setGlobalUser(conf configuration.Configuration, name string, value any) {
+	conf.Set(configresolver.UserGlobalKey(name), &configresolver.LocalConfigField{
+		Value:   value,
+		Changed: true,
+	})
+}
+
+// Without rejection, locked PATCHes are silently shadowed by phase 1 but persist as ghost
+// entries at UserGlobalKey — load-bearing if the admin later lifts the lock.
+func validateLockedMachineFields(settings map[string]*types.ConfigSetting, configResolver types.ConfigResolverInterface, fm workflow.ConfigurationOptionsMetaData, subLogger *zerolog.Logger) bool {
```
andrewrobinsonhodges-snyk, line 251: Where did we land on the expected behaviour here? Machine fields can always be overwritten by MDM, so we need to make sure that we're not breaking that.

application/server/configuration.go
```diff
@@ -677,14 +725,17 @@ func applyOrganization(conf configuration.Configuration, engine workflow.Engine,
 
 func applyCliConfig(conf configuration.Configuration, settings map[string]*types.ConfigSetting) {
 	if v, ok := settingBool(settings, types.SettingProxyInsecure); ok {
-		conf.Set(configresolver.UserGlobalKey(types.SettingProxyInsecure), v)
+		setGlobalUser(conf, types.SettingProxyInsecure, v)
 		conf.Set(configuration.INSECURE_HTTPS, v)
 	}
 	if v, ok := settingStr(settings, types.SettingAdditionalParameters); ok {
+		// SettingCliAdditionalOssParameters is folder-scoped per register_configurations.go but
+		// historically written here at UserGlobalKey. Tracked separately; left raw to keep wire
```
andrewrobinsonhodges-snyk: As above - where is this tracked?

application/server/configuration_test.go
```diff
@@ -1807,3 +1807,73 @@ func TestApplyIssueViewOptions_NeitherChangedIsNoOp(t *testing.T) {
...
+		"locked CodeEndpoint must keep remote value")
+
+	// UserGlobalKey was never written — apply* never received the entry.
+	assert.False(t, conf.IsSet(configresolver.UserGlobalKey(types.SettingPublishSecurityAtInceptionRules)),
+		"PATCH for locked machine setting must not land at UserGlobalKey")
+	assert.False(t, conf.IsSet(configresolver.UserGlobalKey(types.SettingCodeEndpoint)),
+		"PATCH for locked machine setting must not land at UserGlobalKey")
+}
+
+// PATCH writes must land as *LocalConfigField{Changed: true}; otherwise resolver phase 2
```
andrewrobinsonhodges-snyk: As above - unclear what "phase 2" is.

internal/types/config_readers.go
```diff
@@ -24,6 +24,23 @@ import (
...
+	if v == nil {
+		return nil, false
+	}
+	if lf, ok := v.(*configresolver.LocalConfigField); ok {
+		if lf == nil || !lf.Changed {
+			return nil, false
+		}
+		return lf.Value, true
+	}
+	return v, true
```
andrewrobinsonhodges-snyk, line 50: Why are we returning changed = true if the resolver cannot find a value? Might be correct, but worth adding a comment, as it is not obvious.

internal/types/config_readers.go
```diff
@@ -24,6 +24,23 @@ import (
...
+func userGlobalValue(conf configuration.Configuration, key string) (any, bool) {
+	v := conf.Get(configresolver.UserGlobalKey(key))
+	if v == nil {
+		return nil, false
+	}
+	if lf, ok := v.(*configresolver.LocalConfigField); ok {
+		if lf == nil || !lf.Changed {
+			return nil, false
+		}
+		return lf.Value, true
```
andrewrobinsonhodges-snyk: Should we use lf.Changed here instead of true (or put this return in an else statement)? It makes the code a bit more resilient to refactoring in the future.

internal/types/config_readers.go
```diff
@@ -44,38 +61,101 @@ func GetGlobalOrganization(conf configuration.Configuration) string {
...
+	intFrom := func(v any) (int, bool) {
 		switch i := v.(type) {
 		case int:
-			return i
+			return i, true
 		case int64:
-			return int(i)
+			return int(i), true
+		}
+		return 0, false
```
andrewrobinsonhodges-snyk, line 160: Is it safe to assume a default value? what do we use this for?

application/server/configuration.go
nick-y-snyk: Tracked at IDE-1996 (Folder-scope settings written via UserGlobalKey instead of UserFolderKey). Replaced the inline TODO with `types.SetGlobalDeferredFolderScope` whose doc references the ticket — every deferred-scope callsite is now explicit at the call and discoverable from the helper.

application/server/configuration.go
nick-y-snyk: Good call. Pulled the helper out into `types.SetGlobalUser` (`internal/types/config_writers.go`) and pointed its doc at `types.GetGlobalBool`, which already enumerates the 4-phase chain. Avoided re-describing it locally so the phase numbering has one canonical source.

application/server/configuration.go
nick-y-snyk: MDM/locked-remote writes are not affected — the lock arrives via LDX-Sync sync into `RemoteMachineKey`, which this function never touches. We only reject *user PATCH* attempts against already-locked machine settings (phase 2 trying to overwrite phase 1, which would be silently shadowed today but persist as a ghost value if the lock is later lifted). The `IsFolderScopedSetting` guard keeps the loop machine-only. Spelled all of that out in the doc.

application/server/configuration.go
nick-y-snyk: Same as the `SettingScanAutomatic` one — tracked at IDE-1996. Migrated to `types.SetGlobalDeferredFolderScope`; the helper doc references the ticket.

application/server/configuration_test.go
nick-y-snyk: Updated to point at `types.GetGlobalBool` for the phase numbering — thats the canonical 4-phase enumeration.

internal/types/config_readers.go
nick-y-snyk: Done — returns `(lf.Value, lf.Changed)` now.

internal/types/config_readers.go
nick-y-snyk: Reworked the doc to enumerate both shapes at `UserGlobalKey` (wrapped vs raw — see `internal/types/config_writers.go`) and clarify that the bool reports presence, not user-intent. The `(nil, false)` early return already covers the absent / `Changed=false` cases; the resolver-chain caller decides whether to use the value or fall through.

internal/types/config_readers.go
nick-y-snyk: Added a doc explaining that `(0, false)` is the type-mismatch sentinel — the caller falls through the resolver chain, so the zero never reaches the user (phase 4 returns the flagset default). Also added a `float64` case defensively, in case JSON-deserialized framework defaults ever land here.

nick-y-snyk (approved):

PR #1196, opened 2026-04-07
fix: persist and restore user folder config across restarts [IDE-1816]

nick-y-snyk:
### Description

Fixes folder-level user settings (additional params, env, org, base branch, etc.) not surviving LS restarts after the folder config refactoring.

**Root cause:** After the refactoring moved folder config values from a monolithic `StoredConfig` JSON blob to per-key GAF prefix keys (`user:folder:<path>:<name>`), the values were correctly written to the JSON storage file on save, but never loaded back on startup. Additionally, after JSON round-trip, `*configresolver.LocalConfigField` deserializes as `map[string]interface{}` and `[]string` becomes `[]interface{}`, causing type assertion failures.

**Changes:**
- **`RefreshByPrefix` in storage** — loads all keys matching a prefix from the JSON storage file into configuration on startup
- **`SetupStorage` in config** — calls `RefreshByPrefix` for `user:folder:` prefix to restore persisted folder overrides
- **`coerceToLocalConfigField`** — handles both in-memory `*LocalConfigField` and deserialized `map[string]interface{}` representations
- **`ReadFolderConfigSnapshot`** — handles `[]interface{}` → `[]string` conversion for `AdditionalParameters` and `LocalBranches`
- **`getSettingValue` / `getStringSliceFromSetting` / `getScanCommandConfigFromSetting`** — require `Changed: true` flag before applying updates
- **Tests** — unit tests for coercion, round-trip integration tests for string/bool/slice persistence

**Known issue:** `org_set_by_user` / `preferred_org` still gets reset on restart despite being correctly persisted and loaded. Investigation ongoing — the LS should own the `Changed` determination rather than relying on 4 different IDEs to set it correctly.

### Checklist


internal/storage/storage.go
```diff
@@ -69,6 +71,34 @@ func (s *storage) Refresh(config configuration.Configuration, key string) error
...
+	s.mutex.Unlock()
+	for key, value := range doc {
+		if strings.HasPrefix(key, prefix) {
+			config.PersistInStorage(key)
+			config.Set(key, value)
+		}
+	}
+	return nil
+}
+
```
ShawkyZ: instead of doing this, use the already registered keys in https://github.com/snyk/snyk-ls/blob/5e953c4c82bef8e7ded3390839e10e4e6dedb537/internal/types/register_configurations.go#L28

When you register a key you can mark it as persistable.

internal/types/folder_config_helpers.go
```diff
@@ -40,14 +40,31 @@ type FolderConfigSnapshot struct {
...
 func getUserFolderValue(conf configuration.Configuration, fp string, name string) (any, bool) {
 	key := configresolver.UserFolderKey(fp, name)
 	val := conf.Get(key)
 	if val == nil {
 		return nil, false
 	}
-	lf, ok := val.(*configresolver.LocalConfigField)
-	if !ok || lf == nil || !lf.Changed {
+	lf, ok := coerceToLocalConfigField(val)
+	if !ok {
```
ShawkyZ, line 67: why can't we directly cast?

application/server/configuration.go
```diff
@@ -614,7 +612,7 @@ func applyOrganization(conf configuration.Configuration, engine workflow.Engine,
 
 func applyCliConfig(conf configuration.Configuration, settings map[string]*types.ConfigSetting) {
 	if v, ok := settingBool(settings, types.SettingProxyInsecure); ok {
-		conf.Set(configresolver.UserGlobalKey(types.SettingCliInsecure), v)
+		conf.Set(configresolver.UserGlobalKey(types.SettingProxyInsecure), v)
 		conf.Set(configuration.INSECURE_HTTPS, v)
```
ShawkyZ, line 620: what is the difference between `INSECURE_HTTPS` and `SettingProxyInsecure`

domain/ide/command/configuration_command.go
```diff
@@ -95,7 +95,7 @@ func constructSettingsFromConfig(engine workflow.Engine, r types.ConfigResolverI
 		Path:                        conf.GetString("PATH"),
 		ManageBinariesAutomatically: fmt.Sprintf("%v", r.GetBool(types.SettingAutomaticDownload, nil)),
 		AdditionalParams:            additionalOssParams,
-		Insecure:                    fmt.Sprintf("%v", r.GetBool(types.SettingCliInsecure, nil)),
+		Insecure:                    fmt.Sprintf("%v", r.GetBool(types.SettingProxyInsecure, nil)),
```
ShawkyZ, line 98: did we never use `SettingCliInsecure` before?

application/server/configuration.go
bastiandoetsch: I don't think there's a diff

internal/types/folder_config.go
```diff
@@ -371,7 +372,7 @@ func (fc *FolderConfig) applyFolderScopeUpdates(update *LspFolderConfig) bool {
 		return changed
 	}
 	for name, cs := range update.Settings {
-		if handled[name] || cs == nil || !cs.Changed {
+		if handled[name] || cs == nil {
```
ShawkyZ, line 375: why?

application/server/configuration.go
```diff
@@ -291,8 +291,6 @@ func processFolderConfigs(conf configuration.Configuration, engine workflow.Engi
 
 	var processedConfigs []types.FolderConfig
 	var changedConfigs []*types.FolderConfig
-	// Always notify when the client explicitly sends folder configs — it expects the resolved state back.
-	needsToSendUpdateToClient := len(incomingMap) > 0
```
ShawkyZ, line 295: I think this change makes sense to prevent infinite round trips between IDEs and LS, tests still need to be fixed tho.

ShawkyZ (approved):

PR #1210, opened 2026-04-14
feat: folder config migration from legacy JSON to GAF keys [IDE-1905]

nick-y-snyk:
### Description

Migrates folder config from old single-JSON-blob format (`INTERNAL_LS_CONFIG`) to individual GAF prefix keys, and fixes temp folder config pollution from delta base scans.

**Problem 1 — Data loss on upgrade**: The refactoring in IDE-1786 changed folder config storage from a monolithic JSON string to individual GAF keys (`user:folder:<path>:<setting>`). Without migration, users upgrading lose `preferredOrg`, `orgSetByUser`, `additionalParameters`, `baseBranch`, and other folder settings. Worst case: scans silently run against the wrong org.

**Problem 2 — Config file pollution**: `CopyFolderConfigValues()` called `PersistInStorage()` for every key copied to temp delta scan directories, writing 500+ junk entries to the config file that are never cleaned up.

#### Changes

**Commit 1 — Fix temp folder pollution**
- Remove `PersistInStorage()` calls from `CopyFolderConfigValues()` in `folder_config_helpers.go`
- Copied values remain in-memory via `conf.Set()` for scan execution
- Add 2 tests verifying no persistence to disk + correct in-memory copy

**Commit 2 — Legacy config migration**
- New `internal/folderconfig/migration.go`: parses old `INTERNAL_LS_CONFIG` JSON, writes fields as individual GAF keys using existing helpers, clears old key
- Migrated fields: `baseBranch`, `additionalParameters`, `additionalEnv`, `referenceFolderPath`, `preferredOrg`, `orgSetByUser`, `autoDeterminedOrg`
- Not migrated (re-fetched or dead): `localBranches`, `featureFlags`, `sastSettings`, `scanCommandConfig`, `orgMigratedFromGlobalConfig`
- Runs once in `SetupStorage()` after `Refresh()` loads old data
- 7 tests: table-driven with `t.Parallel()`, using JSON fixture files (`testdata/*_input.json` → `*_expected.json`) for actual file-level comparison of storage before/after migration

### Checklist

### DIFF
--- a/internal/folderconfig/migration.go
+++ b/internal/folderconfig/migration.go
@@ -1,0 +1,101 @@
+/*
+ * © 2022-2026 Snyk Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package folderconfig
+
+import (
+	"encoding/json"
+
+	"github.com/rs/zerolog"
+	"github.com/snyk/go-application-framework/pkg/configuration"
+
+	"github.com/snyk/snyk-ls/internal/types"
+)
+
+// legacyFolderConfig represents the old FolderConfig struct fields needed for migration.
+// Only fields that are migrated to individual GAF keys are included.
+type legacyFolderConfig struct {
+	BaseBranch           string         `json:"baseBranch"`
+	AdditionalParameters []string       `json:"additionalParameters,omitempty"`
+	AdditionalEnv        string         `json:"additionalEnv,omitempty"`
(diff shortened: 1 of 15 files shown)


internal/types/folder_config_helpers.go
```diff
@@ -191,7 +191,6 @@ func CopyFolderConfigValues(conf configuration.Configuration, srcPath, dstPath F
 	for _, name := range userSettings {
 		if v := conf.Get(configresolver.UserFolderKey(src, name)); v != nil {
 			key := configresolver.UserFolderKey(dst, name)
-			conf.PersistInStorage(key)
```
ShawkyZ, line 194: nice

ShawkyZ (approved):

internal/folderconfig/migration.go
```diff
@@ -0,0 +1,101 @@
...
+
+	oldData := conf.GetString(ConfigMainKey)
+	if oldData == "" {
+		return
+	}
+
+	var sc legacyStoredConfig
+	if err := json.Unmarshal([]byte(oldData), &sc); err != nil {
+		l.Err(err).Msg("failed to parse legacy folder config, clearing key to prevent re-migration")
+		conf.Set(ConfigMainKey, "")
```
ShawkyZ: conf.Unset would be better to delete it from the config

ShawkyZ: also let's not log with Error level but Debug since we will have migration only once.

PR #1214, opened 2026-04-15
fix: post auth hook locking

ShawkyZ:
### Description

**Fix: move postCredentialUpdateHook execution outside the auth lock**

The `postCredentialUpdateHook` (used to populate feature flags after login) was running inside `updateCredentials`, which is called from `authenticate()` while `a.m.Lock()` is held. The hook makes HTTP calls to fetch feature flags and SAST settings for every workspace folder. This means:

- The auth service's RWMutex write lock was held during multiple network round-trips
- All concurrent `IsAuthenticated()` callers (which need `RLock`) were blocked
- Any code in the hook's call chain that transitively called back into the auth service would **deadlock**

**Fix:** Remove the hook call from `updateCredentials` and move it to `Authenticate()`, where it runs **after** the lock is released. The hook is still injectable from outside via `SetPostCredentialUpdateHook` — the only change is *when* it executes relative to the lock.

Before:
```
Authenticate() {
    a.m.Lock()
    └─ authenticate()
       └─ updateCredentials()
          ├─ store credentials
          ├─ postCredentialUpdateHook()  ← network I/O under lock
          └─ send notification
    a.m.Unlock()
}
```

After:
```
Authenticate() {
    a.m.Lock()
    └─ authenticate()
       └─ updateCredentials()
          ├─ store credentials
          └─ send notification
    a.m.Unlock()
    postCredentialUpdateHook()  ← network I/O, NO lock held
}
```

The test is updated to verify the hook runs outside the lock by calling `IsAuthenticated()` from within the hook — which would deadlock under the old code.

### Checklist


infrastructure/authentication/auth_service_impl.go
```diff
@@ -238,17 +243,6 @@ func (a *AuthenticationServiceImpl) updateCredentials(newToken string, sendNotif
 		a.notifDedup.Unlock()
 	}
 
-	if a.postCredentialUpdateHook != nil && newToken != "" {
```
bastiandoetsch, line 241: why did you remove it?

infrastructure/authentication/auth_service_impl.go
```diff
@@ -111,14 +111,19 @@ func (a *AuthenticationServiceImpl) Authenticate(ctx context.Context) (token str
...
 	ctx, a.previousAuthCtxCancelFunc = context.WithCancel(ctx)
 	a.previousAuthCtxCancelFuncMu.Unlock()
 
-	defer a.previousAuthCtxCancelFunc() // need to clean up resources if we weren't interrupted, impl should ensure its safe to double call
-	return a.authenticate(ctx)
+	token, err = a.authenticate(ctx)
+	a.previousAuthCtxCancelFunc()
+	a.m.Unlock()
+
+	if a.postCredentialUpdateHook != nil && token != "" && err == nil {
```
bastiandoetsch: why is this now extracted?

bastiandoetsch (approved):

infrastructure/authentication/auth_service_impl.go
ShawkyZ: don't want the post hook to be inside the mutex lock, since post hook will trigger a bunch of network calls for fetching FFs. This will affect other calls like `IsAuthenticated`

infrastructure/authentication/auth_service_impl.go
ShawkyZ: this was added by a previous PR and wasn't necessary since we have the recovery on a global level.

PR #1217, opened 2026-04-15
fix: handle case where values sent from ide are unchanged but non-nil

andrewrobinsonhodges-snyk:
### Description

### Checklist


rrama (approved):
Seems legit. I was there when it was written.

internal/types/config_resolver_test.go
```diff
@@ -1049,6 +1049,52 @@ func TestFolderConfig_ApplyLspUpdate(t *testing.T) {
...
+			Settings: map[string]*types.ConfigSetting{
+				types.SettingScanAutomatic: {Value: nil, Changed: false},
+			},
+		}
+
+		changed := fc.ApplyLspUpdate(update)
+
+		assert.False(t, changed, "Changed: false entry should be ignored even with nil Value")
+		assert.True(t, types.HasUserOverride(fc.Conf(), fc.FolderPath, types.SettingScanAutomatic),
+			"ScanAutomatic should remain when Changed is false")
```
bastiandoetsch, line 1095: how do we reset then?

PR #1219, opened 2026-04-15
Revert "fix: handle case where values sent from ide are unchanged but…

andrewrobinsonhodges-snyk:
… non-nil (#1217)"

This reverts commit de974b3400439aa025ba6fbdcf1e56216b976ef2.

### Description

### Checklist


rrama (approved):
This is even better without it.

PR #1222, opened 2026-04-17
refactor: rename missed severity setting namings [IDE-1816]

rrama:
### Description

It has confused me too many times and I wasn't sure if it was a cause of bugs.

### Checklist

 - None added, just updated.
 - N/A
 - N/A


rrama:
Having done this rename, I don't think it is required to fix any bugs. I think the code, while confusing, was at least functioning correctly. Since we are getting close to release, I think we will leave these for now.

PR #1223, opened 2026-04-17
refactor(config): remove Settings struct + JS snapshot test coverage [IDE-1998]

nick-y-snyk:
## Summary

**Refactor**
- Replace `types.Settings` struct (string-encoded booleans, hand-rolled JSON tags) with `map[string]any` keyed by registered pflag names so HTML form `name=` attributes match the LS wire protocol.
- New `TestConfigHtml_FormFieldNamesMatchRegisteredSettings` enforces the form-field ↔ pflag invariant at build time.
- HTML form field names renamed to match the wire payload directly: `enabled_severities_*` → `severity_filter_*`, drop the `_override_` infix on per-folder fields, etc. Previously these went through `processFolderOverrides` / `processFilterSeverity` / `processIssueViewOptions` translation shims in `form-handler.js`. Shims now deleted — flat names flow straight from template to the `folderConfigs` payload.
- Bool coercion in `setFieldValue` now opt-in via `data-bool="1"` on each select / hidden input that carries `"true"`/`"false"` text (4 selects + 1 hidden in `config.html`). Previously inferred from element type — which silently coerced any user-typed `"true"` in a text input to a boolean.

**Bug fixes**
- `risk_score_threshold` value=0 is now preserved in HTML rendering (template guard was stripping the value when it equalled the zero int).
- Per-folder `scan_automatic` select now sends a real bool to LS (was sending the literal strings `"auto"`/`"manual"` via `isAutoScan` helper, now deleted).
- `integration_name` smuggle-via-Settings-map removed; renderer reads `INTEGRATION_ENVIRONMENT` directly from GAF config.

## Wire-payload diff vs PR base

Verified by running `formHandler.collectData()` on the rendered fixture at the PR base (`df2b4f1a`) and at HEAD, then diffing the stable-stringified JSON:

```
diff base-payload.json HEAD/snapshots/form-payload.json
55c55
<       "scan_automatic": false,
---
>       "scan_automatic": true,
```

182 lines, 1 line differs (a dummy-fixture value, not a contract change). **The IDE↔LS settings save payload is byte-identical between base and HEAD.** What changed is the *intermediate representation*: HTML form field names + the JS shims that previously translated those names into payload keys. The shims were deleted; HTML names now equal payload names. Net wire contract preserved.

Concrete intermediate-rep changes (visible in template + Go test diffs but invisible to the IDE):
- HTML `name="enabled_severities_critical"` → `name="severity_filter_critical"` (and `_high`, `_medium`, `_low`)
- HTML per-folder `name="folder_X_override_<setting>"` → `name="folder_X_<setting>"`
- `form-handler.js` lost `processFilterSeverity` / `processFolderOverrides` / `processIssueViewOptions`

## JS test additions
- `js-tests/form-payload-snapshot.test.mjs` — locks full `collectData()` output against checked-in JSON; future field renames, type-coercion shifts, structural changes will fail CI unless `UPDATE_SNAPSHOT=1` is run.
- `js-tests/validation-wiring.test.mjs` — spies on `validation.validateAndShowError` and asserts every listener wired by `initializeAllValidation` (5 global + 2 per-folder suffix patterns) fires when its event is dispatched. Catches template ↔ JS selector drift like the dropped-`_override_`-prefix bug surfaced during review.

## Test plan
- [ ] `make test` — go unit tests + JS tests + ES5 lint
- [ ] Open Settings dialog, toggle each global field, confirm save payload uses pflag-named keys
- [ ] Per-folder: change scan_automatic / severity_filter_*, confirm save payload contains `folder_INDEX_<setting>` flat fields (no `override_`) and `scan_automatic` is a bool
- [ ] Set global `risk_score_threshold` to 0, confirm value persists across reload


infrastructure/configuration/template/js/ui/form-handler.js
```diff
@@ -196,16 +188,11 @@
 		if (el.type === "checkbox") {
 			obj[field] = el.checked;
 		} else if (el.type === "number") {
-			obj[field] = el.value ? parseInt(el.value) : null;
+			obj[field] = el.value ? parseInt(el.value, 10) : null;
+		} else if (el.dataset && el.dataset.bool === "1") {
```
rrama, line 192: Is there really not a better way to tell if it is a checkbox and then use `.checked`?

infrastructure/configuration/template/config.html
```diff
@@ -324,17 +324,17 @@ <h2>Filters and views</h2>
                             <div class="override-indicator-wrapper {{getSourceClass .DefaultsScopes "risk_score_threshold"}}">
                                 <div class="form-group half-width">
                                     <label for="risk_score_threshold"><span data-toggle="tooltip" title="<p>Minimum risk score to display issues.</p><p>Valid range: 0 to 1000 (0 = show all issues).</p>">Filter by risk score<span class="deprecated-label">Closed Beta</span></span> <span class="config-scope-slot" data-config-scope-slot="true" data-setting-key="risk_score_threshold"></span></label>
-                                    <input type="number" id="risk_score_threshold" name="risk_score_threshold" min="0" max="1000" value="{{if ne .Settings.RiskScoreThreshold nil}}{{.Settings.RiskScoreThreshold}}{{end}}" placeholder="0">
+                                    <input type="number" id="risk_score_threshold" name="risk_score_threshold" min="0" max="1000" value="{{$rst := index .Settings "risk_score_threshold"}}{{if ne $rst nil}}{{$rst}}{{end}}" placeholder="0">
```
rrama, line 327: Not saying it is your doing, as it was probably mine, but why are we using 2 different techniques for the risk score threshold? (Global vs folder level)

domain/ide/command/configuration_command_settings_test.go
```diff
@@ -27,178 +27,99 @@ func TestConstructSettingsFromConfig_AllFieldsPopulated(t *testing.T) {
...
-	})
 	engine.GetConfiguration().Set(configresolver.UserGlobalKey(types.SettingAuthenticationMethod), string(types.TokenAuthentication))
-	conf.Set(configresolver.UserGlobalKey(types.SettingEnableSnykLearnCodeActions), true)
-	conf.Set(configresolver.UserGlobalKey(types.SettingEnableSnykOssQuickFixActions), true)
 	engine.GetConfiguration().Set(configresolver.UserGlobalKey(types.SettingScanNetNew), true)
-	conf.Set(configresolver.UserGlobalKey(types.SettingEnableSnykOpenBrowserActions), true)
-	conf.Set(configresolver.UserGlobalKey(types.SettingOsPlatform), "darwin")
-	conf.Set(configresolver.UserGlobalKey(types.SettingOsArch), "arm64")
-	conf.Set(configresolver.UserGlobalKey(types.SettingRuntimeName), "node")
-	conf.Set(configresolver.UserGlobalKey(types.SettingRuntimeVersion), "18.0.0")
```
rrama, line 45: Any reason for removing these or just because they aren't written in the HTML settings page?

domain/ide/command/configuration_command_settings_test.go
```diff
@@ -27,178 +27,99 @@ func TestConstructSettingsFromConfig_AllFieldsPopulated(t *testing.T) {
...
-		assert.Equal(t, "true", settings.EnableTrustedFoldersFeature, "EnableTrustedFoldersFeature should be true")
-		require.NotNil(t, settings.TrustedFolders, "TrustedFolders should be initialized")
-		require.Len(t, settings.TrustedFolders, 1, "TrustedFolders should contain one folder")
-		assert.Equal(t, "/Users/test/trusted-project", settings.TrustedFolders[0], "Trusted folder path should match")
+		tf, ok := settings[types.SettingTrustedFolders].([]string)
+		require.True(t, ok)
+		assert.Empty(t, tf)
 	})
 
 	t.Run("Multiple trusted folders", func(t *testing.T) {
```
rrama: Nitpick: If there is now only one sub-test now, may as well make it not be a sub-test.

rrama (approved):
LGTM. One nitpick.

domain/ide/command/configuration_command_settings_test.go
nick-y-snyk: yes

infrastructure/configuration/template/js/ui/form-handler.js
nick-y-snyk: The code already uses `el.type === "checkbox"` → `el.checked` (lines 188–189) — that's the standard DOM way. The line you're pinned to (192) is the separate `data-bool` branch for `<select>` elements. No change needed here.

infrastructure/configuration/template/config.html
nick-y-snyk: Fixed — folder-level now matches global: empty value + `placeholder="0"` instead of hardcoding `0` as the actual value. 4cbbf690

domain/ide/command/configuration_command_settings_test.go
nick-y-snyk: Done — split into `TestConstructSettingsFromConfig_TrustedFolders_Empty` and `TestConstructSettingsFromConfig_TrustedFolders_Multiple`. 4cbbf690

PR #1224, opened 2026-04-19
feat: hiding HTML sections by default [IDE-1762]

rrama:
### Description

Very much vibe coded, so probably very rough, but the core logic seemed fine.

### Checklist

 - N/A
 - N/A


ShawkyZ (changes requested):
The new FF is added to the new FF service. 
We should to use the new FF gateway in GAF to resolve it.

rrama:
We may take some of this, but certainly not all, so I am closing it.

PR #1227, opened 2026-04-24
fix(config): respect SNYK_LOG_LEVEL env var in tests and runtime

nick-y-snyk:
## Summary
Two related fixes for log-level handling that conspired to silently downgrade test output to INFO regardless of what the developer requested.

### 1. `application/config/config.go` — `SetupLogging`
Inverted check: the error message for `SNYK_LOG_LEVEL` parse failure was guarded by `if levelErr == nil`, so it printed `Can't set log level from flag. Setting to default (=info)` on parse SUCCESS. Cosmetic only (level itself was applied), but masked invalid env values which were silently ignored. Flipped to warn on actual parse failure. Bug introduced in fc922733 (HEAD-317, 2023-06-01).

### 2. `application/server/server_smoke_test.go` — `ensureInitialized`
Unconditionally forced INFO via `config.SetLogLevel(LevelInfoValue)` + `t.Setenv("SNYK_LOG_LEVEL", config.GetLogLevel())`, clobbering any developer-set value. Result: pre-init DBG lines visible, post-init silent at INFO. Now only force INFO when env var is unset, so `SNYK_LOG_LEVEL=debug go test ...` produces full debug output end-to-end.

## Test plan
- [ ] Run `SNYK_LOG_LEVEL=debug go test ./application/server -run Test_SmokeLdxSync_Initialize` and confirm DBG lines appear after `ensureInitialized`.
- [ ] Run with `SNYK_LOG_LEVEL=garbage` and confirm new warning `Can't parse SNYK_LOG_LEVEL. Keeping current level.` appears.
- [ ] Default run (no env var) keeps INFO as before.
- [ ] `make lint` passes (verified locally).


rrama (approved):
I had to think quite a bit about whether this would increase log spam in CI/CD. I think I have convinced myself that it will not, as CI/CD does not set the env var, so would not still execute the body of the `if` and set the level to info.

PR #1228, opened 2026-04-27
fix(server): preserve unchanged issue view options on partial payloads [IDE-1946]

bastiandoetsch:
## Summary

Fixes a bug in `applyIssueViewOptions` where a partial `workspace/didChangeConfiguration` payload would silently overwrite the unchanged issue view flag with its zero value, causing the IntelliJ plugin's tree view to empty when toggling 'Show Ignored Issues'.

## Root cause

`applyIssueViewOptions` built a fresh zero-value `types.IssueViewOptions{}` and overlaid only the field whose `Changed=true`, then unconditionally wrote both `issue_view_open_issues` and `issue_view_ignored_issues` to the config. When the IDE sent only one flag with `Changed=true` (and the other with `Changed=false`), the unchanged flag was silently overwritten with `false`.

## Fix

Seed `ivo` from the current config (`config.GetIssueViewOptions(conf)`), mirroring the existing `extractSeverityFilterFromSettings` pattern, so the per-field gating in `settingBool`/`settingPresent` correctly preserves the flag whose `Changed=false`. Combined with `analytics.SendAnalyticsForFields`'s existing per-field `oldVal != newVal` filter, this also emits exactly one analytics event per actually-toggled flag.

## Commits

1. **`dea22436`** — production fix in `application/server/configuration.go` + 6 unit tests (Gherkin acceptance criteria).
2. **`0a77a8c9`** — QA follow-up (test-only):
   - Renamed `..._NeitherPresentIsNoOp` → `..._EmptySettingsMapIsNoOp` (the empty-map case exits early in `processConfigSettings`, never reaching the inner guard, which is already covered by `..._NeitherChangedIsNoOp`).
   - Renamed `..._EmitsAnalyticsOnlyForActuallyChangedFields` → `..._PreservesAndPropagatesUnchangedFieldWithLspInitialized` and set `SettingIsLspInitialized=true` so the analytics branch is actually exercised. Coverage of `applyIssueViewOptions` rises from 80.0% to **95.0%**.

## Test plan

- [x] 6 new unit tests pass — `go test ./application/server/ -run ApplyIssueViewOptions -v`
- [x] Full `application/server` package: `ok` (105s)
- [x] `make test` (unit + JS): all 127 packages pass on Node 20
- [x] `INTEG_TESTS=1 make test`: all 127 packages pass
- [x] `SMOKE_TESTS=1 make test`: passes locally except `Test_SmokePrecedence_OrgScope_UserFolderOverrideReflectedInNotification` which is **pre-existing** on the parent commit `c9b4a8b3` and requires the `SNYK_TOKEN_CONSISTENT_IGNORES` test-org token (set in CI) for LDX-Sync lock validation.
- [x] `make lint` clean, `go vet` clean.
- [x] `snyk_code_scan` on `application/server`: 0 issues. No dependency changes.

## Out of scope (follow-ups)

- The `propagations` map populated by these helpers is dead code in production today (no reader after `processConfigSettings`); its removal is already in flight on PR #1216.

___

### **PR Type**
Bug fix

___

### **Description**
- Fixes issue view options overwrite bug.

- Preserves unchanged flags from partial payloads.

- Improves configuration update logic.

- Adds comprehensive unit tests for validation.

___

<tr>
  <td>
    <details>
      <summary><strong>configuration.go</strong><dd><code>Preserve issue view options on partial configuration updates</code></dd></summary>
<hr>

application/server/configuration.go

<ul><li>Modified <code>applyIssueViewOptions</code> to initialize <code>types.IssueViewOptions</code> <br>from the current configuration using <code>config.GetIssueViewOptions(conf)</code>.<br> <li> This change prevents partial <code>workspace/didChangeConfiguration</code> payloads <br>from overwriting unchanged issue view flags with their zero values.<br> <li> Ensures that only explicitly changed fields are updated, preserving <br>the state of others.</ul>

</details>

  </td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1228/changes#diff-eb3fe6fd1956591838595b08979653ec23c0b0c16f5340e9e89a05c30e921607">+3/-3</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
<tr>
  <td>
    <details>
      <summary><strong>configuration_test.go</strong><dd><code>Add comprehensive tests for issue view options configuration</code></dd></summary>
<hr>

application/server/configuration_test.go

<ul><li>Introduced new unit tests to specifically verify the correct <br>preservation of issue view options when partial configuration updates <br>are received, covering scenarios where only one flag is changed.<br> <li> Added a helper function <code>seedIssueViewOptions</code> to facilitate setting and <br>verifying configuration states for tests.<br> <li> Added tests for edge cases like empty settings maps and no changed <br>fields.<br> <li> Refactored existing tests for better clarity and coverage, including <br>tests that trigger analytics paths.</ul>

</details>

  </td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1228/changes#diff-e141faafb3f0507956b85e92ac3b584c7c706bd96e7f428c7ef95d0d8b2df564">+129/-0</a>&nbsp; </td>

</tr>

</details>

___

### DIFF
--- a/application/server/configuration.go
+++ b/application/server/configuration.go
@@ -620,7 +620,7 @@ func applyIssueViewOptions(conf configuration.Configuration, engine workflow.Eng
 		return
 	}
 
-	ivo := &types.IssueViewOptions{}
+	ivo := config.GetIssueViewOptions(conf)
 	if v, ok := settingBool(settings, types.SettingIssueViewOpenIssues); ok {
 		ivo.OpenIssues = v
 	}
@@ -629,7 +629,7 @@ func applyIssueViewOptions(conf configuration.Configuration, engine workflow.Eng
 	}
 
 	oldValue := config.GetIssueViewOptions(conf)
-	modified := config.SetIssueViewOptionsOnConfig(conf, ivo, logger)
+	modified := config.SetIssueViewOptionsOnConfig(conf, &ivo, logger)
 	if !modified {
 		return
 	}
@@ -637,7 +637,7 @@ func applyIssueViewOptions(conf configuration.Configuration, engine workflow.Eng
 	propagations[types.SettingIssueViewIgnoredIssues] = ivo.IgnoredIssues
 	sendDiagnosticsForNewSettings(conf, logger)
 	if conf.GetBool(types.SettingIsLspInitialized) {
-		analytics.SendAnalyticsForFields(conf, engine, logger, "issueViewOptions", &oldValue, ivo, triggerSource, map[string]func(*types.IssueViewOptions) any{
+		analytics.SendAnalyticsForFields(conf, engine, logger, "issueViewOptions", &oldValue, &ivo, triggerSource, map[string]func(*types.IssueViewOptions) any{
 			"OpenIssues":    func(s *types.IssueViewOptions) any { return s.OpenIssues },
 			"IgnoredIssues": func(s *types.IssueViewOptions) any { return s.IgnoredIssues },
 		}, configResolver)
--- a/application/server/configuration_test.go
+++ b/application/server/configuration_test.go
@@ -1664,3 +1664,116 @@ func Test_hasFilterChangesInLspConfig(t *testing.T) {
 		})
 	}
 }
+
+// IDE-1946: applyIssueViewOptions must seed from current config so partial payloads
+// (one flag Changed=true, the other Changed=false) preserve the unchanged flag.
+
+// seedIssueViewOptions writes the issue view options directly to conf and asserts
+// they are observable, so the test's precondition is unambiguous regardless of
+// any test-harness-provided defaults.
+func seedIssueViewOptions(t *testing.T, conf configuration.Configuration, opts types.IssueViewOptions) {
+	t.Helper()
+	conf.Set(configresolver.UserGlobalKey(types.SettingIssueViewOpenIssues), opts.OpenIssues)
+	conf.Set(configresolver.UserGlobalKey(types.SettingIssueViewIgnoredIssues), opts.IgnoredIssues)
+	require.Equal(t, opts, config.GetIssueViewOptions(conf), "test seed must be observable via GetIssueViewOptions")
+}
+
+func TestApplyIssueViewOptions_PreservesOpenWhenOnlyIgnoredChanged(t *testing.T) {
+	engine, _ := testutil.UnitTestWithEngine(t)
+	conf := engine.GetConfiguration()
+
+	seedIssueViewOptions(t, conf, types.IssueViewOptions{OpenIssues: true, IgnoredIssues: false})
+
+	UpdateSettings(conf, engine, engine.GetLogger(), map[string]*types.ConfigSetting{
+		types.SettingIssueViewOpenIssues:    {Value: true, Changed: false},
+		types.SettingIssueViewIgnoredIssues: {Value: true, Changed: true},
+	}, nil, analytics.TriggerSourceTest, testutil.DefaultConfigResolver(engine))
+
+	actual := config.GetIssueViewOptions(conf)
+	assert.True(t, actual.OpenIssues, "OpenIssues must be preserved as true when only Ignored is Changed")
+	assert.True(t, actual.IgnoredIssues, "IgnoredIssues must be updated to true")
+}
+
+func TestApplyIssueViewOptions_PreservesIgnoredWhenOnlyOpenChanged(t *testing.T) {
+	engine, _ := testutil.UnitTestWithEngine(t)
+	conf := engine.GetConfiguration()
+
+	seedIssueViewOptions(t, conf, types.IssueViewOptions{OpenIssues: true, IgnoredIssues: true})
+
+	UpdateSettings(conf, engine, engine.GetLogger(), map[string]*types.ConfigSetting{
+		types.SettingIssueViewOpenIssues:    {Value: false, Changed: true},
+		types.SettingIssueViewIgnoredIssues: {Value: true, Changed: false},
+	}, nil, analytics.TriggerSourceTest, testutil.DefaultConfigResolver(engine))
+
+	actual := config.GetIssueViewOptions(conf)
+	assert.False(t, actual.OpenIssues, "OpenIssues must be updated to false")
+	assert.True(t, actual.IgnoredIssues, "IgnoredIssues must be preserved as true when only Open is Changed")
+}
+
+func TestApplyIssueViewOptions_BothChangedWritesBoth(t *testing.T) {
+	engine, _ := testutil.UnitTestWithEngine(t)
+	conf := engine.GetConfiguration()
+
+	seedIssueViewOptions(t, conf, types.IssueViewOptions{OpenIssues: true, IgnoredIssues: false})
+
+	UpdateSettings(conf, engine, engine.GetLogger(), map[string]*types.ConfigSetting{
+		types.SettingIssueViewOpenIssues:    {Value: false, Changed: true},
+		types.SettingIssueViewIgnoredIssues: {Value: true, Changed: true},
+	}, nil, analytics.TriggerSourceTest, testutil.DefaultConfigResolver(engine))
+
+	actual := config.GetIssueViewOptions(conf)
+	assert.False(t, actual.OpenIssues)
+	assert.True(t, actual.IgnoredIssues)
+}
+
+func TestApplyIssueViewOptions_NeitherPresentIsNoOp(t *testing.T) {
+	engine, _ := testutil.UnitTestWithEngine(t)
+	conf := engine.GetConfiguration()
+
+	seed := types.IssueViewOptions{OpenIssues: true, IgnoredIssues: false}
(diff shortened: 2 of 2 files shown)


bastiandoetsch (commented):
## Two-Pass PR Review — IDE-1946 issue view options preservation

**Verdict: Comment-only (looks good to merge — no blocking findings).**

Reviewed against base `refactor/IDE-1786_folder-config-refactoring` (`base-1228`). Diff scope: 6 lines in `application/server/configuration.go` + 129 test lines.

### Critical (must fix before merge)
_None._

### Should Fix
_None._

### Suggestions / Nits
- **[application/server/configuration_test.go:1772]** _Nit — analytics goroutine leak in `TestApplyIssueViewOptions_PreservesAndPropagatesUnchangedFieldWithLspInitialized`._ Setting `SettingIsLspInitialized=true` correctly drives the analytics branch, but `analytics.SendConfigChangedAnalytics` spawns a fire-and-forget `go SendConfigChangedAnalyticsEvent(...)` (`infrastructure/analytics/config_analytics.go:63,69`) that performs a real HTTP call to `api.snyk.io` for each toggled field. The test does not wait for it. This matches the established pattern in `notification_test.go`, `server_test.go`, etc., so it is not introduced here, but a future hardening could either (a) inject a stub `analytics.SendConfigChangedAnalytics` via package-level var, or (b) gate the goroutine on a test-only sync mechanism. **Not blocking** for this PR.

### Verified Correct
- **`&ivo` aliasing safety** — `ivo` (line 623) and `oldValue` (line 631) are independent local value-typed variables. Their addresses do not alias. `SetIssueViewOptionsOnConfig` (`internal/types/config_readers.go:113`) only reads `*opts` and writes the two scalar bool fields back to `conf` via `conf.Set(...)`; it does not retain or mutate `ivo`. `propagations[...] = ivo.{OpenIssues,IgnoredIssues}` (lines 636-637) reads the post-overlay struct correctly.
- **Two `GetIssueViewOptions(conf)` reads (lines 623 and 631) are consistent** — between them, only `settingBool(settings, …)` is called, which is read-only on the `settings` map and does not touch `conf`. Both reads return the same snapshot, so the seed and `oldValue` represent the same pre-write state.
- **Pointer-to-value migration of `analytics.SendAnalyticsForFields`** — `SendAnalyticsForFields[T]` (`infrastructure/analytics/config_analytics.go:112-120`) extracts field values **synchronously** via `getter(oldValue)` / `getter(newValue)` (lines 114-115) into local primitives before any goroutine is spawned. The closures `func(s *types.IssueViewOptions) any { return s.OpenIssues }` therefore see `ivo` at the moment of evaluation, not at goroutine-execution time. Safe.
- **Reference-pattern fidelity vs. `extractSeverityFilterFromSettings`** — the new code mirrors the established pattern at `application/server/configuration.go:580-585` (seed from current config, overlay only present-and-changed fields, set back). `applySeverityFilter` calls `analytics.SendAnalyticsForFields(..., &oldValue, sf, ...)` (line 552) where `sf` is `*SeverityFilter`; here we pass `&ivo` since we changed `ivo` to a value type — equivalent semantics.
- **All other call sites of `config.SetIssueViewOptionsOnConfig` are unaffected.** Verified: `domain/ide/command/toggle_tree_filter.go:111`, `domain/ide/command/toggle_tree_filter_test.go:76,97`, `application/config/config_test.go:239,247,250`, `internal/testutil/test_setup.go:213`, `domain/ide/workspace/folder_test.go:267,536,663` — all use `*IssueViewOptions` (via `util.Ptr(...)` or `&literal`) and continue to compile/work unchanged. The signature itself was not modified.
- **`seedIssueViewOptions` write path agrees with read path.** It writes via `configresolver.UserGlobalKey(...)`. `config.GetIssueViewOptions` → `types.GetIssueViewOptionsFromConfig` → `GetGlobalBool` (`internal/types/config_readers.go:47-54`) reads `conf.Get(configresolver.UserGlobalKey(key))` first, falling back to the bare key. The `require.Equal` postcondition in the helper makes this guarantee an explicit test invariant.
- **`TestApplyIssueViewOptions_EmptySettingsMapIsNoOp` comment matches reality.** `processConfigSettings` (`application/server/configuration.go:242-247`) has `if len(settings) == 0 { return }` _before_ calling `applyIssueViewOptions`. The renamed test exercises that outer guard; the inner `!openPresent && !ignoredPresent` guard is covered by `TestApplyIssueViewOptions_NeitherChangedIsNoOp`.
- **`TestApplyIssueViewOptions_PreservesAndPropagatesUnchangedFieldWithLspInitialized` reaches the analytics branch.** Since `UnitTestWithEngine` returns a fresh engine per test and `conf.GetBool(types.SettingIsLspInitialized)` reads the bare key directly (no `UserGlobalKey` resolver — see `Test_SettingIsLspInitialized_UseBareKey` at line 1479), `conf.Set(types.SettingIsLspInitialized, true)` correctly drives the `if conf.GetBool(...)` branch on line 639. With `oldValue.OpenIssues == newValue.OpenIssues == true`, only the IgnoredIssues analytics is emitted — exactly as the test's outcome assertions imply.
- **Test isolation across the suite.** `testutil.UnitTestWithEngine` creates a fresh `engine := config.InitEngine(...)` per test (`internal/testutil/test_setup.go:101-130`) and registers `t.Cleanup(...)`; configuration state is per-engine, not global. Setting `SettingIsLspInitialized=true` does not bleed into sibling tests.
- **All 6 new tests pass locally.** `go test ./application/server/ -run ApplyIssueViewOptions -v -count=1` → `ok 1.847s`.

### Triage Hypotheses Refuted
- **[H2] Two `GetIssueViewOptions` reads diverging:** REFUTED — no mutation between reads.
- **[H3] Closures over `&ivo` deferred to a goroutine:** REFUTED — `SendAnalyticsForFields` extracts values synchronously before spawning.
- **[H5] `seedIssueViewOptions` write/read path mismatch:** REFUTED — both ends route through `UserGlobalKey`-prefixed reads/writes.
- **[H6] Empty-map test comment misrepresenting which guard runs:** REFUTED — `processConfigSettings` returns early on empty map; comment is accurate.

### Out-of-scope (per request — not flagged)
- `propagations` map dead code (handled in PR #1216).
- `Test_SmokePrecedence_OrgScope_UserFolderOverrideReflectedInNotification` (pre-existing on parent commit; CI-token dependent).

---
**Final verdict: Comment-only. Approve from my side once a human reviewer signs off — the fix is correct, mirrors the established `applySeverityFilter` pattern, and is well-covered by the new tests.**

application/server/configuration.go
```diff
@@ -620,7 +620,7 @@ func applyIssueViewOptions(conf configuration.Configuration, engine workflow.Eng
 		return
 	}
 
-	ivo := &types.IssueViewOptions{}
+	ivo := config.GetIssueViewOptions(conf)
```
bastiandoetsch, line 623: This is the relevant change that fixes the bug. Everything else is necessary signature changes and tests verifying the change.

application/server/configuration_test.go
```diff
@@ -1664,3 +1664,132 @@ func Test_hasFilterChangesInLspConfig(t *testing.T) {
...
+}
+
+func TestApplyIssueViewOptions_PreservesOpenWhenOnlyIgnoredChanged(t *testing.T) {
+	engine, _ := testutil.UnitTestWithEngine(t)
+	conf := engine.GetConfiguration()
+
+	seedIssueViewOptions(t, conf, types.IssueViewOptions{OpenIssues: true, IgnoredIssues: false})
+
+	UpdateSettings(conf, engine, engine.GetLogger(), map[string]*types.ConfigSetting{
+		types.SettingIssueViewOpenIssues:    {Value: true, Changed: false},
```
rrama: Should you not be trying to set it to false, so the assertion below makes more sense?

application/server/configuration_test.go
```diff
@@ -1664,3 +1664,132 @@ func Test_hasFilterChangesInLspConfig(t *testing.T) {
...
+
+func TestApplyIssueViewOptions_PreservesIgnoredWhenOnlyOpenChanged(t *testing.T) {
+	engine, _ := testutil.UnitTestWithEngine(t)
+	conf := engine.GetConfiguration()
+
+	seedIssueViewOptions(t, conf, types.IssueViewOptions{OpenIssues: true, IgnoredIssues: true})
+
+	UpdateSettings(conf, engine, engine.GetLogger(), map[string]*types.ConfigSetting{
+		types.SettingIssueViewOpenIssues:    {Value: false, Changed: true},
+		types.SettingIssueViewIgnoredIssues: {Value: true, Changed: false},
```
rrama: Same as above. Should you not be trying to set it to false, so the assertion below makes more sense?

rrama (approved):
LGTM, just some slight test changes to ensure the value expected is because of the ignoring instead of because it isn't ignoring and the set value happens to match.

application/server/configuration_test.go
bastiandoetsch: Hm. to me the tests appear correct. But maybe we can have a quick chat.

PR #1240, opened 2026-04-30
fix: refresh LDX-Sync before populating SAST cache on login [IDE-1981]

nick-y-snyk:
## Summary
- When LDX-Sync resolves a different org after login than before, the post-credential hook cached feature flags + SAST settings keyed by the pre-login org. Subsequent scan reads the wrong cache entry and reports "Snyk Code is not enabled for this organization".
- Fix: in `loginCommand` post-credential hook, run `RefreshConfigFromLdxSync` before `populateFolderFeatureFlagsAndSastSettings`.
- Rename `populateAllFolderConfigs` → `populateFolderFeatureFlagsAndSastSettings` for clarity.

## Test fix: `Test_loginCommand_StartsAuthentication`
Moving the login-time `RefreshConfigFromLdxSync` call into the post-credential hook surfaced a latent test setup gap:

- During `initialized`, scanner `Init` runs an auth check that stores the fake provider's fixed token in config.
- `snyk.login`'s `Authenticate` then returns the same token, so `updateCredentials` early-returns on the no-op (`oldToken == newToken`).
- The post-credential hook — where the login-time refresh now lives — never fires, and the test's second `RefreshConfigFromLdxSync` expectation goes unmet.

Production behavior is correct: a no-op re-auth shouldn't trigger a fresh `$/snyk.hasAuthenticated`, so refreshing settings on it would be wasted work. The test was implicitly relying on the pre-PR direct call from `loginCommand.Execute`, which ran regardless of token change.

Fix: clear the token after `initialized` in the test, simulating the realistic "unauthenticated user clicks Login" precondition. No production change.

## Test plan
- [x] `go test ./application/server -run Test_loginCommand_StartsAuthentication`
- [x] `go test ./domain/ide/command/ -run TestLoginCommand`
- [x] `make lint`
- [x] Manual repro in VS Code Insiders: pre-fix scan errors with "Snyk Code is not enabled"; post-fix scan succeeds.

### DIFF
--- a/domain/ide/command/login.go
+++ b/domain/ide/command/login.go
@@ -104,6 +104,8 @@ func (cmd *loginCommand) Execute(ctx context.Context) (any, error) {
 
 	// LDX-Sync must run before populate: it can change the resolved org, and populate
 	// caches feature flags and SAST settings keyed by org.
+	// Use context.Background() so this is not canceled if the LSP request context is
+	// canceled (e.g. when the IDE cancels the snyk.login request after auth completes).
 	cmd.authService.SetPostCredentialUpdateHook(func() {
 		cmd.ldxSyncService.RefreshConfigFromLdxSync(context.Background(), conf, cmd.engine, logger, config.GetWorkspace(conf).Folders(), cmd.notifier)
 		populateFolderFeatureFlagsAndSastSettings(conf, cmd.engine, logger, cmd.featureFlagService, cmd.configResolver)
--- a/domain/ide/command/login.go
+++ b/domain/ide/command/login.go
@@ -102,11 +102,11 @@ func (cmd *loginCommand) Execute(ctx context.Context) (any, error) {
 		}
 	}
 
-	// Populate feature flags for all workspace folders after credentials are stored
-	// but before the $/snyk.hasAuthenticated notification reaches the IDE.
-	// This prevents a race where the IDE triggers a scan before SAST settings are cached.
+	// LDX-Sync must run before populate: it can change the resolved org, and populate
+	// caches feature flags and SAST settings keyed by org.
 	cmd.authService.SetPostCredentialUpdateHook(func() {
-		populateAllFolderConfigs(conf, cmd.engine, logger, cmd.featureFlagService, cmd.configResolver)
+		cmd.ldxSyncService.RefreshConfigFromLdxSync(context.Background(), conf, cmd.engine, logger, config.GetWorkspace(conf).Folders(), cmd.notifier)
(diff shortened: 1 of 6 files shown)


domain/ide/command/login.go
```diff
@@ -102,11 +102,11 @@ func (cmd *loginCommand) Execute(ctx context.Context) (any, error) {
...
 	}
 
-	// Populate feature flags for all workspace folders after credentials are stored
-	// but before the $/snyk.hasAuthenticated notification reaches the IDE.
-	// This prevents a race where the IDE triggers a scan before SAST settings are cached.
+	// LDX-Sync must run before populate: it can change the resolved org, and populate
+	// caches feature flags and SAST settings keyed by org.
 	cmd.authService.SetPostCredentialUpdateHook(func() {
-		populateAllFolderConfigs(conf, cmd.engine, logger, cmd.featureFlagService, cmd.configResolver)
+		cmd.ldxSyncService.RefreshConfigFromLdxSync(context.Background(), conf, cmd.engine, logger, config.GetWorkspace(conf).Folders(), cmd.notifier)
```
rrama, line 110: Please do not lose the comment about why `context.Backgroud()`.

rrama (approved):

domain/ide/command/login.go
```diff
@@ -102,11 +102,13 @@ func (cmd *loginCommand) Execute(ctx context.Context) (any, error) {
...
-	// Populate feature flags for all workspace folders after credentials are stored
-	// but before the $/snyk.hasAuthenticated notification reaches the IDE.
-	// This prevents a race where the IDE triggers a scan before SAST settings are cached.
+	// LDX-Sync must run before populate: it can change the resolved org, and populate
+	// caches feature flags and SAST settings keyed by org.
+	// Use context.Background() so this is not canceled if the LSP request context is
+	// canceled (e.g. when the IDE cancels the snyk.login request after auth completes).
 	cmd.authService.SetPostCredentialUpdateHook(func() {
-		populateAllFolderConfigs(conf, cmd.engine, logger, cmd.featureFlagService, cmd.configResolver)
+		cmd.ldxSyncService.RefreshConfigFromLdxSync(context.Background(), conf, cmd.engine, logger, config.GetWorkspace(conf).Folders(), cmd.notifier)
```
ShawkyZ, line 110: am just worried that the experience will be slower than before after adding this.
As we have a mutex on auth and it will wait on the post update hook to finish.

nick-y-snyk: follow-up fix would be to re-visit auth mutex. But in any case we can't start using ls until ldx sync is fetched, right? because it can have everything locked for example

PR #1242, opened 2026-04-30
fix: cache if user is authenticated and check for global org only if …

ShawkyZ:
…authenticated

### Description

### Checklist


infrastructure/authentication/auth_service_impl.go
```diff
@@ -127,6 +127,7 @@ func (a *AuthenticationServiceImpl) authenticate(ctx context.Context) (token str
 		err = errors.New("authentication provider is not configured")
 		a.engine.GetLogger().Warn().Err(err).Msg("Failed to authenticate: auth provider is nil")
 		a.authCache.RemoveAll()
+		a.setAuthState(false)
```
rrama, line 130: If we really want this as a bodge last minute temporary solution, we should never go back to `false` to reduce the blast radius. They can reload the IDE if they need to.

PR #1244, opened 2026-04-30
fix: update tooltip wording, fix inconsistent use of full stops [IDE-1966]

andrewrobinsonhodges-snyk:
### Description

### Checklist


infrastructure/configuration/template/config.html
```diff
@@ -205,29 +205,29 @@ <h3>Scanners</h3>
...
                                             <span class="config-scope-slot" data-config-scope-slot="true" data-setting-key="snyk_oss_enabled"></span>
                                         </label>
                                     </div>
                                     <div class="override-indicator-wrapper {{getSourceClass .DefaultsScopes "snyk_code_enabled"}}">
                                         <label class="checkbox-label">
                                             <input type="checkbox" name="snyk_code_enabled" {{if eq .Settings.ActivateSnykCode "true"}}checked{{end}}>
-                                            <span data-toggle="tooltip" title="<p>Find and fix security issues in your application code in real time.</p><p><strong>Note:</strong> Snyk Code must be enabled for your organization in Snyk settings.</p>">Snyk Code</span>
+                                            <span data-toggle="tooltip" title="<p>Find and fix security issues in your application code.</p><p><strong>Note: </strong>Snyk Code must be enabled for your organization.</p>">Snyk Code</span>
```
bastiandoetsch, line 215: `in Snyk Settings` is not better IMO, arguably, they ARE in Snyk settings, so this confuses. I'd be concrete: If the user is org-admin (we should have that information in LDX-Sync), we could add `Enable it <a href="https://{{.Settings.UIEndpoint}}/org/{{.Settings.OrgSlug}}/manage/snyk-code">here</a>`.

Of course the variables would need to be added and the org role would need to be retrieved if we want that. Else, I'd just keep the old version.

bastiandoetsch (approved):

infrastructure/configuration/template/config.html
andrewrobinsonhodges-snyk: That's why I changed it 😉

PR #1245, opened 2026-04-30
perf: bodge fix to skip /rest/self on hot path via IsSet gate [IDE-1987]

nick-y-snyk:
## Summary

Alternative to #1242. Same symptom — settings page load was slow because every read of `ConfigResolver.GlobalOrg()` synchronously hit `/rest/self` — but the real cause is more specific than #1242 framed it.

### Root cause

`ORGANIZATION` is resolved through GAF's `defaultFuncOrganization`, which calls `/rest/self` whenever the key is unset. When the user is **not logged in** (or the token is invalid), the call returns 401 and the default func returns `("", err)`. GAF's default-value cache **does not cache errors** — see `extendedViper.GetWithError` in `pkg/configuration/configuration.go`:

```go
if cacheEnabled && err == nil {
    localDefaultCache.Set(key, value, cacheDuration)
}
```

So every hot-path read on the unauthenticated path re-fired `/rest/self`. Settings page = many reads = many failing 401 round-trips = visible slowness.

### Fix

1. Gate `GlobalOrg()` on `IsSet(configuration.ORGANIZATION)`. `IsSet` bypasses the default func entirely, so the unauthenticated retry loop is impossible from the hot path.
2. Explicit prime calls via `GetGlobalOrganization` populate viper so the `IsSet`-gated read finds the cached UUID once we *are* authenticated:
   - `infrastructure/authentication/auth_service_impl.go::updateCredentials` — covers OAuth, PAT, token-paste in-session.
   - `application/server/server.go::initializeHandler` — covers cold start with a persisted token. Placed **before** `RefreshConfigFromLdxSync` because LDX-Sync's per-folder logic falls back to `GlobalOrg()` when `OrgSetByUser=true && PreferredOrg==""` (`domain/ide/command/ldx_sync_service.go:217`); priming after LDX-Sync would have skipped those folders.

Drops the `SettingIsAuthenticated` mirror approach (#1242) and its 7 `setAuthState` write sites.

### Why IsSet over the auth-state cache (#1242)

- `IsSet` is the precise question the hot path needs to answer: *is `ORGANIZATION` cached locally right now?* The auth-state mirror in #1242 was an indirect proxy.
- No bookkeeping across 7 auth-service sites → no risk of the `authenticate()` → `updateCredentials()` ordering bug that left `state=false` after a successful login (described in the #1242 review).
- #1242 still calls `Get(ORGANIZATION)` after the gate, so it does not actually avoid `/rest/self` when authenticated-but-uncached. This PR avoids it in all cases.

### Trade-off

After logout, `configuration.ORGANIZATION` stays in viper until the next login overwrites it. `GlobalOrg()` would then still return the stale UUID for any UI snapshot rendered between logout and re-login. Considered harmless — it's a label, not a security boundary, and clearing it would require an `Unset` that also writes to persisted on-disk storage.

## Follow-up

[IDE-1988](https://snyksec.atlassian.net/browse/IDE-1988) — Track auth/network state to gate **all** unnecessary outbound calls (not only `/rest/self`). Same "errors not cached → retry storm" failure mode is latent in other GAF default funcs (`ORGANIZATION_SLUG`, `API_URL` via token audience), LDX-Sync, learn cache, feature flags, analytics, scanner init, and `IsAuthenticated()` revalidation while offline. Once that lands, the `IsSet` gate and both prime calls retire.

## Test plan

- [x] `go test ./internal/types/ ./infrastructure/authentication/ ./domain/ide/command/` — green
- [x] `make lint` — 0 issues
- [x] Originally-failing CI test `Test_RefreshConfigFromLdxSync_OrgSetByUserButPreferredOrgEmpty_FallsBackToGlobalOrg` passes without modification
- [ ] Local IDE smoke: settings page loads fast on cold start, no `/rest/self` in trace
- [ ] Cold start with persisted token, autoScan disabled: LDX-Sync applies remote config to folders that rely on global-org fallback (verifies the prime ordering fix)

### DIFF
--- a/application/server/server.go
+++ b/application/server/server.go
@@ -259,6 +259,10 @@ func initializeHandler(conf configuration.Configuration, engine workflow.Engine,
 		config.SetupStorage(conf, storage, &logger)
 
 		addWorkspaceFolders(conf, &logger, engine, params)
+		// Prime ORGANIZATION for hot-path GlobalOrg(); see GetGlobalOrganization.
+		// Must run before RefreshConfigFromLdxSync and HandleFolders, which rely
+		// on the resolver's global-org fallback for folders without a preferred org.
+		_ = types.GetGlobalOrganization(conf)
 		di.LdxSyncService().RefreshConfigFromLdxSync(ctx, conf, engine, &logger, config.GetWorkspace(conf).Folders(), nil)
 		InitializeSettings(conf, engine, &logger, params.InitializationOptions)
 
@@ -460,9 +464,6 @@ func initializedHandler(conf configuration.Configuration, engine workflow.Engine
 			return nil, nil
 		}
 		command.HandleFolders(conf, engine, &logger, context.Background(), srv, di.Notifier(), di.ScanPersister(), di.ScanStateAggregator(), di.FeatureFlagService(), di.ConfigResolver())
-
-		// Prime ORGANIZATION for hot-path GlobalOrg(); see GetGlobalOrganization.
-		_ = types.GetGlobalOrganization(conf)
 
 		deleteExpiredCache(conf)
 		cacheCtx, cancel := context.WithCancel(context.Background())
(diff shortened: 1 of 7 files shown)


rrama (approved):
Good enough for a last minute solution to fix the problem for CB. Well done keeping it minimal.

PR #1254, opened 2026-05-06
feat(scanners): centralize skip errors, scannercommon resolve/logging [IDE-2002]

acke:
### Summary

- Shared utils errors (folder config, OSS/IaC unsupported paths, product flags).
- Explicit errors from Code, IaC, Secrets, and OSS `Scan` for disabled product, missing auth, unsupported paths; `MsgNotAuthenticatedNoScan` for auth.
- OSS returns error when Open Source is disabled for the folder.
- OSS integration smoke test skips when `SNYK_TOKEN` is unset; scanner tests updated for secrets no-token path.
- **`internal/scannercommon`**: `ResolveFolderAndScanType` returns folder config, delta/working-directory scan type label, and workspace folder path; `WithScanContext` attaches standard zerolog fields (`method`, `pathToScan`, `workspaceFolder`, `scanType`).
- Code and OSS scanners hold `logger *zerolog.Logger` (from `engine.GetLogger()` at construction) so scan entry uses the same base logger pattern as IaC/Secrets.

### Description

See **[IDE-2002](https://snyksec.atlassian.net/browse/IDE-2002)**. Latest push adds scannercommon + unified logging wiring.

### Checklist

[IDE-2002]: https://snyksec.atlassian.net/browse/IDE-2002?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


infrastructure/utils/error_messages.go
```diff
@@ -42,4 +86,32 @@ var ErrorConfig = map[string]ErrorMetadata{
...
+		ShowNotification: false,
+		TreeRootSuffix:   "(not authenticated)",
+	},
+}
+
+var nonFailingScanErrors = map[string]bool{
+	ErrOssScanPathUnsupported:         true,
+	ErrIacScanPathUnsupported:         true,
+	MsgNotAuthenticatedNoScan:         true,
+	ErrSnykCodeNotEnabledForFolder:    true,
```
nick-y-snyk, line 95: https://github.com/snyk/snyk-ls/pull/1254/changes#diff-92ee5447150d64c4b9a6c8c8e8834577ada9e739d1a0a2842ef5bc79fd014f26R385
I think these error won't have tree view suffix (important for product enablement for example)

internal/scannercommon/scan.go
```diff
@@ -0,0 +1,70 @@
...
+	scanType := "WorkingDirectory"
+	if deltaScanType, ok := ctx2.DeltaScanTypeFromContext(ctx); ok {
+		scanType = deltaScanType.String()
+	}
+	return workspaceFolderConfig, scanType, workspaceFolderConfig.FolderPath, nil
+}
+
+// WithScanContext returns a child logger with standard ProductScanner fields:
+// method, pathToScan, workspaceFolder, scanType.
+func WithScanContext(
```
nick-y-snyk: naming is not clear (it's a logger helper)

infrastructure/utils/error_messages.go
nick-y-snyk: can you please check if new tree view is using those suffixes

nick-y-snyk (changes requested):

infrastructure/utils/error_messages.go
acke: The new server-rendered tree does not map those catalog errors to the same short suffixes; it shows a generic - (scan failed) in the description and keeps the raw message elsewhere.

nick-y-snyk: so the new tree ignores these data completly? or it's still visible somewhere? How easy to wire these suffixes?

acke: Easy enough, will do that change in a separate PR.

domain/ide/workspace/folder.go
```diff
@@ -382,6 +383,16 @@ func (f *Folder) ProcessResults(ctx context.Context, scanData types.ScanData) {
 
 func (f *Folder) sendScanError(product product.Product, err error) {
 	f.scanNotifier.SendError(product, f.path, err.Error())
+
```
nick-y-snyk, line 386: `SendError` on line 385 fires unconditionally before `IsNonFailingScanError` is checked. `SendError` always marking `ErrorStatus` is fine — it's what produces the `TreeRootSuffix`. The issue is what happens _after_: `f.logger.Err(err)` logs at ERROR level (→ Sentry) and `f.notifier.SendErrorDiagnostic` shows a user popup — for any error that is missing from `nonFailingScanErrors`. A single missing map entry causes Sentry noise + user popup every scan cycle for what should be a quiet skip. There is no compile-time or test-time signal when an entry is omitted. `ErrSastSettingsNotAvailable` is a concrete example: it is in `ErrorConfig` with `ShowNotification: false` but absent from `nonFailingScanErrors`, so it currently falls through to both destructive side-effects.

infrastructure/iac/iac.go
```diff
@@ -116,47 +117,42 @@ func (iac *Scanner) SupportedCommands() []types.CommandName {
...
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
 	documentURI := uri.PathToUri(pathToScan) // todo get rid of lsp dep
 	if !iac.isSupported(documentURI) {
-		logger.Debug().Msg("IAC scanner: skipping unsupported file/directory")
-		return issues, nil
+		// Unsupported paths are normal for background scans; skip without error (see OSS CLIScanner).
+		logger.Debug().Msg(utils.ErrIacScanPathUnsupported)
+		return []types.Issue{}, nil
```
nick-y-snyk, line 155: `utils.ErrIacScanPathUnsupported` is used as a **log message string** here; the function returns `nil` error. This makes the `nonFailingScanErrors` and `ErrorConfig` entries for `ErrIacScanPathUnsupported` (and `ErrOssScanPathUnsupported` in `cli_scanner.go`) unreachable dead code — `sendScanError` is never called for this path, so the `(unsupported path)` tree suffix never appears in the IDE. `Test_SendError_UnsupportedPath_SendsErrorStatus` also tests behaviour no production path triggers.

Related: https://github.com/snyk/snyk-ls/pull/1270

Pick one design and apply consistently to both scanners:
- **Option A (keep silent skip — matches the comment intent):** Remove `ErrOssScanPathUnsupported` / `ErrIacScanPathUnsupported` from `ErrorConfig` and `nonFailingScanErrors`, delete `Test_SendError_UnsupportedPath_SendsErrorStatus`, use a plain string for the log message.
- **Option B (surface in IDE tree):** Return `nil, errors.New(utils.ErrIacScanPathUnsupported)` — consistent with the disabled-product and no-auth paths; the `nonFailingScanErrors` entry will then suppress the diagnostic while showing the suffix.

bastiandoetsch: The correct solution is Option A, like in #1270.

domain/ide/workspace/folder.go
acke: Compared to main, IDE-2002 reduces Sentry + popup noise for anything listed in nonFailingScanErrors. Anything not in that map still behaves like main for logging + diagnostics (including ErrSastSettingsNotAvailable if it’s only in ErrorConfig).

If we want another behavior, we would have needed to separate errors from skipped scans, a I suggestion that was discussed a few days ago.

infrastructure/iac/iac.go
acke: I implemented Option A.

nick-y-snyk (approved):

infrastructure/utils/error_messages.go
```diff
@@ -42,4 +82,24 @@ var ErrorConfig = map[string]ErrorMetadata{
...
+// or shown as error diagnostics. Do not add log-only strings here (scanners must return these as err).
+var nonFailingScanErrors = map[string]bool{
+	MsgNotAuthenticatedNoScan:         true,
+	ErrSnykCodeNotEnabledForFolder:    true,
+	ErrSnykIacNotEnabledForFolder:     true,
+	ErrSnykOssNotEnabledForFolder:     true,
+	ErrSnykSecretsNotEnabledForFolder: true,
+	ErrSnykCodeNotEnabled:             true,
+	ErrSnykSecretsNotEnabled:          true,
+}
```
bastiandoetsch, line 101: **[Critical]** `ErrSastSettingsNotAvailable` is registered in `ErrorConfig` with `ShowNotification: false` and `TreeRootSuffix: "(Code settings unavailable)"` (line 73), but is absent from `nonFailingScanErrors`. When `code.go` returns this error, `sendScanError` correctly emits the tree suffix via `scanNotifier.SendError`, but `IsNonFailingScanError` returns `false`, so `notifier.SendErrorDiagnostic` is also called. Users see both the tree node suffix AND an error diagnostic pop-up, directly contradicting the PR's intent.

Two fix options:
1. Add `ErrSastSettingsNotAvailable: true` to this map.
2. (Better) Derive `IsNonFailingScanError` directly from `ErrorConfig` (`!ErrorConfig[msg].ShowNotification`), making this class of divergence structurally impossible.

application/server/server_smoke_test.go
```diff
@@ -902,23 +902,25 @@ func checkForScanParamsWithMaxWait(t *testing.T, jsonRPCRecorder *testsupport.Js
...
+				types.PathKey(scanParams.FolderPath) != expectedFolderKey ||
 				scanParams.Status == types.InProgress {
 				continue
 			}
 			finalScanParams = &scanParams
 			return true
 		}
 		return false
-	}, maxWait, time.Millisecond,
+	}, maxIntegTestDuration, time.Millisecond,
```
bastiandoetsch: **[Critical]** The `maxWait` parameter is accepted by this function but silently ignored — `maxIntegTestDuration` (15 min) is hardcoded in the `require.Eventually` call instead. Two callers pass `monorepoRealScanPhaseMaxWait = 90 * time.Minute`, explicitly documented as needed for large fixture scans. Those callers now wait 15 minutes, not 90. Slow CI runners will hit spurious timeouts with misleading assertion errors.

Fix: replace `maxIntegTestDuration` with `maxWait` here.

application/server/server_smoke_test.go
```diff
@@ -1431,44 +1433,176 @@ func Test_SmokeScanUnmanaged(t *testing.T) {
...
+func folderConfigsHaveNonEmptyAutoDeterminedOrgForValidators(
+	param types.LspConfigurationParam,
+	validators map[types.FilePath]func(types.LspFolderConfig),
+) bool {
+	if len(validators) == 1 && len(param.FolderConfigs) == 1 {
+		var want types.FilePath
+		for w := range validators {
+			want = w
+			break
+		}
```
bastiandoetsch, line 1495: **[Should Fix]** Dead code: both branches of the `if folderConfigPathsMatch` check return the identical expression, so the path comparison result is discarded. Any single-folder `$/snyk.configuration` notification satisfies any single validator regardless of folder path, weakening the LDX-sync assertions that depend on `lspFolderConfigWaitForAutoDeterminedOrg`.

Fix: remove the `if/else` and return `configSettingHasNonEmptyStringValue(fc.Settings[types.SettingAutoDeterminedOrg])` directly — or add the path-match to the return condition.

application/server/server_smoke_test.go
```diff
@@ -1431,44 +1433,176 @@ func Test_SmokeScanUnmanaged(t *testing.T) {
...
+func countValidatedFolderConfigs(
+	lastConfigParam types.LspConfigurationParam,
+	validators map[types.FilePath]func(types.LspFolderConfig),
+) int {
 	validationsCount := 0
-	for _, folderConfig := range lastConfigParam.FolderConfigs {
-		validator, ok := validators[folderConfig.FolderPath]
-		if ok {
+	if len(lastConfigParam.FolderConfigs) == 1 && len(validators) == 1 {
+		var onlyValidator func(types.LspFolderConfig)
```
bastiandoetsch: **[Should Fix]** The single-folder/single-validator fast path applies the validator to `folderConfigs[0]` without checking whether the folder path matches the validator key. Any single-folder notification passes as long as one entry arrives, regardless of which folder it describes. This weakens all single-folder `requireLspFolderConfigNotification` assertions.

Fix: apply the same `folderConfigPathsMatch` guard used by the multi-folder path below.

internal/scannercommon/scan.go
```diff
@@ -0,0 +1,70 @@
+/*
```
bastiandoetsch, line 1: **[Should Fix]** New package with 0% test coverage. The contract it enforces — that every error produced by `RequireProductEnabled`/`RequireAuthToken` is classified as non-failing by `utils.IsNonFailingScanError` — has no unit tests. A developer adding a new product check or error constant will compile successfully while silently regressing the "no diagnostic pop-up" UX behavior.

Suggested test: `Test_RequireProductEnabled_ErrorIsNonFailing` asserting `utils.IsNonFailingScanError(err.Error())` for every error string the helpers can produce.

bastiandoetsch (commented):
Adversarial verification findings — **CONTESTED**. Two critical bugs where the implementation does not achieve its stated intent, plus three should-fix items.

infrastructure/utils/error_messages.go
acke: Users see both the tree node suffix AND an error diagnostic pop-up, this is the intended behavior. Do we want to change this?

bastiandoetsch: Hm. Secrets handles it as "no error" when the feature flag is disabled. We can leave it be for now and should just have a general solution.

acke: To make this a limited PR, I didnt create a clear separation between actual errors and cases where we skip a scan for other reasons. I think that would have been the cleaner solution. But to keep the it limited, I kept it to just adding errors, and handle them correctly. In the end it is a bit tricky to handle the different behaviors with this approach, so I really think we need to revisit this.

application/server/server_smoke_test.go
acke: What we did: Keep the single-folder fast path, but return false when paths do not match, then require non-empty SettingAutoDeterminedOrg.

internal/scannercommon/scan.go
acke: In internal/scannercommon/scan_test.go, we added three tests:

Test_RequireProductEnabled_ErrorIsNonFailing — For each errNotEnabledForFolder string the real scanners pass into RequireProductEnabled (the four utils.Err*NotEnabledForFolder constants), calls RequireProductEnabled(false, msg) and asserts utils.IsNonFailingScanError(err.Error()).

Test_RequireAuthToken_NoTokenErrorIsNonFailing — Empty token via configuration.NewWithOpts(), then asserts the returned error is non-failing (that message is always utils.MsgNotAuthenticatedNoScan).

Test_RequireProductEnabled_EnabledReturnsNil — Sanity check that the happy path returns no error.

PR #1277, opened 2026-05-13
fix(autofix): increase `GetAutofixDiffs` timeout to 4 mins

pasiunaite:
### Description

The Autofix service is transitioning to an agentic LLM approach, which introduces increased latency during fix generation. We observed that some test requests were timing out on the client side even though the fixes were generated successfully on the Autofix backend. 

While we are actively working on improving the generation latency, we would like to increase the autofix polling timeout in `GetAutofixDiffs` from **2 minutes to 4 minutes**. This gives the agentic Autofix requests sufficient time to complete and return the generated diffs.

### Checklist

### DIFF
--- a/infrastructure/code/autofix.go
+++ b/infrastructure/code/autofix.go
@@ -70,8 +70,8 @@ func (sc *Scanner) GetAutofixDiffs(ctx context.Context, baseDir types.FilePath, 
 	// ticker sends a trigger every second to its channel
 	ticker := time.NewTicker(1 * time.Second)
 	defer ticker.Stop()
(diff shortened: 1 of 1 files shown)


rrama (approved):
LGTM. 
An aside for Team IDE, I was checking if any of the test timeouts need raising as a result and was going to suggest
https://github.com/snyk/snyk-ls/blob/1cc2994ab8d9aef45170fe23b0fe9bc0f4bdead7/application/server/server_smoke_test.go#L974-L1003
but I realised it doesn't actually wait for the diffs like it implies it does. It just waits for the initial HTML response to be sent back...

PR #1282, opened 2026-05-15
fix(snyk-ls): reset summary panel on org change [IDE-1969]

acke:
### Description

Fixes [IDE-1969](https://snyksec.atlassian.net/browse/IDE-1969): the Summary Panel should reset when org context changes so it does not keep stale scan results and org labels.

**Behavior**

- **Global org** (`organization` in workspace settings): after LSP initialization, when the resolved global org id changes, call `resetSummaryPanelForOrgChange` → `ScanStateAggregator.Init` → `$/snyk.scanSummary` (empty / not-started state).
- **Folder org** (`updateFolderOrgIfNeeded`): when folder org snapshot fields change per `folderConfigsOrgSettingsEqual` — `preferred_org`, `org_set_by_user`, and/or `auto_determined_org` — reset the summary the same way (covers manual org, auto-select toggle, and LDX-only org metadata changes).
- **PR review follow-up** (`8eb96730`): scope summary reset to affected folders only; batch a single `Init` per folder-config pass; inject scan aggregator via `TestInitWithScanStateAggregator` instead of a global setter; restore LDX smoke org assertions with explicit skip when CI has no mapping.

**Also includes** [IDE-2042](https://snyksec.atlassian.net/browse/IDE-2042): pre-push `check-tests` requires only `make test`; integration and smoke are advisory (warn, do not block).

### Checklist

- [x] Tests added and all succeed (`make test` locally)
- [x] Regenerated mocks, etc. (`make generate`) — pre-commit verify-generate clean
- [x] Linted (`make lint-fix` / `make format`)

### Test plan

- [x] `go test ./application/server/... -run 'ResetsSummaryPanelOnOrgChange|PreferredOrgChange'`
- [x] `make test` (unit)
- [ ] CI: integration + smoke
- [ ] Manual: project org / auto-select / global defaults — summary org label matches effective scan org after apply
- [ ] Manual: severity filter still refreshes summary (regression)

[IDE-1969]: https://snyksec.atlassian.net/browse/IDE-1969
[IDE-2042]: https://snyksec.atlassian.net/browse/IDE-2042

___

### **PR Type**
Bug fix, Enhancement

___

### **Description**
- Reset scan state aggregator when organization context changes.

- Handles global and per-folder organization changes.

- Ensures reset only after LSP initialization.

- Adds comprehensive test cases for the new logic.

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A[IDE sends org change] --> B{LSP Initialized?};
  B -- Yes --> C{Global Org Changed?};
  B -- No --> Z[No Reset];
  C -- Yes --> D[Apply Global Org Change];
  C -- No --> E{Folder Org Changed?};
  D --> F[Signal Global Org Change];
  F --> G{Process Folder Configs};
  E -- Yes --> G;
  E -- No --> G;
  G -- Org Change Detected --> H[Collect Changed Folder Paths];
  G -- Global Org Changed --> H;
  H --> I{LSP Initialized AND Paths Found?};
  I -- Yes --> J[Call resetSummaryPanelForOrgChange];
  J --> K[ScanAggregator.Init called for affected folders];
  I -- No --> Z;
  Z -.-> L[Summary Panel Stale];
  K -.-> M[Summary Panel Reset];

  subgraph Configuration Handling
    A
    B
    C
    D
    E
    F
  end

  subgraph Folder Configuration Processing
    G
    H
    I
  end

  subgraph Summary Panel Reset Logic
    J
    K
    L
    M
  end
```

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Tests</strong></td><td><details><summary>7 files</summary><table>
<tr>
  <td><strong>test_init.go</strong><dd><code>Add test utility for scan state aggregator injection</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1282/changes#diff-fc304a8ceac9f55997a1ed4f30020de933d3ada1a11eacd32c236acdce0cec81">+11/-1</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>configuration_test.go</strong><dd><code>Add tests for summary panel reset on org change</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1282/changes#diff-e141faafb3f0507956b85e92ac3b584c7c706bd96e7f428c7ef95d0d8b2df564">+261/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>ldx_sync_smoke_test.go</strong><dd><code>Refactor LDX sync smoke tests with helper functions</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1282/changes#diff-e86bba96f063a23a5407b87f1d43156fbf47ff435cf44fc7324550746bc13305">+26/-34</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>checker_test.go</strong><dd><code>Add filesystem probe for read-only tests</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1282/changes#diff-678179d9dc9c3bae2216c65f6e869d924ae6dec02b68b552729b21372daf3c47">+8/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>flock_unix_test.go</strong><dd><code>Helper to detect flock enforcement on Unix-like systems</code>&nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1282/changes#diff-b86b7397fd101daa3cf43b8c81b7a8932846af2ab0cdb13b6b9f07f788fcfede">+52/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>flock_windows_test.go</strong><dd><code>Define flock enforcement for Windows (always false)</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1282/changes#diff-a0a06aee5dbfcd7f52ac2d4066689694431a1d20caa9e36c8eaf2743b6c1ea9d">+25/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>storage_test.go</strong><dd><code>Skip parallel file locking tests on non-enforcing filesystems</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1282/changes#diff-ea16c7f9e9f3a1352ce3a259461e3886479146970ba1f8baaa26ceafa1a83117">+3/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
<tr>
  <td><strong>configuration.go</strong><dd><code>Implement summary panel reset logic on org change</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1282/changes#diff-eb3fe6fd1956591838595b08979653ec23c0b0c16f5340e9e89a05c30e921607">+103/-23</a></td>

</tr>
<tr>
  <td><strong>check-tests-run.sh</strong><dd><code>Clarify advisory stages in pre-push hook script</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1282/changes#diff-73ba738421dae7ee6ae3e0fd9175ffaba58b7be97904d5ce0c0dc5e36f1120dd">+5/-2</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

</details>

___

### DIFF
--- a/application/di/init.go
+++ b/application/di/init.go
@@ -247,6 +247,12 @@ func ScanStateAggregator() scanstates.Aggregator {
 	return scanStateAggregator
 }
 
+func SetScanStateAggregator(agg scanstates.Aggregator) {
+	initMutex.Lock()
+	defer initMutex.Unlock()
+	scanStateAggregator = agg
+}
+
 func ScanNotifier() scanner2.ScanNotifier {
 	initMutex.Lock()
 	defer initMutex.Unlock()
--- a/application/di/init.go
+++ b/application/di/init.go
@@ -247,12 +247,6 @@ func ScanStateAggregator() scanstates.Aggregator {
 	return scanStateAggregator
 }
 
-func SetScanStateAggregator(agg scanstates.Aggregator) {
-	initMutex.Lock()
-	defer initMutex.Unlock()
-	scanStateAggregator = agg
-}
-
 func ScanNotifier() scanner2.ScanNotifier {
 	initMutex.Lock()
 	defer initMutex.Unlock()
--- a/application/server/configuration.go
+++ b/application/server/configuration.go
@@ -41,6 +41,7 @@ import (
 	"github.com/snyk/snyk-ls/application/config"
 	"github.com/snyk/snyk-ls/application/di"
 	"github.com/snyk/snyk-ls/domain/ide/command"
+	"github.com/snyk/snyk-ls/domain/scanstates"
 	"github.com/snyk/snyk-ls/infrastructure/analytics"
 	ctx2 "github.com/snyk/snyk-ls/internal/context"
 	"github.com/snyk/snyk-ls/internal/notification"
@@ -191,9 +192,10 @@ func InitializeSettings(conf configuration.Configuration, engine workflow.Engine
 
 	processInitMetadata(conf, engine, logger, opts)
 	// global
-	processConfigSettings(conf, engine, logger, opts.Settings, analytics.TriggerSourceInitialize, resolver)
+	scanAgg := di.ScanStateAggregator()
+	processConfigSettings(conf, engine, logger, opts.Settings, analytics.TriggerSourceInitialize, resolver, scanAgg)
 	// folder
-	processFolderConfigs(conf, engine, logger, opts.FolderConfigs, analytics.TriggerSourceInitialize, resolver)
+	processFolderConfigs(conf, engine, logger, opts.FolderConfigs, analytics.TriggerSourceInitialize, resolver, scanAgg)
 }
 
 // UpdateSettings processes settings from workspace/didChangeConfiguration.
@@ -208,8 +210,9 @@ func UpdateSettings(conf configuration.Configuration, engine workflow.Engine, lo
 		}
 	}
 
-	processConfigSettings(conf, engine, logger, settings, triggerSource, configResolver)
-	processFolderConfigs(conf, engine, logger, folderConfigs, triggerSource, configResolver)
+	scanAgg := di.ScanStateAggregator()
+	processConfigSettings(conf, engine, logger, settings, triggerSource, configResolver, scanAgg)
+	processFolderConfigs(conf, engine, logger, folderConfigs, triggerSource, configResolver, scanAgg)
 
 	if ws != nil {
 		for _, folder := range ws.Folders() {
@@ -274,7 +277,7 @@ func validateLockedMachineFields(settings map[string]*types.ConfigSetting, confi
 
 // processConfigSettings writes incoming settings to configuration and applies side effects.
 // This replaces the old writeSettings + update* functions.
-func processConfigSettings(conf configuration.Configuration, engine workflow.Engine, logger *zerolog.Logger, settings map[string]*types.ConfigSetting, triggerSource analytics.TriggerSource, configResolver types.ConfigResolverInterface) {
+func processConfigSettings(conf configuration.Configuration, engine workflow.Engine, logger *zerolog.Logger, settings map[string]*types.ConfigSetting, triggerSource analytics.TriggerSource, configResolver types.ConfigResolverInterface, scanAgg scanstates.Aggregator) {
 	conf.ClearCache()
 
 	if len(settings) == 0 {
@@ -301,7 +304,7 @@ func processConfigSettings(conf configuration.Configuration, engine workflow.Eng
 	applyIssueViewOptions(conf, engine, logger, settings, triggerSource, configResolver)
 	applyDeltaFindings(conf, engine, logger, settings, triggerSource, configResolver)
 	applyAutoScan(conf, settings)
-	applyOrganization(conf, engine, logger, settings, triggerSource, configResolver)
+	applyOrganization(conf, engine, logger, settings, triggerSource, configResolver, scanAgg)
 	applyCliConfig(conf, settings)
 	applyEnvironment(conf, logger, settings)
 	applyCliBaseDownloadURL(conf, engine, logger, settings, triggerSource, configResolver)
@@ -348,7 +351,7 @@ func hasFilterChangesInLspConfig(lspConfig *types.LspFolderConfig) bool {
 }
 
 // processFolderConfigs handles the folder configuration portion of incoming settings.
-func processFolderConfigs(conf configuration.Configuration, engine workflow.Engine, logger *zerolog.Logger, folderConfigs []types.LspFolderConfig, triggerSource analytics.TriggerSource, configResolver types.ConfigResolverInterface) {
+func processFolderConfigs(conf configuration.Configuration, engine workflow.Engine, logger *zerolog.Logger, folderConfigs []types.LspFolderConfig, triggerSource analytics.TriggerSource, configResolver types.ConfigResolverInterface, scanAgg scanstates.Aggregator) {
 	notifier := di.Notifier()
 	incomingMap := buildIncomingLspConfigMap(folderConfigs)
 	allPaths := gatherAllFolderPathsFromLspConfigs(incomingMap, config.GetWorkspace(conf))
@@ -362,9 +365,13 @@ func processFolderConfigs(conf configuration.Configuration, engine workflow.Engi
 	var processedConfigs []types.FolderConfig
 	var changedConfigs []*types.FolderConfig
 	filterChanged := false
+	var orgChangedFolderPaths []types.FilePath
 
 	for path := range allPaths {
-		folderConfig, oldSnapshot, newSnapshot, configChanged := processSingleLspFolderConfig(conf, engine, logger, path, incomingMap, notifier)
+		folderConfig, oldSnapshot, newSnapshot, configChanged, orgSettingsChanged := processSingleLspFolderConfig(conf, engine, logger, path, incomingMap, notifier)
+		if orgSettingsChanged {
+			orgChangedFolderPaths = append(orgChangedFolderPaths, path)
+		}
 
 		if configChanged {
 			changedConfigs = append(changedConfigs, &folderConfig)
@@ -391,6 +398,10 @@ func processFolderConfigs(conf configuration.Configuration, engine workflow.Engi
 	// Trigger diagnostics republishing if filter changes detected
 	if filterChanged {
 		sendDiagnosticsForNewSettings(conf, logger)
+	}
+
+	if conf.GetBool(types.SettingIsLspInitialized) && len(orgChangedFolderPaths) > 0 {
+		resetSummaryPanelForOrgChange(scanAgg, orgChangedFolderPaths)
 	}
 
 	sendFolderConfigUpdateIfNeeded(conf, engine, logger, notifier, processedConfigs, len(changedConfigs) > 0, triggerSource, configResolver)
@@ -706,7 +717,7 @@ func applyAutoScan(conf configuration.Configuration, settings map[string]*types.
 	types.SetGlobalDeferredFolderScope(conf, types.SettingScanAutomatic, autoScan)
 }
 
-func applyOrganization(conf configuration.Configuration, engine workflow.Engine, logger *zerolog.Logger, settings map[string]*types.ConfigSetting, triggerSource analytics.TriggerSource, configResolver types.ConfigResolverInterface) {
+func applyOrganization(conf configuration.Configuration, engine workflow.Engine, logger *zerolog.Logger, settings map[string]*types.ConfigSetting, triggerSource analytics.TriggerSource, configResolver types.ConfigResolverInterface, scanAgg scanstates.Aggregator) {
 	v, ok := settingStr(settings, types.SettingOrganization)
 	if !ok {
 		return
@@ -717,25 +728,31 @@ func applyOrganization(conf configuration.Configuration, engine workflow.Engine,
 	newOrgId := types.GetGlobalOrganization(conf)
 	if oldOrgId != newOrgId && conf.GetBool(types.SettingIsLspInitialized) {
 		analytics.SendConfigChangedAnalytics(conf, engine, logger, configOrganization, oldOrgId, newOrgId, triggerSource, configResolver)
-		resetSummaryPanelForOrgChange(conf)
+		resetSummaryPanelForOrgChange(scanAgg, workspaceFolderPaths(conf))
 	}
 }
 
-// resetSummaryPanelForOrgChange clears the scan-state aggregator so the
+// resetSummaryPanelForOrgChange clears scan state for the given folders so the
 // Summary Panel returns to its initial "no scans yet" state. The aggregator's
-// Init re-emits a fresh snapshot, which IDE clients render as the empty
-// summary panel.
-func resetSummaryPanelForOrgChange(conf configuration.Configuration) {
+// Init re-emits a fresh snapshot, which IDE clients render as the empty summary panel.
+func resetSummaryPanelForOrgChange(scanAgg scanstates.Aggregator, folderPaths []types.FilePath) {
+	if scanAgg == nil || len(folderPaths) == 0 {
+		return
+	}
+	scanAgg.Init(folderPaths)
+}
+
+func workspaceFolderPaths(conf configuration.Configuration) []types.FilePath {
 	ws := config.GetWorkspace(conf)
 	if ws == nil {
-		return
+		return nil
 	}
 	folders := ws.Folders()
 	folderPaths := make([]types.FilePath, 0, len(folders))
 	for _, f := range folders {
 		folderPaths = append(folderPaths, f.Path())
 	}
-	di.ScanStateAggregator().Init(folderPaths)
+	return folderPaths
 }
 
 func applyCliConfig(conf configuration.Configuration, settings map[string]*types.ConfigSetting) {
@@ -984,7 +1001,7 @@ func gatherAllFolderPathsFromLspConfigs(incomingMap map[types.FilePath]types.Lsp
 // It loads the existing FolderConfig (unenriched), applies the LspFolderConfig updates, and returns
 // the processed config without persisting. The caller is responsible for batch-persisting all changes.
 // Returns: (processedConfig, oldSnapshot, newSnapshot, configChanged)
-func processSingleLspFolderConfig(conf configuration.Configuration, engine workflow.Engine, logger *zerolog.Logger, path types.FilePath, incomingMap map[types.FilePath]types.LspFolderConfig, notifier notification.Notifier) (types.FolderConfig, types.FolderConfigSnapshot, types.FolderConfigSnapshot, bool) {
+func processSingleLspFolderConfig(conf configuration.Configuration, engine workflow.Engine, logger *zerolog.Logger, path types.FilePath, incomingMap map[types.FilePath]types.LspFolderConfig, notifier notification.Notifier) (types.FolderConfig, types.FolderConfigSnapshot, types.FolderConfigSnapshot, bool, bool) {
 	subLogger := logger.With().Str("method", "processSingleLspFolderConfig").Str("path", string(path)).Logger()
 	resolver := di.ConfigResolver()
 	fc := config.GetUnenrichedFolderConfigFromEngine(engine, resolver, path, logger)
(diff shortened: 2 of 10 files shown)


application/server/configuration.go
```diff
@@ -717,9 +717,27 @@ func applyOrganization(conf configuration.Configuration, engine workflow.Engine,
...
+	ws := config.GetWorkspace(conf)
+	if ws == nil {
+		return
+	}
+	folders := ws.Folders()
+	folderPaths := make([]types.FilePath, 0, len(folders))
+	for _, f := range folders {
+		folderPaths = append(folderPaths, f.Path())
+	}
+	di.ScanStateAggregator().Init(folderPaths)
```
bastiandoetsch: **[high] All-folders over-reset + nil panic**

`resetSummaryPanelForOrgChange` is called once per folder-org change but collects `ws.Folders()` — every folder — and resets them all. If only folder 2 of 3 changes org, scan results for folders 1 and 3 are destroyed.

Additionally, `di.ScanStateAggregator()` may return `nil` before `di.Init()` completes (e.g. a `workspace/didChangeConfiguration` arriving during startup), causing a nil-deref panic here.

Suggested fixes:
1. Pass only the changed `folderPath` into this function and call `Init([]types.FilePath{changedFolderPath})`.
2. Nil-guard: `if agg := di.ScanStateAggregator(); agg != nil { agg.Init(folderPaths) }`.

application/server/configuration.go
```diff
@@ -1083,6 +1101,10 @@ func updateFolderOrgIfNeeded(conf configuration.Configuration, engine workflow.E
 		if folder != nil {
 			di.LdxSyncService().RefreshConfigFromLdxSync(context.Background(), conf, engine, logger, []types.Folder{folder}, notifier)
 		}
+		if conf.GetBool(types.SettingIsLspInitialized) {
+			// orgSettingsChanged matches folderConfigsOrgSettingsEqual (PreferredOrg, OrgSetByUser, AutoDeterminedOrg).
+			resetSummaryPanelForOrgChange(conf)
```
bastiandoetsch: **[medium] N folder changes → N `Init()` calls → N LSP notifications**

`resetSummaryPanelForOrgChange` is called inside the per-folder loop. With N folders changing org simultaneously, the aggregator resets N times and the IDE receives N consecutive "everything is NotStarted" snapshots.

The pattern 8 lines below in `processFolderConfigs` (the `filterChanged` → `sendDiagnosticsForNewSettings` approach) is the right model: collect a boolean `anyOrgChanged` across the loop, then call the reset once after the loop completes.

application/di/init.go
```diff
@@ -247,6 +247,12 @@ func ScanStateAggregator() scanstates.Aggregator {
 	return scanStateAggregator
 }
 
+func SetScanStateAggregator(agg scanstates.Aggregator) {
```
bastiandoetsch: **[high] Exported DI setter breaks write-once invariant and creates stale references**

Every other dependency in this file is captured at construction time and not swappable post-`Init()`. This setter mutates only the package-level variable; the `scanner`, `workspace`, and `folder` instances each hold the aggregator by-interface from when they were constructed and are unaffected by the swap.

Consequence in tests: the test asserts on `di.ScanStateAggregator()` (the swapped-in `realAgg`) while the scanner goroutine writes scan events to the original noop captured at `TestInit()`. A race where a scan goroutine writes stale data after `Init()` resets is invisible — the write goes to a disconnected object.

Suggested fix: remove this setter and inject the aggregator as a parameter through `applyOrganization` / `updateFolderOrgIfNeeded`. Tests supply the real aggregator via parameter; production code calls `di.ScanStateAggregator()` at the call site. No singleton mutation needed.

application/server/ldx_sync_smoke_test.go
```diff
@@ -93,6 +93,28 @@ func requireLspConfigurationNotification(t *testing.T, jsonRpcRecorder *testsupp
...
+	require.NotNil(t, fc.Settings[types.SettingSnykCodeEnabled], "folder settings must include snyk_code_enabled")
+	require.NotNil(t, fc.Settings[types.SettingSnykCodeEnabled].Value, "snyk_code_enabled value must be set (true or false)")
+	require.NotNil(t, fc.Settings[types.SettingSnykIacEnabled], "folder settings must include snyk_iac_enabled")
+	require.NotNil(t, fc.Settings[types.SettingSnykIacEnabled].Value, "snyk_iac_enabled value must be set (true or false)")
+}
+
+// assertSmokeLdxFolderOrgResolution accepts auto_determined_org from LDX-Sync when present.
+// When LDX-Sync returns no organizations for the project, the field is omitted from the
+// notification and scans fall back to the global organization (see ldx_sync_service_test NoMapping).
+func assertSmokeLdxFolderOrgResolution(t *testing.T, engine workflow.Engine, folder types.FilePath, fc types.LspFolderConfig) {
```
bastiandoetsch: **[medium] Smoke test no longer proves LDX-Sync org resolution works**

The previous assertion (`require.NotNil(t, fc.Settings[SettingAutoDeterminedOrg])` with `lspFolderConfigWaitForAutoDeterminedOrg()`) polled until LDX-Sync actually returned an org. The replacement silently passes when `auto_determined_org` is absent, because `config.FolderOrganization()` always returns a non-empty string (it falls back to the global org).

A real regression in LDX-Sync org resolution — the cache never populates, the notification drops the field — would not be caught.

If the CI environment genuinely has no LDX-Sync mapping for this test repo, prefer an explicit skip with a comment rather than a silent fallback:
```go
if !configSettingHasNonEmptyStringValue(fc.Settings[types.SettingAutoDeterminedOrg]) {
    t.Skip("CI has no LDX-Sync org mapping for this repo; skipping auto_determined_org assertion")
}
```

bastiandoetsch (commented):
Verification findings from IDE-1969 review — 3 high / 4 medium issues. Comments below are on the specific lines.

application/server/configuration.go
```diff
@@ -717,7 +728,31 @@ func applyOrganization(conf configuration.Configuration, engine workflow.Engine,
 	newOrgId := types.GetGlobalOrganization(conf)
 	if oldOrgId != newOrgId && conf.GetBool(types.SettingIsLspInitialized) {
 		analytics.SendConfigChangedAnalytics(conf, engine, logger, configOrganization, oldOrgId, newOrgId, triggerSource, configResolver)
+		resetSummaryPanelForOrgChange(scanAgg, workspaceFolderPaths(conf))
```
bastiandoetsch: [Should Fix] `applyOrganization` calls `resetSummaryPanelForOrgChange` for all workspace folders (line 731), and then `processFolderConfigs` can call it again for the same folders (line 404) when `folderConfigsOrgSettingsEqual` detects a change — two sequential `Init` calls emit two `Emit` snapshots to the IDE for a single `UpdateSettings` call, which can cause a visible double-flash in the Summary Panel.

Fix: remove the `scanAgg` parameter from `applyOrganization` and `processConfigSettings`; move the global-org reset into `processFolderConfigs` by mapping the global org change to all workspace folder paths there, so a single `resetSummaryPanelForOrgChange` call covers both paths.

— AI review

bastiandoetsch (commented):

application/server/ldx_sync_smoke_test.go
```diff
@@ -93,6 +93,26 @@ func requireLspConfigurationNotification(t *testing.T, jsonRpcRecorder *testsupp
...
+	require.NotNil(t, fc.Settings[types.SettingSnykCodeEnabled].Value, "snyk_code_enabled value must be set (true or false)")
+	require.NotNil(t, fc.Settings[types.SettingSnykIacEnabled], "folder settings must include snyk_iac_enabled")
+	require.NotNil(t, fc.Settings[types.SettingSnykIacEnabled].Value, "snyk_iac_enabled value must be set (true or false)")
+}
+
+// assertSmokeLdxFolderOrgResolution verifies LDX-Sync populated auto_determined_org on the folder config.
+func assertSmokeLdxFolderOrgResolution(t *testing.T, fc types.LspFolderConfig) {
+	t.Helper()
+	if !configSettingHasNonEmptyStringValue(fc.Settings[types.SettingAutoDeterminedOrg]) {
+		t.Skip("CI has no LDX-Sync org mapping for this repo; skipping auto_determined_org assertion")
```
bastiandoetsch: [Suggestion] `assertSmokeLdxFolderOrgResolution` silently skips the `auto_determined_org` assertion in any CI environment without an LDX-Sync org mapping, meaning this PR's specific feature (resetting the panel on org change driven by LDX-Sync) is never enforced in CI.

Fix: guard the skip with an explicit environment variable (e.g., `testutil.RequiresLdxSync(t)`) at the test-function level so the skip is auditable in CI logs rather than silently swallowed inside a shared helper.

— AI review

application/server/configuration.go
```diff
@@ -966,7 +1001,7 @@ func gatherAllFolderPathsFromLspConfigs(incomingMap map[types.FilePath]types.Lsp
 // It loads the existing FolderConfig (unenriched), applies the LspFolderConfig updates, and returns
 // the processed config without persisting. The caller is responsible for batch-persisting all changes.
 // Returns: (processedConfig, oldSnapshot, newSnapshot, configChanged)
-func processSingleLspFolderConfig(conf configuration.Configuration, engine workflow.Engine, logger *zerolog.Logger, path types.FilePath, incomingMap map[types.FilePath]types.LspFolderConfig, notifier notification.Notifier) (types.FolderConfig, types.FolderConfigSnapshot, types.FolderConfigSnapshot, bool) {
+func processSingleLspFolderConfig(conf configuration.Configuration, engine workflow.Engine, logger *zerolog.Logger, path types.FilePath, incomingMap map[types.FilePath]types.LspFolderConfig, notifier notification.Notifier) (types.FolderConfig, types.FolderConfigSnapshot, types.FolderConfigSnapshot, bool, bool) {
```
bastiandoetsch: [Suggestion] `processSingleLspFolderConfig` now returns a 5-tuple — this is the second time the return signature has grown. All callers must destructure all 5 positions, and `orgSettingsChanged` at position 5 is easy to miss or swap.

Fix: define a small result struct with named fields.
```go
type singleFolderResult struct {
    config            types.FolderConfig
    oldSnapshot       types.FolderConfigSnapshot
    newSnapshot       types.FolderConfigSnapshot
    configChanged     bool
    orgSettingsChanged bool
}
```

— AI review

application/server/configuration.go
```diff
@@ -717,7 +728,31 @@ func applyOrganization(conf configuration.Configuration, engine workflow.Engine,
...
 	if oldOrgId != newOrgId && conf.GetBool(types.SettingIsLspInitialized) {
 		analytics.SendConfigChangedAnalytics(conf, engine, logger, configOrganization, oldOrgId, newOrgId, triggerSource, configResolver)
+		resetSummaryPanelForOrgChange(scanAgg, workspaceFolderPaths(conf))
+	}
+}
+
+// resetSummaryPanelForOrgChange clears scan state for the given folders so the
+// Summary Panel returns to its initial "no scans yet" state. The aggregator's
+// Init re-emits a fresh snapshot, which IDE clients render as the empty summary panel.
+func resetSummaryPanelForOrgChange(scanAgg scanstates.Aggregator, folderPaths []types.FilePath) {
```
bastiandoetsch, line 765: [Suggestion] `resetSummaryPanelForOrgChange` has 66.7% coverage — the `scanAgg == nil || len(folderPaths) == 0` guard branch is not tested. A single test would close the gap and document the boundary contract.
```go
func TestResetSummaryPanelForOrgChange_NilAgg_DoesNotPanic(t *testing.T) {
    assert.NotPanics(t, func() {
        resetSummaryPanelForOrgChange(nil, []types.FilePath{"/some/path"})
    })
}
```

— AI review

scripts/check-tests-run.sh
```diff
@@ -8,8 +8,8 @@
 set -e
 
 HASH_FILE=".tests-hash"
-REQUIRED_STAGES=("test" "test-integ" "test-smoke")   # block push if stale
-ADVISORY_STAGES=()                                   # warn only — takes 30-90 min
+REQUIRED_STAGES=("test")   # block push if stale
+ADVISORY_STAGES=("test-integ" "test-smoke")                                   # warn only — takes 30-90 min
```
bastiandoetsch, line 15: [Suggestion] Moving `test-integ` and `test-smoke` to `ADVISORY_STAGES` means a developer can push with failing integration or smoke tests — the pre-push hook will warn but not block. The comment notes the 30-90 min runtime saving but not the quality-gate trade-off.

Fix: add a comment documenting that these stages are enforced by CI (not the pre-push hook) so the intent is clear to future readers.

— AI review

bastiandoetsch:
[Suggestion] `applyOrganization` is a configuration-application function that now directly resets the scan-state aggregator — a UI/domain concern. The `processConfigSettings` function signature carries the aggregator solely for this one sub-function, polluting 5+ function signatures with a dependency that has nothing to do with parsing and persisting configuration values. If a future org-change source (e.g., LDX-Sync itself) triggers a config update through a different code path, the reset won't fire because it is coupled to `applyOrganization`.

Fix: consider a post-config hook or observer registered at the server level that detects org changes by comparing before/after state, keeping the configuration application layer unaware of scan-state reset logic. This is a follow-up refactor candidate, not a ship blocker for this PR.

— AI review

application/server/configuration.go
acke: Will fix: drop `scanAgg` from `applyOrganization`/`processConfigSettings` and consolidate the org-change reset into `processFolderConfigs` so a single `resetSummaryPanelForOrgChange` covers both global and per-folder paths.

application/server/ldx_sync_smoke_test.go
acke: Will fix: add an explicit `testutil.RequiresLdxSync(t)` guard at the test-function level so the skip is logged as a clear precondition instead of being silently swallowed inside `assertSmokeLdxFolderOrgResolution`.

application/server/configuration.go
acke: Will fix: replace the 5-tuple return with a named `singleFolderResult` struct so callers reference fields by name instead of positional destructuring.

application/server/configuration.go
acke: Will fix: add `TestResetSummaryPanelForOrgChange_NilAgg_DoesNotPanic` (and an empty-folderPaths case) to cover the guard branch.

scripts/check-tests-run.sh
acke: Will fix: extend the comment above `ADVISORY_STAGES` in `scripts/check-tests-run.sh` to call out that `test-integ` and `test-smoke` are advisory locally because CI enforces them as the blocking quality gate.

acke:
Re: https://github.com/snyk/snyk-ls/pull/1282#issuecomment-4507734374

Partially addressed by the fix for https://github.com/snyk/snyk-ls/pull/1282#discussion_r3280780041 — removing `scanAgg` from `applyOrganization`/`processConfigSettings` drops the aggregator from those signatures. The broader observer/post-config-hook decoupling is a worthwhile follow-up; I'll open a separate ticket rather than expand the scope of this PR.

application/server/ldx_sync_smoke_test.go
acke: Update — changed approach. Rather than guarding with a new `testutil.RequiresLdxSync(t)` env-var skip, the silent skip inside `assertSmokeLdxFolderOrgResolution` was removed outright. The LDX-Sync smoke tests now assert `auto_determined_org` unconditionally — we want them to actually run and fail loudly if the mapping isn't there, not skip auditably.

application/server/configuration.go
bastiandoetsch: Fixed at configuration.go:421. `scanAgg` removed from `applyOrganization`/`processConfigSettings`; the single `resetSummaryPanelForOrgChange` call in `processFolderConfigs` now covers both global and per-folder org changes. No further action needed. — AI review

application/server/ldx_sync_smoke_test.go
bastiandoetsch: The changed approach — removing the silent fallback entirely and asserting `auto_determined_org` unconditionally — is valid and more rigorous than the original `RequiresLdxSync` skip suggestion. The test now fails loudly if LDX-Sync resolution is broken. Finding closed. — AI review

application/server/configuration.go
bastiandoetsch: Fixed. `singleFolderResult` struct implemented with named fields. No further action needed. — AI review

application/server/configuration.go
bastiandoetsch: Fixed. `TestResetSummaryPanelForOrgChange_NilAgg_DoesNotPanic` and the empty-paths case cover the guard branch. No further action needed. — AI review

scripts/check-tests-run.sh
bastiandoetsch: Fixed. Comment added at `scripts/check-tests-run.sh:11-15` documenting that CI is the blocking quality gate for integ/smoke. No further action needed. — AI review

bastiandoetsch:
Re: https://github.com/snyk/snyk-ls/pull/1282#issuecomment-4507734374

Acknowledged. Removing `scanAgg` from `applyOrganization`/`processConfigSettings` addresses the immediate signature pollution. The broader observer/post-config-hook decoupling is a valid follow-up. Finding closed. — AI review

application/server/configuration.go
```diff
@@ -1073,7 +1149,7 @@ func temporarilyApplyNewOrgForValidation(conf configuration.Configuration, folde
 	}
 }
 
-func updateFolderOrgIfNeeded(conf configuration.Configuration, engine workflow.Engine, logger *zerolog.Logger, fc *types.FolderConfig, folderConfig *types.FolderConfig, oldSnapshot types.FolderConfigSnapshot, notifier notification.Notifier) {
+func updateFolderOrgIfNeeded(conf configuration.Configuration, engine workflow.Engine, logger *zerolog.Logger, fc *types.FolderConfig, folderConfig *types.FolderConfig, oldSnapshot types.FolderConfigSnapshot, notifier notification.Notifier) bool {
```
bastiandoetsch, line 1155: [Suggestion] `updateFolderOrgIfNeeded` intentionally omits the `SettingIsLspInitialized` guard used by its callers, but this is undocumented. A future contributor could add the guard inside this function thinking it is missing; that would prevent per-folder org changes from populating `orgChangedFolderPaths`, silently breaking the Summary Panel reset for per-folder org changes after LSP init.

Fix: add a one-line comment explaining the intent.
```go
// No SettingIsLspInitialized guard here — the outer check in processFolderConfigs
// (line 421) prevents the summary panel reset before init. This function intentionally
// runs unconditionally so that LDX-Sync refresh and orgChangedFolderPaths are always populated.
func updateFolderOrgIfNeeded(...) bool {
```

— AI review

bastiandoetsch (commented):

bastiandoetsch (approved):

PR #1286, opened 2026-05-18
chore(scripts): require only unit tests in pre-push check [IDE-2042]

acke:
### Description

Relaxes the pre-push `check-tests` hook so only `make test` (unit tests) blocks push. Integration and smoke stages are **advisory** (warn only).

Fixes the friction where `test-integ` and `test-smoke` must be recorded in `.tests-hash` before every push, even though those suites take 30–90 minutes and run in CI.

**Change** (`scripts/check-tests-run.sh` only):

```bash
REQUIRED_STAGES=("test")
ADVISORY_STAGES=("test-integ" "test-smoke")
```

Jira: [IDE-2042](https://snyksec.atlassian.net/browse/IDE-2042)

### Checklist

- [x] Linted (`make lint-fix` via pre-commit)
- [ ] Tests added and all succeed — hook-only change

### Test plan

- [ ] `make test` at HEAD → push succeeds
- [ ] Push without integ/smoke recorded → advisory warnings, no block
- [ ] Push without `make test` → blocked

[IDE-2042]: https://snyksec.atlassian.net/browse/IDE-2042

### DIFF
--- a/scripts/check-tests-run.sh
+++ b/scripts/check-tests-run.sh
@@ -2,14 +2,36 @@
 # Pre-push hook: verify that the required test stages have been recorded for
 # the current HEAD commit. Adapted from ldx-sync/scripts/tests-run-commit-hook.sh.
 #
-# Stage list coupling: REQUIRED_STAGES and ADVISORY_STAGES here must stay in sync
-# with the stages written by the Makefile _save-test-hash target (make test,
-# make test-all). Update both files when adding a new stage.
+# Stage list coupling: REQUIRED_STAGES and smoke shard stage names here must stay
+# in sync with the Makefile _save-test-hash targets (make test, make test-all,
+# _smoke-shard-N, test-smoke-parallel). Update both files when adding a new stage.
 set -e
 
 HASH_FILE=".tests-hash"
-REQUIRED_STAGES=("test" "test-integ" "test-smoke")   # block push if stale
-ADVISORY_STAGES=()                                   # warn only — takes 30-90 min
+REQUIRED_STAGES=("test" "test-integ")   # block push if stale
+SMOKE_STAGE="test-smoke"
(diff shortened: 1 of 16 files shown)


scripts/check-tests-run.sh
```diff
@@ -2,14 +2,36 @@
...
-# make test-all). Update both files when adding a new stage.
+# Stage list coupling: REQUIRED_STAGES and smoke shard stage names here must stay
+# in sync with the Makefile _save-test-hash targets (make test, make test-all,
+# _smoke-shard-N, test-smoke-parallel). Update both files when adding a new stage.
 set -e
 
 HASH_FILE=".tests-hash"
-REQUIRED_STAGES=("test" "test-integ" "test-smoke")   # block push if stale
-ADVISORY_STAGES=()                                   # warn only — takes 30-90 min
+REQUIRED_STAGES=("test" "test-integ")   # block push if stale
```
bastiandoetsch: Why the shards of we don't use the required stages?

scripts/check-tests-run.sh
```diff
@@ -66,13 +87,39 @@ if [ ${#outdated[@]} -gt 0 ]; then
     echo "❌ Commits since these stages last ran:"
     for s in "${outdated[@]}"; do
         echo "   make $s"
-        [ "$s" = "test-smoke" ] && echo "   (or faster: make test-smoke-parallel)"
     done
     blocked=1
 fi
 
+if ! smoke_satisfied; then
```
bastiandoetsch: This seems to break the generic approach to check for stages by hard coding smoke test logic.

scripts/check-tests-run.sh
acke: yes, was waiting for test to push the cleanup.

bastiandoetsch (approved):

PR #1289, opened 2026-05-19
fix(secrets): return errors on skip so scan analytics are not sent [I…

acke:
When secrets scanning is skipped (no token or org feature flag off), return an error instead of nil so ProcessResults does not emit a successful "Scan done" analytics event for a scan that never ran.

### Description

### Checklist


rrama (approved):
LGTM, hopefully it can make it in 🤞

PR #1293, opened 2026-05-23
fix(snyk-ls): dedupe locked-settings warning into one notification [IDE-1970]

acke:
## Summary
- Collapses the N+1 locked-settings notifications previously fired per `workspace/didChangeConfiguration` into exactly one warning per triggering event, listing every affected field (deduplicated and sorted) — see [IDE-1970](https://snyksec.atlassian.net/browse/IDE-1970).
- Refactors `validateLockedMachineFields` / `validateLockedFields` from `bool` to `[]string` so the helper `notifyLockedFieldsRejected` can fold machine-scope and folder-scope rejections into a single `SendShowMessage` call.
- Existing inline `SendShowMessage` calls in `processConfigSettings` and `processSingleLspFolderConfig` are removed; the single notification is now emitted at the end of `UpdateSettings` and `InitializeSettings`.

## Why
With multiple workspace folders, every folder that had at least one locked-field rejection produced its own popup, and the machine-scope path emitted a second generic popup. A single org policy push that locked one field across 4 folders would surface 5 popups for the same event — the bug described in IDE-1970's AC ("exactly one notification per triggering event regardless of number of locked fields").

## Test plan
- [x] `go vet ./...`
- [x] `go test ./application/server/...` (full suite passes, 74s)
- [x] New `Test_notifyLockedFieldsRejected_DeduplicatesAcrossGroupsAndEmitsOnce` covers: dedup across machine + folder scope, no-emit on empty, single-notification across multiple folders with overlapping locks.
- [x] Updated `Test_validateLockedFields_UsesNewOrgPolicyOnOrgSwitch` to assert on the returned rejected-field names (the new contract).
- [ ] Manual: trigger an org policy push that locks a setting affecting multiple workspace folders and confirm exactly one warning popup appears in the IDE.

[IDE-1970]: https://snyksec.atlassian.net/browse/IDE-1970?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

___

### **PR Type**
Bug fix

___

### **Description**
- Deduplicate locked settings notifications into one.

- Display setting names using their registered display names.

- Add comprehensive end-to-end test for notification logic.

___

<tr>
  <td>
    <details>
      <summary><strong>configuration.go</strong><dd><code>Refactor Locked Settings Validation and Notification Logic</code></dd></summary>
<hr>

application/server/configuration.go

<ul><li>Refactored validation functions (<code>validateLockedMachineFields</code>, <br><code>validateLockedFields</code>) to return rejected setting names as a slice <br>instead of a boolean.<br> <li> Introduced <code>notifyLockedFieldsRejected</code> to consolidate, deduplicate, and <br>format notifications for all rejected settings, using display names.<br> <li> Modified <code>InitializeSettings</code>, <code>UpdateSettings</code>, <code>processConfigSettings</code>, <br><code>processFolderConfigs</code>, and <code>processSingleLspFolderConfig</code> to leverage the <br>new notification logic and propagate rejected fields.<br> <li> Added helper functions <code>configOptionsMetaData</code> and <code>displayNameFor</code> for <br>robust notification message generation.</ul>

</details>

  </td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1293/changes#diff-eb3fe6fd1956591838595b08979653ec23c0b0c16f5340e9e89a05c30e921607">+118/-43</a></td>

</tr>
<tr>
  <td>
    <details>
      <summary><strong>configuration_test.go</strong><dd><code>Add Tests for Locked Settings Notification Deduplication</code>&nbsp; </dd></summary>
<hr>

application/server/configuration_test.go

<ul><li>Added new unit and end-to-end tests <br>(<code>Test_notifyLockedFieldsRejected_DeduplicatesAcrossGroupsAndEmitsOnce</code>, <br><code>Test_UpdateSettings_LockedFields_EmitsExactlyOneNotification</code>) to <br>verify the correctness of the deduplicated notification logic for <br>locked settings.<br> <li> Updated existing tests to assert on the returned slice of rejected <br>fields rather than a boolean, matching the new function signatures.</ul>

</details>

  </td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1293/changes#diff-e141faafb3f0507956b85e92ac3b584c7c706bd96e7f428c7ef95d0d8b2df564">+255/-2</a>&nbsp; </td>

</tr>

</details>

___


application/server/configuration_test.go
```diff
@@ -2257,3 +2261,252 @@ func TestResetSummaryPanelForOrgChange_EmptyFolderPaths_DoesNotCallInit(t *testi
...
+
+	resolver := testutil.DefaultConfigResolver(engine)
+	di.SetConfigResolver(resolver)
+
+	// Subscribe to the real notifier before driving the change so we don't miss
+	// any messages produced on the request goroutine.
+	var (
+		mu              sync.Mutex
+		showMessages    []sglsp.ShowMessageParams
+		listenerStarted = make(chan struct{})
```
bastiandoetsch: [Should Fix] `listenerStarted` is allocated and closed inside the listener callback but is never awaited anywhere in the test — it is dead code that misleads readers into thinking it synchronises something.

Fix: delete `listenerStarted` and the `select` block around it. `require.Eventually` on line 2485 already provides the correct synchronisation.

— AI review

application/server/configuration_test.go
```diff
@@ -2257,3 +2261,252 @@ func TestResetSummaryPanelForOrgChange_EmptyFolderPaths_DoesNotCallInit(t *testi
...
+	require.Eventually(t, func() bool {
+		mu.Lock()
+		defer mu.Unlock()
+		return len(showMessages) >= 1
+	}, 2*time.Second, 10*time.Millisecond,
+		"expected at least one ShowMessage notification after UpdateSettings with locked fields")
+
+	// Settle window: if any additional notifications were going to fire
+	// (the pre-IDE-1970 bug), they would arrive within this period.
+	time.Sleep(200 * time.Millisecond)
```
bastiandoetsch: [Should Fix] `time.Sleep(200 * time.Millisecond)` is a probabilistic settle window. On a slow CI runner extra notifications may arrive after 200 ms (false negative); on a fast one 200 ms may expire before the goroutine has dispatched them (false positive). The `notifier-async-race` brain note explicitly calls out this pattern.

Fix: because `notifyLockedFieldsRejected` is synchronous, all notifications are enqueued before `UpdateSettings` returns. Assert the count with a single `require.Eventually` and remove the sleep; or close the notifier after the call and drain with a select with timeout.

— AI review

application/server/configuration.go
```diff
@@ -268,32 +275,89 @@ func validateLockedMachineFields(settings map[string]*types.ConfigSetting, confi
...
+func notifyLockedFieldsRejected(notifier notification.Notifier, fm workflow.ConfigurationOptionsMetaData, rejectedFieldGroups ...[]string) {
+	seen := make(map[string]struct{})
+	var displayNames []string
+	for _, group := range rejectedFieldGroups {
+		for _, name := range group {
+			if _, ok := seen[name]; ok {
+				continue
+			}
+			seen[name] = struct{}{}
+			displayNames = append(displayNames, displayNameFor(fm, name))
```
bastiandoetsch: [Should Fix] `seen` deduplicates raw setting keys but the resolved display name is appended to `displayNames` without a second dedup pass. If two distinct raw keys resolve to the same display-name string via `displayNameFor`, that string will appear twice in the user-facing notification.

Fix: add a `seenDisplay` map and skip appending when the resolved name is already present, or run a dedup pass on `displayNames` after sorting.

```go
seenDisplay := map[string]struct{}{}
// ...
dn := displayNameFor(fm, name)
if _, ok := seenDisplay[dn]; ok {
    continue
}
seenDisplay[dn] = struct{}{}
displayNames = append(displayNames, dn)
```

— AI review

application/server/configuration.go
```diff
@@ -1065,22 +1137,25 @@ func processSingleLspFolderConfig(conf configuration.Configuration, engine workf
...
 // validateLockedFields checks if any fields in the incoming LspFolderConfig are locked by LDX-Sync.
-// Returns true if any fields were rejected due to being locked.
+// Returns the names of rejected (locked) settings so the caller can fold them
+// into the single deduplicated locked-fields notification per triggering event
+// (see [IDE-1970]).
 // If the incoming update changes PreferredOrg, locks are evaluated against the NEW org's policies
 // to prevent bypassing stricter locks during an org switch.
-func validateLockedFields(conf configuration.Configuration, folderConfig *types.FolderConfig, incoming *types.LspFolderConfig, subLogger *zerolog.Logger) bool {
+func validateLockedFields(conf configuration.Configuration, folderConfig *types.FolderConfig, incoming *types.LspFolderConfig, subLogger *zerolog.Logger) []string {
 	resolver := di.ConfigResolver()
```
bastiandoetsch: [Should Fix] `validateLockedFields` fetches the resolver via `di.ConfigResolver()` on this line while `validateLockedMachineFields` receives it as a parameter. A caller that injects a non-DI resolver via `UpdateSettings`'s `configResolver` parameter gets different resolver instances for machine-scope and folder-scope validation in the same triggering event.

Fix: add a `resolver types.ConfigResolverInterface` parameter to `validateLockedFields` and pass it from `processSingleLspFolderConfig` (which already has access via the chain that calls `validateLockedMachineFields` with the injected resolver).

— AI review

application/server/configuration.go
```diff
@@ -268,32 +275,89 @@ func validateLockedMachineFields(settings map[string]*types.ConfigSetting, confi
...
+	notifier.SendShowMessage(sglsp.MTWarning,
+		fmt.Sprintf("Failed to update some settings: locked by your organization's policy (%s)",
+			strings.Join(displayNames, ", ")))
+}
+
+// configOptionsMetaData safely extracts ConfigurationOptionsMetaData from a
+// possibly-nil ConfigResolverInterface. Returns nil when the resolver is nil so
+// callers can still emit notifications using raw setting identifiers as a
+// fallback (e.g. in tests without a fully-wired resolver).
+func configOptionsMetaData(resolver types.ConfigResolverInterface) workflow.ConfigurationOptionsMetaData {
```
bastiandoetsch: [Suggestion] `configOptionsMetaData` is a nil-guard wrapper with exactly two call sites. Its only effect — returning `nil` when `resolver` is nil — is already handled inside `displayNameFor`, which accepts a nil `fm`. The extra private function adds a level of indirection for a one-liner.

Fix: delete `configOptionsMetaData` and inline the nil guard at the two call sites, or pass `nil` directly and rely on `displayNameFor`'s existing nil check.

— AI review

application/server/configuration.go
```diff
@@ -395,6 +465,10 @@ func processFolderConfigs(conf configuration.Configuration, engine workflow.Engi
 			changedConfigs = append(changedConfigs, &cfg)
 		}
 
+		if len(result.rejectedLockedFields) > 0 {
```
bastiandoetsch: [Suggestion] The `if len(result.rejectedLockedFields) > 0` guard before `append` is redundant. Go's `append(slice, []string{}...)` is a no-op; the guard adds noise without altering behaviour.

Fix: remove the `if` guard and keep only the `append` line.

— AI review

application/server/configuration_test.go
```diff
@@ -2257,3 +2261,252 @@ func TestResetSummaryPanelForOrgChange_EmptyFolderPaths_DoesNotCallInit(t *testi
...
+		msg := mock.SentMessages()[0].(sglsp.ShowMessageParams)
+		// 'severity_filter_critical' < 'snyk_code_enabled', and snyk_code_enabled
+		// appears only once despite being rejected by two folders.
+		assert.Equal(t,
+			"Failed to update some settings: locked by your organization's policy (severity_filter_critical, snyk_code_enabled)",
+			msg.Message,
+			"flattened multi-folder rejections must still dedupe and sort")
+	})
+
+	t.Run("flat group ordering is irrelevant; sort is deterministic", func(t *testing.T) {
```
bastiandoetsch: [Suggestion] The "flat group ordering is irrelevant; sort is deterministic" subtest tests a property already guaranteed by `sort.Strings` and already covered by the preceding subtests. It exercises no independent behavioural contract.

Fix: remove this subtest. The three remaining subtests cover the same correctness property.

— AI review

bastiandoetsch:
[Suggestion] Coverage gap: the nil-resolver and empty-settings early returns in `validateLockedMachineFields` are not directly exercised by the new locked-fields unit tests. Under the full integration suite these paths are hit, but a targeted unit test would catch regressions faster.

Fix: add two sub-tests to the `notifyLockedFieldsRejected` table (or a separate `validateLockedMachineFields` test) for: (a) nil `configResolver` → expect `nil` return, (b) empty `settings` map → expect `nil` return.

— AI review

bastiandoetsch (commented):

application/server/configuration.go
```diff
@@ -192,9 +192,11 @@ func InitializeSettings(conf configuration.Configuration, engine workflow.Engine
 
 	processInitMetadata(conf, engine, logger, opts)
 	// global
-	globalOrgChanged := processConfigSettings(conf, engine, logger, opts.Settings, analytics.TriggerSourceInitialize, resolver)
+	globalOrgChanged, rejectedMachineFields := processConfigSettings(conf, engine, logger, opts.Settings, analytics.TriggerSourceInitialize, resolver)
```
bastiandoetsch: I'd call them `locked` instead of `rejected`. That's more consistent with our usual naming scheme.

bastiandoetsch (approved):
approved, but changes would be nice.

application/server/configuration_test.go
acke: Replaced `time.Sleep(200ms) + require.Len` with `require.Never(len > 1, 200ms, 10ms)`. Same upper bound but fails fast if a second message arrives, and uses an explicit assertion API. `require.Never` is already established in the codebase (`precedence_smoke_test.go`, `initializer_test.go`). Ran the test ×3 — green.

application/server/configuration.go
acke: Switched dedup from raw setting identifier to resolved display name in `notifyLockedFieldsRejected`. Two raw keys that resolve to the same display name (or the same raw fallback when no annotation is registered) now collapse to one entry. Added subtest `two raw keys that resolve to the same display name collapse to one entry` with a stub `ConfigurationOptionsMetaData` that maps two raw keys to the same display name — pins the new behavior.

application/server/configuration.go
acke: Threaded the injected `configResolver` from `processFolderConfigs` → `processSingleLspFolderConfig` → `validateLockedFields` so machine-scope and folder-scope validation share the same resolver instance. `validateLockedFields` now takes a `resolver types.ConfigResolverInterface` parameter instead of calling `di.ConfigResolver()`. Same change for the `GetUnenrichedFolderConfigFromEngine` call inside `processSingleLspFolderConfig`. Updated 3 test callsites to pass the resolver explicitly. All tests green.

application/server/configuration.go
acke: Deleted `configOptionsMetaData`. Inlined the nil check at both call sites (`InitializeSettings`, `UpdateSettings`) — 3 lines each, but the indirection is gone. `displayNameFor` still handles `nil fm` so the fallback to raw identifiers is preserved.

application/server/configuration.go
acke: Removed. `append(s, []string{}...)` is a no-op so the guard was noise.

application/server/configuration_test.go
acke: Removed. The exact-message assertions in the preceding subtests pin the sorted output, so any argument-order sensitivity would be caught.

application/server/configuration.go
acke: Renamed throughout: `rejectedMachineFields → lockedMachineFields`, `rejectedFolderFields → lockedFolderFields`, `rejectedLockedFields → lockedFields` (struct field + locals), `rejectedFieldGroups → lockedFieldGroups`, and the small `rejected` locals inside the two validators. Function name `notifyLockedFieldsRejected` left as-is since "rejected" there describes the action (the change was rejected), not the field state.

acke:
Re: https://github.com/snyk/snyk-ls/pull/1293#issuecomment-4555800254

Added `Test_validateLockedMachineFields_EarlyReturns` with two subtests:
- `nil configResolver returns nil without panicking` — passes nil and asserts the settings map is left untouched
- `empty settings map returns nil without consulting the resolver` — uses `mock_types.NewMockConfigResolverInterface` with **no `EXPECT()` calls**, so any resolver invocation would fail the test, pinning the contract that the early return must not touch the resolver

Both green.

PR #1295, opened 2026-05-26
fix(settings-fallback): send only changed fields; fix validation wiring [IDE-2072]

nick-y-snyk:
## Summary

Fixes four bugs in settings-fallback.html — the self-contained fallback page shown by all Snyk IDE plugins before the LS becomes available.

**Bug 2 — Full snapshot on save (Eclipse root cause):** collectData() sent all 6 fields on every save. Eclipse marked all as explicitlyChanged, suppressing LS/org-policy defaults. Fix: collectChangedData() diffs against captureInitialState() snapshot and sends only changed fields.

**Bug 4 — Stale validation warnings:** validateCliVersion() only fired on custom textbox input event. Fix: runs on load and on select change; save blocked via __ideSaveAttemptFinished__('validation_error') on invalid version.

Additional: single FIELDS constant, diff-based isDirty(), baseline resets after save.

**Tests:** 19 new JS tests in js-tests/settings-fallback.test.mjs + fixture generator.

## IDE PRs (merge after this one)

- VS Code: https://github.com/snyk/vscode-extension/pull/753
- IntelliJ: https://github.com/snyk/snyk-intellij-plugin/pull/832
- Eclipse: https://github.com/snyk/snyk-eclipse-plugin/pull/400

## Test plan

- [ ] make test-js — 93 tests pass (19 new).
- [ ] make lint — 0 errors.
- [ ] cd js-tests && npm run lint:es5 — passes.
- [ ] Manual Eclipse: tab away from cli_path without changing — NOT in changed=true payload.
- [ ] Manual all IDEs: invalid custom version blocks save; error visible on load; clears on select change.


js-tests/settings-fallback.test.mjs
```diff
@@ -0,0 +1,332 @@
...
+  const win = await buildFallbackDom();
+  const saveAttempts = [];
+  const saveFinished = [];
+  win.__saveIdeConfig__ = (json) => saveAttempts.push(json);
+  win.__ideSaveAttemptFinished__ = (status) => saveFinished.push(status);
+
+  const sel = win.document.getElementById("cli_release_channel");
+  const customInput = win.document.getElementById("cli_release_channel_custom");
+  sel.value = "custom";
+  customInput.value = "not-a-version";
```
rrama, line 194: Did we not establish that you were allowed to have random custom versions and it was weirdly supported in VS Code even though it would give the warning? I can't remember where we landed on it for the HTML Settings pages though...

shared_ide_resources/ui/html/settings-fallback.html
```diff
@@ -356,123 +356,162 @@ <h2>CLI Configuration</h2>
...
+        if (sel.value === 'custom') {
+          customInput.className = customInput.className.replace(/\bhidden\b/, '').trim();
+        } else {
+          if (customInput.className.indexOf('hidden') === -1) { customInput.className += ' hidden'; }
+        }
       }
+    }
 
-      // Auto-save if enabled
+    function markDirtyAndSave() {
```
rrama: Nitpick: Name is misleading, it only saves if auto-save is enabled.

shared_ide_resources/ui/html/settings-fallback.html
```diff
@@ -356,123 +356,162 @@ <h2>CLI Configuration</h2>
...
       } catch (e) {
         console.error('Error saving configuration:', e);
       }
     };
 
-    // Expose dirty state query function
-    window.__isFormDirty__ = function() {
-      return window.__modified__;
-    };
+    window.__isFormDirty__ = function() { return isDirty(); };
```
rrama, line 482: Q: Does it need to be a `__` function anymore then or is this to be compat with the real settings page?

shared_ide_resources/ui/html/settings-fallback.html
```diff
@@ -356,123 +356,162 @@ <h2>CLI Configuration</h2>
...
+    window.collectData = collectData;
+    window.collectChangedData = collectChangedData;
+    window.markDirtyAndSave = markDirtyAndSave;
 
-    // Reposition tooltip popups to stay within the viewport
     function repositionTooltip(trigger) {
       var popup = trigger.querySelector('.tooltip-popup');
       if (!popup) return;
 
-      // Reset to default positioning before measuring
```
rrama, line 421: I liked these comments...

rrama (approved):
Just a few nitpicks on code comments that were actually quite nice being deleted.

shared_ide_resources/ui/html/settings-fallback.html
nick-y-snyk: Renamed to markDirty — the save only happens when autosave is enabled.

shared_ide_resources/ui/html/settings-fallback.html
nick-y-snyk: Restored the tooltip positioning comments (flip below / nudge left / nudge right) — agreed they explain non-obvious geometry.

js-tests/settings-fallback.test.mjs
nick-y-snyk: The validation matches the real config dialog (same semver regex in validation.js). VS Code's leniency is in the extension layer — it sends whatever string the user types without pre-validating. The HTML page gates at the form level to prevent sending garbage to the LS.

shared_ide_resources/ui/html/settings-fallback.html
nick-y-snyk: Yes, needs __. It's part of the IDE bridge contract — IDEs call window.__isFormDirty__() to query form state, same as __saveIdeConfig__ / __onFormDirtyChange__ / __IS_IDE_AUTOSAVE_ENABLED__. collectData/collectChangedData/markDirty are exposed without __ because they're only used by the JS tests.

PR #1297, opened 2026-05-26
refactor(configuration): LdxSyncService from ctx [IDE-1974]

rrama:
### Description

Plus add the notifier to the context.
Refactor(test) `setupServer` to use functional options.
 - This includes changing a lot of tests that did their own `di.TestInit` or `di.Init` to rely on this function.
Make more tests use `enableOnlyProducts`.
Drop `Test_updateFolderConfig_UserSendsNewOrg_SetsOrgByUser`, it was a duplicate of `Test_updateFolderConfig_SwitchFromAutoToManualOrg`.

### Checklist

 - Tests just updated for the refactor.
 - N/A
 - N/A
 - N/A


application/server/configuration.go
```diff
@@ -236,7 +236,15 @@ func refreshLdxSyncOnTokenChange(conf configuration.Configuration, engine workfl
 		return
 	}
 	logger.Info().Msg("token changed via settings, refreshing LDX-Sync configuration")
-	di.LdxSyncService().RefreshConfigFromLdxSync(context.Background(), conf, engine, logger, folders, di.Notifier())
+	ldxSyncService, ok := ldxSyncServiceFromContext(ctx)
+	if !ok {
```
andrewrobinsonhodges-snyk: If we're just panicking each time the helper function returns ok=false, should we make that behaviour part of the helper function? Same applies to the notifierFromContext function below.

application/server/ldx_sync_smoke_test.go
```diff
@@ -50,16 +50,13 @@ func setupLdxSyncTest(t *testing.T) (workflow.Engine, *config.TokenServiceImpl,
...
 
-	loc, jsonRpcRecorder := setupServer(t, engine, tokenService)
+	loc, jsonRpcRecorder, _ := setupServer(t, engine, tokenService, WithRealDI())
 
 	// Disable scanning products - only testing cache behavior
 	engine.GetConfiguration().Set(configresolver.UserGlobalKey(types.SettingSnykCodeEnabled), false)
 	engine.GetConfiguration().Set(configresolver.UserGlobalKey(types.SettingSnykIacEnabled), false)
 	engine.GetConfiguration().Set(configresolver.UserGlobalKey(types.SettingSnykOssEnabled), false)
 
-	cleanupChannels()
```
andrewrobinsonhodges-snyk, line 60: Why was this removed?

application/server/server_test.go
```diff
@@ -88,48 +88,75 @@ func didOpenTextParams(t *testing.T) (sglsp.DidOpenTextDocumentParams, types.Fil
...
+	// Initialize dependencies
+	var deps di.Dependencies
+	if cfg.useRealDI {
+		deps = di.Init(engine, tokenService)
+	} else {
+		deps = di.TestInit(t, engine, tokenService, cfg.overrideDeps)
+
+		// Merge WithDeps overrides into deps struct for fields that are not handled by TestInit,
+		// i.e. InlineValueProvider.
+		if cfg.overrideDeps != nil && cfg.overrideDeps.InlineValueProvider != nil {
```
andrewrobinsonhodges-snyk, line 151: I've seen this pattern a lot. Should we have a "getDependencyIfSet" function or similar so we don't need to keep nil checking cfg.overrides?

andrewrobinsonhodges-snyk (approved):

application/server/ldx_sync_smoke_test.go
rrama: It was done as part of `setupServer`. Which now also covers the `di.Init()` below when you pass `WithRealDI()`.

application/server/server_test.go
rrama: Maybe in `test_init.go`, but I wouldn't do it here, as this is a one off place we need to do it since `InlineValueProvider` is a special snowflake.

rrama: Nevermind I looked into it for `test_init.go`. A helper would clean this up in TypeScript. But in Go we would need a typed getter per field instead of keyof, so each call site ends up with two func literals, which isworse than the explicit `overrideDeps != nil && overrideDeps.X != nil` checks.

PR #1299, opened 2026-05-27
feat(html-settings): add Eclipse-only Path field in config dialog [IDE-2037]

nick-y-snyk:
## Summary

- Adds a **Path** input to the HTML settings dialog, visible only when `INTEGRATION_ENVIRONMENT=ECLIPSE`
- Maps to the existing `user_settings_path` / `SettingUserSettingsPath` setting (already `machineScope`, already excluded from LDX sync)
- Directories entered are prepended to `PATH` before CLI subprocesses start; value is persisted via `applyUserSettingsPath` on `didChangeConfiguration` (conf-write only — no mid-session `os.Setenv` mutation, restart required)
- Non-Eclipse IDEs (VS Code, IntelliJ, Visual Studio) are unaffected: `{{if .IsEclipse}}` guard hides the field entirely

## Changes

| File | Change |
|------|--------|
| `infrastructure/configuration/config_html.go` | `isEclipse()` helper + `"IsEclipse"` data key in `GetConfigHtml` |
| `infrastructure/configuration/template/config.html` | conditional `user_settings_path` form-group in CLI section |
| `domain/ide/command/configuration_command.go` | expose `user_settings_path` in `ConstructSettingsFromConfig` |
| `application/server/configuration.go` | `applyUserSettingsPath()` + wired into `processConfigSettings` |
| Test files (×3) | rendering show/hide, settings-map key, apply round-trip |

## Test plan

- [ ] `go test ./infrastructure/configuration/... ./domain/ide/command/... ./application/server/...` passes
- [ ] `make lint-fix` → 0 issues
- [ ] Manual: build LS, start Eclipse against it → Settings → CLI section shows **Path** input
- [ ] Manual: VS Code / IntelliJ / Visual Studio → Path input must **not** appear
- [ ] Enter `/tmp/fake-bin`, save, restart → LS log shows `applyPathToEnv` prepending the value

## Related

Jira: [IDE-2037](https://snyksec.atlassian.net/browse/IDE-2037)
Eclipse plugin wiring (EC1-EC5) is a separate plugin-side change; this PR covers the snyk-ls side only.

[IDE-2037]: https://snyksec.atlassian.net/browse/IDE-2037?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/infrastructure/configuration/config_html.go
+++ b/infrastructure/configuration/config_html.go
@@ -372,7 +372,8 @@ func isVisualStudio(integrationName string) bool {
 }
 
 func isEclipse(integrationName string) bool {
-	return strings.EqualFold(integrationName, "ECLIPSE")
+	upper := strings.ToUpper(integrationName)
+	return upper == "ECLIPSE" || upper == "ECLIPSE IDE"
 }
 
 // getCliReleaseChannel derives the CLI release channel from the runtime version
--- a/infrastructure/configuration/config_html.go
+++ b/infrastructure/configuration/config_html.go
@@ -282,7 +282,6 @@ func (r *ConfigHtmlRenderer) GetConfigHtml(settings map[string]any, folderConfig
 	// as it's more user-friendly. Other than Visual Studio, which we will respect.
 	folderLabel := "Project"
 	integrationEnv := conf.GetString(configuration.INTEGRATION_ENVIRONMENT)
-	integrationName := conf.GetString(configuration.INTEGRATION_NAME)
 	if isVisualStudio(integrationEnv) {
 		folderLabel = "Solution"
 	}
@@ -345,7 +344,7 @@ func (r *ConfigHtmlRenderer) GetConfigHtml(settings map[string]any, folderConfig
 		"FolderNames":             folderNames,
 		"CliReleaseChannel":       cliReleaseChannel,
 		"IsSecretsFeatureEnabled": isAnyFolderSecretsEnabled(folderConfigs),
-		"IsEclipse":               isEclipse(integrationName),
+		"IsEclipse":               isEclipse(integrationEnv),
 	}
 
 	var buffer bytes.Buffer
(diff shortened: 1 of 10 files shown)


infrastructure/configuration/config_html.go
```diff
@@ -369,6 +371,11 @@ func isVisualStudio(integrationName string) bool {
 	return integrationName == "VISUAL_STUDIO" || integrationName == "Visual Studio"
 }
 
+func isEclipse(integrationName string) bool {
+	upper := strings.ToUpper(integrationName)
+	return upper == "ECLIPSE" || upper == "ECLIPSE IDE" || upper == "ECLIPSE PLATFORM"
```
andrewrobinsonhodges-snyk: Where did this list come from? It might well be correct, but I'm always wary of magic strings.

andrewrobinsonhodges-snyk: Should we detect that the integration name contains "ECLIPSE" instead?

andrewrobinsonhodges-snyk (approved):

infrastructure/configuration/config_html.go
nick-y-snyk: Good catch. The list came from lsp4e internals (`Platform.getProduct().getName()`) which is runtime-dependent and not under our control. Fixed in the latest commit — both `isEclipse` and `isVisualStudio` now read `INTEGRATION_NAME` instead of `INTEGRATION_ENVIRONMENT`. Each plugin hardcodes this explicitly: Eclipse → `"ECLIPSE"` (`Activator.java`), VS → `"VISUAL_STUDIO"` (`SnykExtension.cs`), IntelliJ → `"JETBRAINS_IDE"`, VS Code → `"VS_CODE"`. No more magic strings from platform internals.

PR #1300, opened 2026-05-28
fix: LSP init perf — cache flush on reauth, TOCTOU guard, no-persist [IDE-1898]

bastiandoetsch:
## Summary

- **flushGen TOCTOU guard** in `fetch()`: captures `gen := s.flushGen` before releasing the mutex on cache-miss; skips `orgToFlag.Set()` if `FlushCache()` ran while flag goroutines were in-flight (mirrors the existing `fetchSastSettings` guard).
- **Flush before populate**: `FlushCache()` called before `PopulateFolderConfig` in `postCredentialUpdateHook` (`login.go`) and before `RefreshConfigFromLdxSync` on token change (`configuration.go`), so stale cached 401s are cleared before the first re-fetch after login.
- **Remove `PersistInStorage` from `SetFeatureFlag`** (`folder_config.go`): caused O(N²) full-JSON re-serialization on every folder's flag write at init time with N folders.
- **Negative error cache for SAST settings**: `fetchSastSettings` caches provider errors for 60s so N folders sharing a 401 org make one HTTP call instead of N. Gen-guard prevents the error from re-poisoning after a flush.
- **Channel-based LSP init wait**: replaces the 1ms busy-poll spin loop in the `registerNotifier` callback with `WaitForLspInitialized`, eliminating the deadlock that blocked `HandleFolders` channel writes when N goroutines spun instead of draining.

## Test plan

- [x] New TOCTOU subtest for `fetch()` uses channel sync (not `time.Sleep`) for deterministic race-window coverage; race detector clean
- [x] Existing SAST TOCTOU subtest updated to use atomic nil-close pattern; race detector clean
- [x] Login ordering asserted as `["flush", "populate"]` (not just `callOrder[0] == "flush"`)
- [x] `Test_LSPInitCompletesWithManyFolders` (N=200 fake HTTP, ≤30s) and `Test_LSPInitCompletesWithManyFoldersRealHTTP` (real HTTP, ≤40s, skipped under `-short`) added to `lsp_init_perf_test.go`

___

### **PR Type**
Bug fix, Enhancement

___

### **Description**
- Optimize LSP initialization performance and reliability.

- Enhance cache handling with TOCTOU guards.

- Ensure feature flags and SAST settings are flushed on re-authentication.

- Refactor dependency injection to use context.

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A[RPC Request] --> B(Server Handler);
  B --> C(withContext Wrapper);
  C --> D(Inject Dependencies into Context);
  D --> E(Handler Logic);
  E --> F(Get Service from Context);
  F --> G(Perform Action);
```

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><details><summary>5 files</summary><table>
<tr>
  <td><strong>init.go</strong><dd><code>Add comprehensive dependencies to DI struct</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1300/changes#diff-3dfa2a6c1d0ed218f91d033ea2c5e5575f64d80c2575b71fe0605ec4fa41d4f6">+30/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>configuration.go</strong><dd><code>Flush feature flag cache on token change</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1300/changes#diff-eb3fe6fd1956591838595b08979653ec23c0b0c16f5340e9e89a05c30e921607">+19/-6</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>login.go</strong><dd><code>Flush cache before populating on login</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1300/changes#diff-21b93267e7967e84d95e104e3b17f130b4b47b4a37ac176c5bba7dd4c35cee31">+2/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>featureflag.go</strong><dd><code>Add TOCTOU guard and negative error cache for SAST</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1300/changes#diff-46f9f8e9fa897e638a86943984fb86d3f24b82e6d55cebc21b1bb7828a92e3dc">+55/-25</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>config_readers.go</strong><dd><code>Implement channel-based LSP initialization wait</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1300/changes#diff-3b315cd032c0bc1cd1052fd4bb3cc0c762129e8d5cd03c1ef15f42cd926c4588">+38/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Tests</strong></td><td><details><summary>9 files</summary><table>
<tr>
  <td><strong>init_test.go</strong><dd><code>Test all dependency fields are populated</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1300/changes#diff-fcd7ecadf7c2cc1571bea1e6d6238bb60ffce015cbe3f70cd9b3978a3facd621">+75/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>lsp_init_perf_test.go</strong><dd><code>Add LSP init performance tests</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1300/changes#diff-4a9aa6bc23842643d5e3f21685c2fbb78b8df5a26f36952e14767a6db36e3ae1">+188/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>notification_test.go</strong><dd><code>Test LSP init channel wait and notifier delivery</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1300/changes#diff-09b80ca28f0c9f8db60c941c24e92dfc65ded48c06249e70e2004faec30b9891">+35/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>server_test.go</strong><dd><code>Test context dependency injection and LSP init waits</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1300/changes#diff-165bad981a0cc202462a2fea7f236eab9ca3a72bd47fe8bae6b3adad801351fb">+49/-3</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>trust_test.go</strong><dd><code>Update trust tests with injected dependencies</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1300/changes#diff-e429a057cdac11e281a6ee8603bf8a5ab6dfaf0daedaec873982c47b28d7d431">+4/-4</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>login_test.go</strong><dd><code>Test cache flush order on login</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1300/changes#diff-4862594eb7581e0be85b55a20010df26b21d42fa6a82337c79af7f0342632ee0">+59/-4</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>featureflag_test.go</strong><dd><code>Test feature flag cache TOCTOU guard and negative caching</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1300/changes#diff-cd6009395d9cf45b7fc17fc2660414b2a02058efdcb6640207015a095f20b08c">+199/-6</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>context_test.go</strong><dd><code>Test new context dependency keys and round-trip</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1300/changes#diff-d66ddc393ac513ecf609ea8b085a69f378872c89054197f1e6041aac0fd9e0db">+59/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>channels.go</strong><dd><code>Add utility for testing channel closes</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1300/changes#diff-46d04268ab60ff7ca13961938b8b33cd6efac87179b0d1ae1da2ffe7312f6c40">+15/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Refactoring</strong></td><td><details><summary>6 files</summary><table>
<tr>
  <td><strong>codeaction_handlers.go</strong><dd><code>Use injected CodeActionService instead of DI global</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1300/changes#diff-69c83c3e9f7f13f9534ab34f5f36cf03307e6db0e05732846bed509729293cc1">+2/-3</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>execute_command.go</strong><dd><code>Use error reporter from context</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1300/changes#diff-523ba95647cc7cf643285dc5a700fafb1c295bfbdc06065624232a888d1df3ef">+2/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>notification.go</strong><dd><code>Use channel for LSP init wait and injected Notifier</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1300/changes#diff-3ac97f07d29586cd80b9c63b5ac84fa3c0c029c0786ad485ae98d5d143cdaa53">+5/-7</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>server.go</strong><dd><code>Refactor handlers to use context-injected dependencies</code>&nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1300/changes#diff-1b5a200b7dc3f37ed67632d0c2682eeb53762813f96edb15cb7660e076fa6031">+313/-103</a></td>

</tr>

<tr>
  <td><strong>context.go</strong><dd><code>Add new dependency keys for context injection</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1300/changes#diff-5ba4125877fb56856b4bb75f14ff968c77cdee83e09ea4cefd9d24fc2d453923">+7/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>ldx_sync_config.go</strong><dd><code>Add LSP initialized channel setting key</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1300/changes#diff-e42a6f2c9aece134bd052ddbd7e835aefd3661d3d5f9446f7d5b621c421eea4d">+1/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
<tr>
  <td><strong>folder_config.go</strong><dd><code>Remove PersistInStorage from SetFeatureFlag</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1300/changes#diff-74d6e30f4ca44c7cdb655e9b89d7166a3f22765b35a6d94e71574c8620958a65">+3/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

</details>

___

### DIFF
--- a/application/di/init.go
+++ b/application/di/init.go
@@ -297,6 +297,15 @@ func FeatureFlagService() featureflag.Service {
 	return featureFlagService
 }
 
+// SetFeatureFlagService replaces the global featureFlagService for future calls
+// to FeatureFlagService(). Objects already constructed before this call retain
+// their previous reference. Intended for test use only.
+func SetFeatureFlagService(service featureflag.Service) {
+	initMutex.Lock()
+	defer initMutex.Unlock()
+	featureFlagService = service
+}
+
 func LdxSyncService() command.LdxSyncService {
 	initMutex.Lock()
 	defer initMutex.Unlock()
--- a/application/di/init.go
+++ b/application/di/init.go
@@ -101,6 +101,18 @@ type Dependencies struct {
 	LdxSyncService        command.LdxSyncService
 	ScanStateAggregator   scanstates.Aggregator
 	InlineValueProvider   snyk.InlineValueProvider
+	// Added fields — all handler-accessed singletons now live here so that
+	// withContext can inject them into the request context rather than handlers
+	// reaching for the package-level globals via di.*() accessor functions.
+	FileWatcher       *watcher.FileWatcher
+	ErrorReporter     er.ErrorReporter
+	HoverService      hover.Service
+	Scanner           scanner2.Scanner
+	ScanPersister     persistence.ScanSnapshotPersister
+	ScanNotifier      scanner2.ScanNotifier
+	Installer         install.Installer
+	CodeActionService *codeaction.CodeActionsService
+	Initializer       initialize.Initializer
 }
 
 func currentDependencies() Dependencies {
@@ -117,6 +129,15 @@ func currentDependencies() Dependencies {
 		LdxSyncService:        ldxSyncService,
 		ScanStateAggregator:   scanStateAggregator,
 		InlineValueProvider:   inlineValueProvider,
+		FileWatcher:           fileWatcher,
+		ErrorReporter:         errorReporter,
+		HoverService:          hoverService,
+		Scanner:               scanner,
+		ScanPersister:         scanPersister,
+		ScanNotifier:          scanNotifier,
+		Installer:             installer,
+		CodeActionService:     codeActionService,
+		Initializer:           scanInitializer,
 	}
 }
 
--- a/application/server/configuration.go
+++ b/application/server/configuration.go
@@ -228,12 +228,20 @@ func UpdateSettings(ctx context.Context, conf configuration.Configuration, engin
 
 func refreshLdxSyncOnTokenChange(ctx context.Context, conf configuration.Configuration, engine workflow.Engine, logger *zerolog.Logger, ws types.Workspace, oldToken string) {
 	newToken := config.GetToken(conf)
-	if newToken == oldToken || newToken == "" || ws == nil {
+	if newToken == oldToken || newToken == "" {
+		return
+	}
+	if ws == nil {
 		return
 	}
 	folders := ws.Folders()
 	if len(folders) == 0 {
 		return
+	}
+	// Flush stale cached 401 errors so fresh calls use the new token.
+	// Placed here (after ws/folders guards) so flush only fires when populate follows.
+	if svc := di.FeatureFlagService(); svc != nil {
+		svc.FlushCache()
 	}
 	logger.Info().Msg("token changed via settings, refreshing LDX-Sync configuration")
 	ldxSyncService := mustLdxSyncServiceFromContext(ctx)
--- a/application/server/configuration.go
+++ b/application/server/configuration.go
@@ -216,6 +216,16 @@ func UpdateSettings(ctx context.Context, conf configuration.Configuration, engin
 	}
 
 	globalOrgChanged, lockedMachineFields := processConfigSettings(ctx, conf, engine, logger, settings, triggerSource, configResolver)
+
+	// Flush stale cached errors (e.g. 401s from a previous token) before
+	// PopulateFolderConfig runs inside processFolderConfigs. Flushing here
+	// ensures every folder sees fresh results with the new token.
+	if newToken := config.GetToken(conf); newToken != "" && newToken != oldToken {
+		if svc := di.FeatureFlagService(); svc != nil {
+			svc.FlushCache()
+		}
+	}
+
 	lockedFolderFields := processFolderConfigs(ctx, conf, engine, logger, folderConfigs, triggerSource, configResolver, globalOrgChanged)
 
 	var fm workflow.ConfigurationOptionsMetaData
@@ -249,11 +259,6 @@ func refreshLdxSyncOnTokenChange(ctx context.Context, conf configuration.Configu
 	folders := ws.Folders()
 	if len(folders) == 0 {
 		return
-	}
-	// Flush stale cached 401 errors so fresh calls use the new token.
-	// Placed here (after ws/folders guards) so flush only fires when populate follows.
-	if svc := di.FeatureFlagService(); svc != nil {
-		svc.FlushCache()
 	}
 	logger.Info().Msg("token changed via settings, refreshing LDX-Sync configuration")
 	ldxSyncService := mustLdxSyncServiceFromContext(ctx)
@@ -395,7 +400,8 @@ func processConfigSettings(ctx context.Context, conf configuration.Configuration
 	applySnykLearnCodeActions(conf, engine, logger, settings, triggerSource, configResolver)
 	applySnykOssQuickFixCodeActions(conf, engine, logger, settings, triggerSource, configResolver)
 	applySnykOpenBrowserActions(conf, settings)
-	applyMcpConfiguration(conf, engine, logger, settings, triggerSource, configResolver)
+	n, _ := notifierFromContext(ctx)
+	applyMcpConfiguration(n, conf, engine, logger, settings, triggerSource, configResolver)
 	applyPublishSecurityAtInceptionRules(conf, settings)
 	// this is without function right now, we do not use/distribute proxy settings from/to IDEs
 	applyProxyConfig(conf, settings)
@@ -1018,8 +1024,7 @@ func applySnykOpenBrowserActions(conf configuration.Configuration, settings map[
 	}
 }
 
-func applyMcpConfiguration(conf configuration.Configuration, engine workflow.Engine, logger *zerolog.Logger, settings map[string]*types.ConfigSetting, triggerSource analytics.TriggerSource, configResolver types.ConfigResolverInterface) {
-	n := di.Notifier()
+func applyMcpConfiguration(n notification.Notifier, conf configuration.Configuration, engine workflow.Engine, logger *zerolog.Logger, settings map[string]*types.ConfigSetting, triggerSource analytics.TriggerSource, configResolver types.ConfigResolverInterface) {
 	if v, ok := settingBool(settings, types.SettingAutoConfigureMcpServer); ok {
 		oldValue := types.GetGlobalBool(conf, types.SettingAutoConfigureMcpServer)
 		types.SetGlobalUser(conf, types.SettingAutoConfigureMcpServer, v)
@@ -1027,7 +1032,7 @@ func applyMcpConfiguration(conf configuration.Configuration, engine workflow.Eng
 			if conf.GetBool(types.SettingIsLspInitialized) {
 				go analytics.SendConfigChangedAnalytics(conf, engine, logger, configAutoConfigureSnykMcpServer, oldValue, v, triggerSource, configResolver)
 			}
-			mcpWorkflow.CallMcpConfigWorkflow(conf, di.ConfigResolver(), engine, logger, n, true, false)
+			mcpWorkflow.CallMcpConfigWorkflow(conf, configResolver, engine, logger, n, true, false)
 		}
 	}
 
@@ -1038,7 +1043,7 @@ func applyMcpConfiguration(conf configuration.Configuration, engine workflow.Eng
 			if conf.GetBool(types.SettingIsLspInitialized) {
 				go analytics.SendConfigChangedAnalytics(conf, engine, logger, configSecureAtInceptionExecutionFrequency, oldValue, v, triggerSource, configResolver)
 			}
-			mcpWorkflow.CallMcpConfigWorkflow(conf, di.ConfigResolver(), engine, logger, n, false, true)
+			mcpWorkflow.CallMcpConfigWorkflow(conf, configResolver, engine, logger, n, false, true)
 		}
 	}
 }
--- a/application/server/lsp_init_perf_test.go
+++ b/application/server/lsp_init_perf_test.go
@@ -16,23 +16,21 @@
 
 package server
 
-// IDE-1898: LSP initialization performance at protocol boundary.
+// LSP initialization performance at protocol boundary.
 //
 // This file contains:
-//   - BenchmarkLSPInitializedWithNFolders — measures wall time of the full
-//     initialize+initialized sequence for N workspace folders; run with
-//     -cpuprofile to locate the actual hotspot before making assertions.
-//   - Test_IDE1898_LSPInitCompletesWithManyFolders — Level-3 requirement test:
-//     the `initialized` handler must complete within a profiling-derived bound
-//     for N=200 folders.
+//   - TestProfileLSPInit — captures a CPU profile of one initialize+initialized
+//     cycle; skipped in -short mode, run explicitly to locate hotspots.
+//   - Test_LSPInitCompletesWithManyFolders — Level-3 requirement test:
+//     the `initialized` handler must complete within 30s for N=200 folders
+//     (fake featureFlagService, CI-safe).
+//   - Test_LSPInitCompletesWithManyFoldersRealHTTP — real-HTTP variant with
+//     a 40s limit; skipped in -short mode.
 //
 // Usage:
 //
-//	go test ./application/server/... \
-//	  -run=^$ -bench=BenchmarkLSPInitializedWithNFolders \
-//	  -benchtime=3x -cpuprofile=/tmp/lsp-init-cpu.prof
-//
-//	go tool pprof -top -nodecount=20 /tmp/lsp-init-cpu.prof
+//	go test ./application/server/... -run TestProfileLSPInit -v -count=1
+//	go tool pprof -top -nodecount=30 /tmp/lsp-init-cpu-*.prof
 
 import (
 	"os"
@@ -91,7 +89,7 @@ func TestProfileLSPInit(t *testing.T) {
 	engine, tokenService := testutil.UnitTestWithEngine(t)
 	params := buildInitParams(t, lspInitPerfFolderCount)
 
-	f, err := os.CreateTemp("", "lsp-init-cpu-*.prof")
+	f, err := os.CreateTemp(t.TempDir(), "lsp-init-cpu-*.prof")
 	require.NoError(t, err)
 	t.Logf("CPU profile: %s", f.Name())
 
@@ -124,7 +122,7 @@ func TestProfileLSPInit(t *testing.T) {
 // Design: uses a fake featureFlagService (no HTTP calls) so the test is reliable in CI
 // regardless of network access or token validity.  The test validates the initialization
 // code path — folder processing, JSON config writes, notifier plumbing — for N=200 folders.
-// Real HTTP behaviour is exercised by Test_LSPInitCompletesWithManyFoldersRealHTTP.
+// Real HTTP behavior is exercised by Test_LSPInitCompletesWithManyFoldersRealHTTP.
 //
 // The 30s bound is loose enough to catch O(N²) regressions in folder processing and
 // tight enough to detect hangs in the notification/channel machinery.
--- a/application/server/lsp_init_perf_test.go
+++ b/application/server/lsp_init_perf_test.go
@@ -100,7 +100,7 @@ func TestProfileLSPInit(t *testing.T) {
 		t.Logf("profile written; inspect with: go tool pprof -top -nodecount=30 %s", f.Name())
 	}()
 
-	loc, _ := setupServer(t, engine, tokenService)
+	loc, _, _ := setupServer(t, engine, tokenService)
 
 	_, err = loc.Client.Call(t.Context(), "initialize", params)
 	require.NoError(t, err)
@@ -132,7 +132,7 @@ func Test_LSPInitCompletesWithManyFolders(t *testing.T) {
 	engine, tokenService := testutil.UnitTestWithEngine(t)
 	params := buildInitParams(t, lspInitPerfFolderCount)
 
-	loc, _ := setupServer(t, engine, tokenService)
+	loc, _, _ := setupServer(t, engine, tokenService)
 
 	_, err := loc.Client.Call(t.Context(), "initialize", params)
 	require.NoError(t, err)
@@ -180,7 +180,7 @@ func Test_LSPInitCompletesWithManyFoldersRealHTTP(t *testing.T) {
 	engine, tokenService := testutil.UnitTestWithEngine(t)
 	params := buildInitParams(t, lspInitPerfFolderCount)
 
-	loc, _ := setupServer(t, engine, tokenService)
+	loc, _, _ := setupServer(t, engine, tokenService)
 
 	_, err := loc.Client.Call(t.Context(), "initialize", params)
 	require.NoError(t, err)
--- a/application/server/lsp_init_perf_test.go
+++ b/application/server/lsp_init_perf_test.go
@@ -1,0 +1,202 @@
+/*
+ * © 2026 Snyk Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package server
+
+// IDE-1898: LSP initialization performance at protocol boundary.
+//
+// This file contains:
+//   - BenchmarkLSPInitializedWithNFolders — measures wall time of the full
+//     initialize+initialized sequence for N workspace folders; run with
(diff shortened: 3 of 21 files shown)


bastiandoetsch:
> ## PR Reviewer Guide 🔍
> 🧪 **PR contains tests**
> 🔒 **No security concerns identified**
> ⚡ **Recommended focus areas for review**
> 
> [**Behavioral Change**](https://github.com/snyk/snyk-ls/pull/1300/changes#diff-74d6e30f4ca44c7cdb655e9b89d7166a3f22765b35a6d94e71574c8620958a65R152-R152) 🟡 [minor]
> The removal of `conf.PersistInStorage(key)` means that folder-specific feature flags will no longer survive an IDE restart. While this fixes the O(N^2) performance issue during initialization, it increases reliance on the API being reachable immediately after restart to determine folder capabilities.
> 
> ```go
> conf.Set(key, value)
> ```

that's expected and wanted.

application/di/init.go
```diff
@@ -297,6 +297,15 @@ func FeatureFlagService() featureflag.Service {
...
 }
 
+// SetFeatureFlagService replaces the global featureFlagService for future calls
+// to FeatureFlagService(). Objects already constructed before this call retain
+// their previous reference. Intended for test use only.
+func SetFeatureFlagService(service featureflag.Service) {
+	initMutex.Lock()
+	defer initMutex.Unlock()
+	featureFlagService = service
+}
```
bastiandoetsch, line 328: this pattern is gonna be resolved in a follow-up pr that's created right now.

application/server/lsp_init_perf_test.go
```diff
@@ -0,0 +1,200 @@
...
+
+// profileLSPInitialized captures a CPU profile of one initialize+initialized
+// cycle into /tmp/lsp-init-cpu.prof.  Call via:
+//
+//	go test ./application/server/... -run TestProfileLSPInit -v -count=1
+//
+// Then inspect with: go tool pprof -top -nodecount=30 /tmp/lsp-init-cpu.prof
+func TestProfileLSPInit(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping profiling test in short mode; run explicitly: go test -run TestProfileLSPInit -v")
```
rrama, line 87: Q: Is this test handled very differently to how you did it before for `make benchmark` and `make benchmark-real` should you align it to be more like them or is it deliberately differently done?

application/server/lsp_init_perf_test.go
```diff
@@ -0,0 +1,200 @@
...
+	// Disable auto-scan after initialize: measure initialization overhead, not scan time.
+	disableAutoScan(t, engine.GetConfiguration())
+
+	start := time.Now()
+	_, err = loc.Client.Call(t.Context(), "initialized", types.InitializedParams{})
+	require.NoError(t, err)
+	t.Logf("initialized with %d folders took %v", lspInitPerfFolderCount, time.Since(start))
+}
+
+// Test_LSPInitCompletesWithManyFolders is the Level-3 requirement test.
```
rrama: Nitpick: AI left in comment.

application/server/configuration.go
```diff
@@ -240,13 +240,21 @@ func UpdateSettings(ctx context.Context, conf configuration.Configuration, engin
...
 	}
 	folders := ws.Folders()
 	if len(folders) == 0 {
 		return
 	}
+	// Flush stale cached 401 errors so fresh calls use the new token.
+	// Placed here (after ws/folders guards) so flush only fires when populate follows.
+	if svc := di.FeatureFlagService(); svc != nil {
+		svc.FlushCache()
+	}
```
rrama: I asked my AI if this is the best place for this call, it said:
> By the time refreshLdxSyncOnTokenChange runs at configuration.go:238, processFolderConfigs has already executed at line 219, which calls di.FeatureFlagService().PopulateFolderConfig(fc) at configuration.go:1155 for every folder — all against the still-stale cache containing the pre-token 401 entries. The flush here only benefits the subsequent RefreshConfigFromLdxSync call, not the N folder populates that just happened.
>
> To match the intent (and the ordering enforced by TestLoginCommand_Execute_FlushCacheBeforePopulate for the login path), the flush should happen earlier in UpdateSettings — after applyToken inside processConfigSettings but before processFolderConfigs at line 219. That way both code paths (login command and settings-driven token change) flush before any PopulateFolderConfig runs.
>
> Side note: moving it earlier is also a good opportunity to thread the FF service through as a parameter rather than calling di.FeatureFlagService() here — the .cursor/rules/server-dependencies.mdc rule flags new di.* lookups in production code (also called out by Lifeguard).

Do what you will with this information.

infrastructure/featureflag/featureflag_test.go
```diff
@@ -254,6 +281,33 @@ func TestFetch(t *testing.T) {
...
+		org := "toctou-flags-org"
+
+		done := make(chan struct{})
+		go func() {
+			defer close(done)
+			_ = service.fetch(org)
+		}()
+
+		// Wait for the first flag goroutine to signal it's mid-fetch, then flush.
+		<-mockProvider.flagReadyCh
```
rrama: Nitpick: Safer to use the util I wrote `testsupport.RequireEventuallyReceive(t, mockProvider.flagReadyCh, 5*time.Second, time.Millisecond, "flag fetch did not start in time")`, so that if it doesn't trigger we don't have to wait on the Go overall test timeout.

infrastructure/featureflag/featureflag_test.go
```diff
@@ -254,6 +281,33 @@ func TestFetch(t *testing.T) {
...
+		service.FlushCache()
+
+		<-done // wait for in-flight goroutines to finish
```
rrama: Nitpick: I would say, to do the same as the above of using testsupport.RequireEventuallyReceive, but I think it doesn't like channel closes, so you'd need a new func like it called testsupport.RequireEventuallyClosed.

infrastructure/featureflag/featureflag_test.go
```diff
@@ -740,3 +796,147 @@ func TestServiceImpl_Override_PinsFlag(t *testing.T) {
...
+
+		// Start a fetch that will sleep 50ms inside getSastSettings.
+		done := make(chan struct{})
+		go func() {
+			defer close(done)
+			_, _ = service.fetchSastSettings(org)
+		}()
+
+		// Wait for getSastSettings to signal it is mid-sleep, then flush.
+		<-sastReady // use local var — mock nils the field under lock before closing
```
rrama: Nitpick: Similar to before, please use a new func called testsupport.RequireEventuallyClosed or change to use testsupport.RequireEventuallyReceive here and for <-done below.

infrastructure/featureflag/featureflag_test.go
```diff
@@ -87,7 +105,16 @@ func (m *mockExternalCallsProvider) getFeatureFlag(flag string, org string) (boo
 func (m *mockExternalCallsProvider) getSastSettings(org string) (*sast_contract.SastResponse, error) {
 	m.mu.Lock()
 	m.sastSettingsCalls++
+	delay := m.sastDelay
+	readyCh := m.sastReadyCh
+	m.sastReadyCh = nil // nil atomically with the capture to prevent double-close
```
rrama: Very very minor nitpick: Or just signal on the channel like above (I spent a bit of time wondering why the difference tbh). Or `sync.Once` the close.

infrastructure/featureflag/featureflag_test.go
```diff
@@ -40,8 +43,12 @@ type mockExternalCallsProvider struct {
 	ignoreErr           error
 	featureFlagsByOrg   map[string]map[string]bool
 	flagErr             error
+	flagDelay           time.Duration // artificial latency injected into getFeatureFlag
```
rrama, line 47: Nitpick: Instead of relying on sleeps (I know they are only short), how about:
flagReadyCh   chan struct{} // mock -> test: signals "I'm in the call, flush the cache."
flagProceedCh chan struct{} // test -> mock: signals "I've flushed the cache, carry on."
Then in getFeatureFlag:
if proceedCh != nil {
    testsupport.RequireEventuallyReceive(t, proceedCh, 50*time.Millisecond, time.Millisecond, "test did not signal to resume in time")
}
and in the test:
service.FlushCache()
mockProvider.flagProceedCh <- struct{}{}

infrastructure/featureflag/featureflag_test.go
```diff
@@ -740,3 +796,147 @@ func TestServiceImpl_Override_PinsFlag(t *testing.T) {
...
+		}
+
+		mockProvider.mu.Lock()
+		calls := mockProvider.sastSettingsCalls
+		mockProvider.mu.Unlock()
+
+		assert.Equal(t, 1, calls, "provider should only be called once; subsequent calls should use the negative cache")
+	})
+
+	t.Run("negative cache does not affect success path", func(t *testing.T) {
```
rrama: This test is just the golden path... The name suggests it doesn something with the negative cache, but it doesn't.

infrastructure/featureflag/featureflag_test.go
```diff
@@ -740,3 +796,147 @@ func TestServiceImpl_Override_PinsFlag(t *testing.T) {
...
+			_, _ = service.fetchSastSettings(org)
+		}()
+
+		// Wait for getSastSettings to signal it is mid-sleep, then flush.
+		<-sastReady // use local var — mock nils the field under lock before closing
+		service.FlushCache()
+
+		<-done // wait for the in-flight goroutine to finish
+
+		// A second call must hit the provider again, not the re-poisoned cache.
```
rrama: Q: Is there not a better way to assert the cache like we did in the other test when it did:
```go
		// The gen guard must have prevented stale flags from being written back.
		_, found := service.orgToFlag.Get(org)
		assert.False(t, found, "flag cache must not be re-populated after concurrent FlushCache (TOCTOU guard)")
```
Or is counting the only way?

infrastructure/featureflag/featureflag.go
```diff
@@ -219,18 +231,19 @@ func (s *serviceImpl) fetch(org string) map[string]bool {
...
 		}()
 	}
 
 	wg.Wait()
 
 	s.mutex.Lock()
-	s.orgToFlag.Set(org, orgFlags, imcache.WithExpiration(time.Minute))
+	if s.flushGen == gen {
+		// 30s TTL: flags may be deactivated quickly (e.g. org plan changes), so keep
+		// the cache short — well below the 1-minute SAST settings TTL.
```
rrama, line 244: Q: But why does it need to be shorter?

internal/types/folder_config.go
```diff
@@ -137,6 +137,9 @@ func (fc *FolderConfig) GetFeatureFlag(flag string) bool {
 }
 
 // SetFeatureFlag writes a feature flag value to configuration under the folder metadata prefix.
+// Feature flags are re-fetched from the API on every initialized call; they must not be persisted
+// to disk because PersistInStorage triggers a full JSON re-serialization on every Set, which is
+// O(N²) across N folders × M flags per folder.
```
rrama, line 142: Q: Does this really matter that much? Does it not only write if the flag changes value?

rrama (approved):
Left some questions. which are more for my understanding than anything else.

infrastructure/featureflag/featureflag.go
bastiandoetsch: because I found it better to have a higher chance of a settings change to take effect without stale caches ;). Just a requirement I put.

infrastructure/featureflag/featureflag_test.go
bastiandoetsch: hm. in this case this is supposed to simulate a sloooow service. That's a different scenario than coordinating signals over channels. I think your solution would make the mockCallsProvider expose a different behavior than the one needed for the performance testing.

infrastructure/featureflag/featureflag_test.go
bastiandoetsch: yep, it confirms that it works even though there is a negative cache. potentially it could be removed, but in the end it's a regression test. I wonder if we already have it, though.

application/server/lsp_init_perf_test.go
bastiandoetsch: It's deliberately different — `make benchmark` and `make benchmark-real` are `go test -bench` microbenchmarks measuring raw operation throughput, while this test measures wall-clock LSP initialization latency end-to-end. The different structure reflects the different thing being measured.

application/server/configuration.go
bastiandoetsch: Good catch — the AI is correct that the flush at `refreshLdxSyncOnTokenChange` is too late; `processFolderConfigs` has already run N `PopulateFolderConfig` calls against the stale cache. Will move the flush to before `processFolderConfigs` in `UpdateSettings`. The `di.FeatureFlagService()` threading is tracked separately under the server-dependencies rules.

infrastructure/featureflag/featureflag_test.go
bastiandoetsch: Will rename the test to better reflect what it actually asserts — it's a regression guard confirming the golden path still works in the presence of a negative cache, not a dedicated negative-cache scenario test.

application/server/lsp_init_perf_test.go
bastiandoetsch: Fixed — removing the stray AI comment.

infrastructure/featureflag/featureflag_test.go
bastiandoetsch: Will make it consistent — switching to a channel signal to match the pattern used above.

infrastructure/featureflag/featureflag_test.go
bastiandoetsch: Good point — will switch to `testsupport.RequireEventuallyReceive` for a faster, clearer failure message.

infrastructure/featureflag/featureflag_test.go
bastiandoetsch: Will add `testsupport.RequireEventuallyClosed` and use it here.

infrastructure/featureflag/featureflag_test.go
bastiandoetsch: Will use `testsupport.RequireEventuallyClosed` here (and for the `<-done` below) once the helper is added.

internal/types/folder_config.go
bastiandoetsch: Yes, it matters for init perf: `PersistInStorage` writes to disk on every call regardless of whether the value changed. With N folders each fetching M flags during startup, that's N×M disk writes. Removing it trades away persistence across restarts (flags are re-fetched from the API) for a much cheaper init path.

infrastructure/featureflag/featureflag_test.go
bastiandoetsch: Good point — will switch to direct cache inspection (`_, found := service.orgToFlag.Get(org); assert.False(t, found, ...)`) to match the TOCTOU test pattern, which is a cleaner assertion than counting.

application/di/init.go
bastiandoetsch: Follow-up PR created.

bastiandoetsch:
Re: https://github.com/snyk/snyk-ls/pull/1300#pullrequestreview-4381002292

Thanks for the review! All questions answered in the thread — and several of the nits will be addressed in a follow-up commit.

PR #1304, opened 2026-05-28
feat(ignore): refresh tree view immediately after ignore action [IDE-2000]

nick-y-snyk:
## Summary
image

- After ignore/unignore completes, `submitIgnoreRequest` now calls `TreeScanStateEmitter.Emit()` to push a fresh tree view to the IDE — no rescan triggered
- Injected via a minimal `treeRefresher` interface (one method); `*TreeScanStateEmitter` satisfies it with zero new code in the treeview package
- Expanded node state preserved automatically via the server-side `GlobalExpandState` singleton already baked into the emitter
- Tree refresh skipped on failure — `Execute` returns early before the emit call, matching AC

Jira: https://snyksec.atlassian.net/browse/IDE-2000

## Test plan

- [ ] New unit tests: `Test_submitIgnoreRequest_TriggersTreeRefreshOnSuccess` (create/update/delete), `Test_submitIgnoreRequest_NoTreeRefreshOnFailure`, `Test_submitIgnoreRequest_NilTreeRefresher_DoesNotPanic`
- [ ] All existing `domain/ide/command` tests pass unchanged (nil-guard on new fields preserves prior behaviour)
- [ ] Manual: ignore an issue in VS Code / IntelliJ → IGNORED badge appears in tree immediately, no rescan progress bar, expanded folders stay expanded
- [ ] Manual: unignore → badge disappears immediately
- [ ] Manual: force ignore to fail → tree unchanged, error surfaces to user

### DIFF
--- a/.cursor/agents/planner.md
+++ b/.cursor/agents/planner.md
@@ -8,7 +8,7 @@ You are a planning specialist responsible for creating thorough, structured impl
 ## When Invoked
 
 1. **Read the `create-implementation-plan` skill** at `.cursor/skills/create-implementation-plan/SKILL.md` and follow it exactly.
-2. **Read the implementation plan template** at `.github/IMPLEMENTATION_PLAN_TEMPLATE.md` and use it as the base.
+2. **Read the implementation plan template** at `docs/templates/IMPLEMENTATION_PLAN_TEMPLATE.md` and use it as the base.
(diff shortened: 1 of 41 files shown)


rrama (approved):
LGTM!
You could put the dependency on the context like we are starting to do elsewhere, but this is fine tbh.

PR #1305, opened 2026-05-28
fix(treeview): position scan-error overlay using measured height [IDE-1808]

acke:
## Summary

The scan-error overlay (shown when clicking a `.tree-node-error` row in the LS-generated tree webview) was clipped at the bottom of the panel for product-error rows near the bottom of the sidebar. All four IDE clients (VS Code, IntelliJ, Eclipse, Visual Studio) inherit this webview, so the bug was IDE-agnostic.

Three issues in the previous positioning code:
1. **Hardcoded `200`** as the overlay height estimate — real overlays with non-trivial error messages exceed 300px, so the flip-above branch was effectively dead.
2. **Wrong flip-above math** (`top = rect.top - 4`) — placed the overlay's *top* 4px above the row's top, leaving the body overlapping the row downward instead of sitting above it.
3. **Position computed before the overlay was in the DOM**, so the real rendered height could never be measured.

## What changed

- Append the overlay with `visibility: hidden` first, then measure its real height via `getBoundingClientRect()`, then position it.
- New `positionErrorOverlay(overlay, row)` helper. Policy: **prefer above whenever there's room above**; fall back to below otherwise. Tree views in IDEs usually have more chrome below them (status bars, help panels) than above, so above is the safer default. Matches the policy the VS Code extension used to apply as a client-side shim, which can now be removed.
- When flipping above, use `rect.top - overlayH - gap` so the overlay's bottom sits a gap above the row's top.
- Clear `bottom` and `transform` defensively when writing `top`.
- Add a `window.resize` listener that re-runs the positioning, with matching teardown in `dismissErrorOverlay`.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./domain/ide/treeview/...` passes
- [x] Manually verified in IntelliJ IDEA 2025.3 by clicking a Secrets product-error row at the bottom of the sidebar: overlay now flips above instead of being clipped
- [ ] Smoke-check on remaining IDE plugins (VS Code, Eclipse, Visual Studio) — happy path, no regression for rows that have room below

## Follow-up

Once this ships in a snyk-ls release and IDEs pin to that LS protocol version, the VS Code-side `treeViewOverlayPositioner` shim and its tests can be removed (tracked separately under IDE-1808 cleanup).

### DIFF
--- a/domain/ide/treeview/template/tree.js
+++ b/domain/ide/treeview/template/tree.js
@@ -77,7 +77,6 @@
   var activeErrorOutsideClick = null;
   var activeErrorResize = null;
   var activeErrorRow = null;
-  var ERROR_OVERLAY_GAP = 4;
 
   function dismissErrorOverlay() {
     if (activeErrorOverlay && activeErrorOverlay.parentNode) {
@@ -110,19 +109,27 @@
   function positionErrorOverlay(overlay, row) {
     if (!overlay || !row) return;
     var rect = row.getBoundingClientRect();
+    // 600 = sensible fallback viewport width when neither window.innerWidth
+    // nor documentElement.clientWidth is reported (very old WebViews / tests).
     var vw = window.innerWidth || document.documentElement.clientWidth || 600;
-    var vh = window.innerHeight || document.documentElement.clientHeight || 400;
     var overlayH = overlay.getBoundingClientRect().height;
     if (overlayH <= 0) return;
 
-    var gap = ERROR_OVERLAY_GAP;
+    // 4 = gap in px between the row and the overlay (and a minimum margin
+    // from the viewport edge).
+    var gap = 4;
     var topPos = rect.bottom + gap;
     if (rect.top - overlayH - gap >= gap) {
       topPos = rect.top - overlayH - gap;
     }
     topPos = Math.max(gap, topPos);
 
+    // 520 = preferred max overlay width (keeps error text at a comfortable
+    // reading width); 16 = total horizontal viewport padding (8px each side)
+    // so a narrow viewport still leaves a small margin.
     var overlayW = Math.min(520, vw - 16);
+    // 8 = right-edge margin: when the row is far right, pin the overlay so
+    // there is still ~8px between its right edge and the viewport.
     var leftPos = Math.max(gap, Math.min(rect.left, vw - overlayW - 8));
 
     overlay.style.position = 'fixed';
--- a/domain/ide/treeview/template/tree.js
+++ b/domain/ide/treeview/template/tree.js
@@ -101,9 +101,10 @@
 
   // Computes and applies the final overlay position using its measured height,
   // so a tall overlay against a row at the bottom of the viewport flips above
-  // the row instead of being clipped (IDE-1808). Defaults to below; flips above
-  // only when there isn't room below AND there is room above — otherwise the
-  // overlay stays below (clipped is preferable to overlapping the row).
+  // the row instead of being clipped (IDE-1808). Prefers above whenever there
+  // is room above: tree views in IDEs typically have more chrome below them
+  // (status bars, help panels) than above, so above is the safer default and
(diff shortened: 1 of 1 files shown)


domain/ide/treeview/template/tree.js
```diff
@@ -75,6 +75,9 @@
   var activeErrorOverlay = null;
   var activeErrorKeyDown = null;
   var activeErrorOutsideClick = null;
+  var activeErrorResize = null;
+  var activeErrorRow = null;
+  var ERROR_OVERLAY_GAP = 4;
```
andrewrobinsonhodges-snyk: Why is this a global var? We hard-code plenty of numbers in the function where it is used; I don't understand why this is different.

domain/ide/treeview/template/tree.js
```diff
@@ -88,7 +91,46 @@
...
+  // the row instead of being clipped (IDE-1808). Prefers above whenever there
+  // is room above: tree views in IDEs typically have more chrome below them
+  // (status bars, help panels) than above, so above is the safer default and
+  // matches the behavior of the previous VS Code-side shim.
+  // Clears `bottom` and `transform` defensively so future style sources can't
+  // stretch the overlay between opposing anchors.
+  function positionErrorOverlay(overlay, row) {
+    if (!overlay || !row) return;
+    var rect = row.getBoundingClientRect();
+    var vw = window.innerWidth || document.documentElement.clientWidth || 600;
```
andrewrobinsonhodges-snyk, line 120: Would be nice to have a one-line comment explaining the choice of numbers

domain/ide/treeview/template/tree.js
```diff
@@ -88,7 +91,46 @@
...
+    var vh = window.innerHeight || document.documentElement.clientHeight || 400;
```
andrewrobinsonhodges-snyk: Same here

domain/ide/treeview/template/tree.js
```diff
@@ -88,7 +91,46 @@
...
+    if (overlayH <= 0) return;
+
+    var gap = ERROR_OVERLAY_GAP;
+    var topPos = rect.bottom + gap;
+    if (rect.top - overlayH - gap >= gap) {
+      topPos = rect.top - overlayH - gap;
+    }
+    topPos = Math.max(gap, topPos);
+
+    var overlayW = Math.min(520, vw - 16);
```
andrewrobinsonhodges-snyk: Same here

domain/ide/treeview/template/tree.js
```diff
@@ -88,7 +91,46 @@
...
+    var leftPos = Math.max(gap, Math.min(rect.left, vw - overlayW - 8));
```
andrewrobinsonhodges-snyk, line 150: Same here

andrewrobinsonhodges-snyk (approved):

domain/ide/treeview/template/tree.js
acke: Good point — inlined it as a local var gap = 4 to match the other hard-coded numbers in the function.

domain/ide/treeview/template/tree.js
acke: Added a comment. Also removed the adjacent unused vh declaration so it doesn't need explaining.

domain/ide/treeview/template/tree.js
acke: Removed — the vh variable was declared but never used in this function.

domain/ide/treeview/template/tree.js
acke: Added a comment explaining 520 (preferred max width) and 16 (8px padding on each side).

domain/ide/treeview/template/tree.js
acke: Added a comment explaining the 8px right-edge margin.

PR #1302, opened 2026-05-28
refactor: migrate LSP handler deps from di.*() globals to context injection [IDE-1898]

bastiandoetsch:
## Summary
- Migrate all `di.*()` global accessor calls in LSP handler functions to context-based dependency injection
- Extend `di.Dependencies` struct with handler-accessed fields: `Scanner`, `HoverService`, `ScanNotifier`, `ScanPersister`, `FileWatcher`, `ErrorReporter`, `CodeActionService`, `FeatureFlagService`
- Add nil-guarded injections for all new fields in `withContext` middleware
- Add typed context extractors for each new field (`hoverServiceFromContext`, `scannerFromContext`, etc.)
- Remove dead `initializerFromContext` / `installerFromContext` helpers (Installer/Initializer are process-lifecycle deps, not per-request)
- Fix `logError` to call `di.ErrorReporter().CaptureError(err)` (regression restore)
- Add nil guard in `notifyLockedFieldsRejected`; soft-fail in `refreshLdxSyncOnTokenChange`

## PR Stack — Merge Order
```mermaid
flowchart LR
    main(["main"])
    PR1["#1302 PR-1 ← YOU ARE HERE\nDI context injection\n(production code)"]
    PR2["#1303 PR-2\nTDD contract tests"]
    main --> PR1 --> PR2
    style PR1 fill:#ffd700,color:#000
```

## Deferred to later PRs
- `di.TestInit` write-path still mutates package-level globals (PR2 notes this as a known gap; full parallel test safety requires fixing this too)
- Full `t.Parallel()` adoption once the write-path is fixed

## Test plan
- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./application/server/...` passes with `-race`
- [x] `go test ./application/di/...` passes
- [x] PR size: 601 changed lines (under 700 limit)

___

### **PR Type**
Enhancement, Bug fix

___

### **Description**
- Migrate LSP handler dependencies to context injection.

- Introduce `must*FromContext` helpers for mandatory dependencies.

- Enhance nil-safety and error handling in config updates.

- Update DI struct, tests, and server startup logic.

___

<details> <summary><h3> File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><details><summary>5 files</summary><table>
<tr>
[...]

PR #1308, opened 2026-05-28
feat(folder): emit failure analytics for failed scans [IDE-1668]

acke:
## Summary

- snyk-ls now emits an analytics interaction with `status:failure` when a scan errors, so the [Is Snyk OK?](https://app.datadoghq.com/dashboard/f56-fxs-zmp) dashboard's IDE panel starts populating.
- New helpers `categorizeError` and `errorCode` derive a low-cardinality category (e.g. `SNYK-CLI`) and the full catalog code (e.g. `SNYK-CLI-0008`) from `github.com/snyk/error-catalog-golang-public/snyk_errors`, already imported in three other snyk-ls call sites.
- `ProcessResults` now calls `sendAnalytics` on the error branch (with the same reference-baseline guard the success path uses) and `sendAnalytics` no longer short-circuits on `data.Err`.

## Why

The dashboard query

```
@service:analytics-service message:"analytics payload"
@analytics-service.interaction.status:failure
@analytics-service.runtime.application.name:snyk-ls
```

returned zero results in production despite real IDE scan failures. Two specific spots dropped the emission: `ProcessResults` short-circuited on `scanData.Err != nil`, and `sendAnalytics` had its own early-return guard and a hardcoded `Status: gafanalytics.Success`. This PR fixes both.

The CLI side already does the equivalent in `cliv2/cmd/cliv2/instrumentation.go:77-80` (`if exitCode == 2 { SetStatus(analytics.Failure) }`), which is why the CLI panel works today.

## What's in the payload now

| Field | Success | Failure (catalog error) | Failure (non-catalog error) |
|---|---|---|---|
| `interaction.status` | `success` | `failure` | `failure` |
| `extension.error_category` | (absent) | catalog prefix, e.g. `SNYK-CLI` | `unknown` |
| `extension.error_code` | (absent) | full code, e.g. `SNYK-CLI-0008` | (absent) |

Raw `err.Error()` strings are never passed to analytics fields — only stable catalog identifiers — to keep events free of file paths, repo URLs, and other PII.

## PR stack

```mermaid
flowchart LR
  main --> pr["#CURRENT: emit failure analytics ← you are here"]
```

Single PR, no stacking needed (~415 lines changed, under the 700-line limit).

## Test plan

- [x] `make test` (full suite) — green
- [x] `make format` — 0 issues
- [x] New tests for the helpers cover: nil, non-catalog, catalog `SNYK-CLI-*`, catalog `SNYK-OS-*`, malformed code, wrapped catalog errors.
- [x] New integration-level tests in `folder_failure_analytics_test.go` cover:
  - failed scan with catalog error → `status:failure` + correct prefix + full code
  - failed scan with non-catalog error → `status:failure` + `error_category:"unknown"`, no `error_code`
  - successful scan still emits `status:success` (regression guard, no error fields)
  - failed scan with empty `Product` → no emission
  - failed scan with `SendAnalytics:false` → no emission
  - failed scan with partial `ScanData` → no panic, still emits Failure
- [x] Existing `Test_processResults_ShouldSendAnalyticsToAPI` still passes (regression guard).
- [ ] Post-deploy: confirm the IDE panel on dashboard `f56-fxs-zmp` starts showing non-zero failure counts within one scan cycle.

## Acceptance criteria (from RFC)

- [x] Early-return on `data.Err != nil` in `sendAnalytics` removed.
- [x] `Status` derived from `data.Err` rather than hardcoded to `Success`.
- [x] `extension["error_category"]` set on failure via catalog prefix, `unknown` fallback for non-catalog errors.
- [x] `extension["error_code"]` set when a full catalog `ErrorCode` is available.
- [x] Unit tests cover success, catalog error, non-catalog error, empty product, `SendAnalytics:false`, partial `ScanData`.
- [x] `createTestSummary` handles partial `ScanData` without panicking (existing early-return on empty severity counts already handles this; T6 confirms).
- [x] No raw `err.Error()` strings appear in any analytics field.
- [ ] Post-deploy verification on the dashboard.

## References

- **RFC v7**: https://snyksec.atlassian.net/wiki/spaces/IDE/pages/4827676673/RFC+Emit+Errors+from+Snyk+Language+Server
- **Jira**: [IDE-1668](https://snyksec.atlassian.net/browse/IDE-1668)
- **Dashboard**: https://app.datadoghq.com/dashboard/f56-fxs-zmp
- **Error catalog**: https://github.com/snyk/error-catalog-golang-public

[IDE-1668]: https://snyksec.atlassian.net/browse/IDE-1668?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

___

### **PR Type**
Enhancement, Tests

___

### **Description**
- Emit analytics events for failed scans to populate the "Is Snyk OK?" dashboard.

- Categorize errors into low-cardinality prefixes and full codes for improved reporting.

- Ensure existing analytics guards for cancellation and opt-outs remain effective.

___

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><details><summary>4 files</summary><table>
<tr>
  <td><strong>analytics_errors.go</strong><dd><code>Add functions to classify and control scan analytics emission</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1308/changes#diff-2bfec5ce7cec0d627af7bd9198e6e442623e5931ae0ec088321189b08dc5b5da">+80/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>folder.go</strong><dd><code>Update scan processing and analytics for error reporting</code>&nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1308/changes#diff-92ee5447150d64c4b9a6c8c8e8834577ada9e739d1a0a2842ef5bc79fd014f26">+40/-19</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>scanner.go</strong><dd><code>Route pre-scan errors for analytics and avoid duplicate notifications</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1308/changes#diff-ca9273debc8b1b659b1a630df729fe1be56f139b823c0b6d4d5a296d55ba47cb">+10/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>scan.go</strong><dd><code>Add UserNotified flag to ScanData</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1308/changes#diff-f0637e314c04ee2a64ce1284958e7e4b782719a4d225264a67d302eb8225aaa0">+5/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Tests</strong></td><td><details><summary>3 files</summary><table>
<tr>
  <td><strong>analytics_errors_test.go</strong><dd><code>Unit tests for error classification and analytics emission logic</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1308/changes#diff-7f31fcb38d6acfc25d8eb5d3b017f79a26d13b77a822b1ad48965c3316a12fa3">+85/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>folder_failure_analytics_test.go</strong><dd><code>Comprehensive tests for failure analytics reporting</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1308/changes#diff-cfb8d9ba9012fbbf2707589fac82884563ddadd847e5ab19cc1296d3b914a579">+352/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>channels.go</strong><dd><code>Add utility to assert channel never receives data</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1308/changes#diff-46d04268ab60ff7ca13961938b8b33cd6efac87179b0d1ae1da2ffe7312f6c40">+16/-0</a>&nbsp; &nbsp; </td>

</tr>

</details>

___

### DIFF
--- a/domain/ide/workspace/analytics_errors.go
+++ b/domain/ide/workspace/analytics_errors.go
@@ -23,44 +23,28 @@ import (
 	"github.com/snyk/error-catalog-golang-public/snyk_errors"
 )
 
-// categorizeError returns a low-cardinality product prefix derived from the
-// Snyk error catalog code on err (e.g. "SNYK-CLI", "SNYK-OS"). For errors that
-// do not carry a catalog entry it returns "unknown". For a nil error it
-// returns the empty string.
+// classifyError returns the analytics (category, code) pair for err, unwrapping
+// the Snyk error catalog entry exactly once so the two values cannot drift.
 //
-// Why a prefix and not the full ErrorCode: ErrorCodes are ~hundreds across the
-// catalog and would inflate analytics cardinality without adding signal to the
-// default "Is Snyk OK?" dashboard rollup. The full code is emitted alongside
-// via errorCode() for drill-downs.
-func categorizeError(err error) string {
+// category is a low-cardinality product prefix (e.g. "SNYK-CLI") suitable for
+// the default "Is Snyk OK?" dashboard rollup. code is the full catalog code
+// (e.g. "SNYK-CLI-0008") for drill-down.
+//
+// For a nil error: ("", ""). For errors without a catalog entry:
+// ("unknown", ""). Raw error.Error() text is never returned — only stable
+// catalog identifiers — to keep analytics free of PII.
+func classifyError(err error) (category, code string) {
 	if err == nil {
-		return ""
+		return "", ""
 	}
 	var snykErr snyk_errors.Error
-	if stderrors.As(err, &snykErr) && snykErr.ErrorCode != "" {
-		parts := strings.SplitN(snykErr.ErrorCode, "-", 3)
-		if len(parts) >= 2 {
-			return parts[0] + "-" + parts[1]
-		}
-		return snykErr.ErrorCode
+	if !stderrors.As(err, &snykErr) || snykErr.ErrorCode == "" {
+		return "unknown", ""
 	}
-	return "unknown"
-}
-
-// errorCode returns the full Snyk catalog ErrorCode (e.g. "SNYK-CLI-0008") for
-// errors that wrap a catalog entry, or the empty string otherwise. It is
-// emitted alongside categorizeError so dashboards can drill from prefix bucket
-// (e.g. "SNYK-CLI") into the specific code.
-//
-// Raw error.Error() text is never returned — only stable catalog identifiers —
-// to keep analytics free of file paths, repo URLs, and other PII.
-func errorCode(err error) string {
-	if err == nil {
-		return ""
+	code = snykErr.ErrorCode
+	parts := strings.SplitN(code, "-", 3)
+	if len(parts) >= 2 && parts[0] != "" && parts[1] != "" {
+		return parts[0] + "-" + parts[1], code
 	}
-	var snykErr snyk_errors.Error
-	if stderrors.As(err, &snykErr) {
-		return snykErr.ErrorCode
-	}
-	return ""
+	return code, code
 }
--- a/domain/ide/workspace/analytics_errors.go
+++ b/domain/ide/workspace/analytics_errors.go
@@ -1,0 +1,66 @@
+/*
+ * © 2026 Snyk Limited All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package workspace
+
+import (
+	stderrors "errors"
+	"strings"
+
+	"github.com/snyk/error-catalog-golang-public/snyk_errors"
+)
+
+// categorizeError returns a low-cardinality product prefix derived from the
+// Snyk error catalog code on err (e.g. "SNYK-CLI", "SNYK-OS"). For errors that
+// do not carry a catalog entry it returns "unknown". For a nil error it
+// returns the empty string.
+//
+// Why a prefix and not the full ErrorCode: ErrorCodes are ~hundreds across the
+// catalog and would inflate analytics cardinality without adding signal to the
+// default "Is Snyk OK?" dashboard rollup. The full code is emitted alongside
+// via errorCode() for drill-downs.
+func categorizeError(err error) string {
+	if err == nil {
+		return ""
+	}
+	var snykErr snyk_errors.Error
+	if stderrors.As(err, &snykErr) && snykErr.ErrorCode != "" {
+		parts := strings.SplitN(snykErr.ErrorCode, "-", 3)
+		if len(parts) >= 2 {
+			return parts[0] + "-" + parts[1]
+		}
+		return snykErr.ErrorCode
+	}
+	return "unknown"
+}
+
+// errorCode returns the full Snyk catalog ErrorCode (e.g. "SNYK-CLI-0008") for
+// errors that wrap a catalog entry, or the empty string otherwise. It is
+// emitted alongside categorizeError so dashboards can drill from prefix bucket
+// (e.g. "SNYK-CLI") into the specific code.
+//
+// Raw error.Error() text is never returned — only stable catalog identifiers —
+// to keep analytics free of file paths, repo URLs, and other PII.
+func errorCode(err error) string {
+	if err == nil {
+		return ""
+	}
+	var snykErr snyk_errors.Error
+	if stderrors.As(err, &snykErr) {
+		return snykErr.ErrorCode
+	}
+	return ""
+}
--- a/domain/ide/workspace/analytics_errors.go
+++ b/domain/ide/workspace/analytics_errors.go
@@ -22,11 +22,33 @@ import (
 	"strings"
 
 	"github.com/snyk/error-catalog-golang-public/snyk_errors"
+
+	"github.com/snyk/snyk-ls/infrastructure/utils"
+	"github.com/snyk/snyk-ls/internal/types"
 )
 
+// shouldEmitAnalytics centralizes the "do we record this scan on the
+// 'Is Snyk OK?' dashboard?" policy. Checked at the call site so cancelled and
+// non-failing scans skip the goroutine entirely instead of spawning one that
+// bails after categorization work.
+func shouldEmitAnalytics(data *types.ScanData) bool {
+	if data == nil || !data.SendAnalytics || data.Product == "" {
+		return false
+	}
+	if data.Err != nil {
+		if utils.IsNonFailingScanError(data.Err.Error()) {
+			return false
+		}
+		if isCancellationError(data.Err) {
+			return false
+		}
+	}
+	return true
+}
+
 // isCancellationError reports whether err represents a routine cancellation
-// (user abort, credential rotation, timeout). Analytics callers skip emission
-// for these so cancellations don't count on the "Is Snyk OK?" rollup.
+// (user abort, credential rotation, timeout). Routine cancellations are not
+// scan failures and must not count on the "Is Snyk OK?" rollup.
 func isCancellationError(err error) bool {
 	return err != nil && (stderrors.Is(err, context.Canceled) || stderrors.Is(err, context.DeadlineExceeded))
 }
--- a/domain/ide/workspace/analytics_errors.go
+++ b/domain/ide/workspace/analytics_errors.go
@@ -17,11 +17,19 @@
 package workspace
 
 import (
+	"context"
 	stderrors "errors"
 	"strings"
 
 	"github.com/snyk/error-catalog-golang-public/snyk_errors"
 )
+
+// isCancellationError reports whether err represents a routine cancellation
+// (user abort, credential rotation, timeout). Analytics callers skip emission
+// for these so cancellations don't count on the "Is Snyk OK?" rollup.
+func isCancellationError(err error) bool {
+	return err != nil && (stderrors.Is(err, context.Canceled) || stderrors.Is(err, context.DeadlineExceeded))
+}
 
 // classifyError returns the analytics (category, code) pair for err, unwrapping
 // the Snyk error catalog entry exactly once so the two values cannot drift.
--- a/domain/ide/workspace/analytics_errors.go
+++ b/domain/ide/workspace/analytics_errors.go
@@ -28,7 +28,7 @@ import (
 )
 
 // shouldEmitAnalytics centralizes the "do we record this scan on the
-// 'Is Snyk OK?' dashboard?" policy. Checked at the call site so cancelled and
+// 'Is Snyk OK?' dashboard?" policy. Checked at the call site so canceled and
 // non-failing scans skip the goroutine entirely instead of spawning one that
 // bails after categorization work.
 func shouldEmitAnalytics(data *types.ScanData) bool {
--- a/domain/ide/workspace/folder.go
+++ b/domain/ide/workspace/folder.go
@@ -493,8 +493,8 @@ func sendAnalytics(ctx context.Context, engine workflow.Engine, configResolver t
 	// scanner.cancelFunc) surface as context.Canceled / DeadlineExceeded. These
 	// are not scan failures — skip emission entirely so they don't show up as
 	// either status on the "Is Snyk OK?" rollup.
-	if data.Err != nil && (errors.Is(data.Err, context.Canceled) || errors.Is(data.Err, context.DeadlineExceeded)) {
-		log.Debug().Err(data.Err).Msg("skipping analytics emission for cancelled scan")
+	if isCancellationError(data.Err) {
+		log.Debug().Err(data.Err).Msg("skipping analytics emission for canceled scan")
 		return
 	}
 
--- a/domain/ide/workspace/folder.go
+++ b/domain/ide/workspace/folder.go
@@ -358,6 +358,13 @@ func (f *Folder) scan(ctx context.Context, path types.FilePath) {
 func (f *Folder) ProcessResults(ctx context.Context, scanData types.ScanData) {
 	if scanData.Err != nil {
 		f.sendScanError(scanData.Product, scanData.Err)
+		// IDE-1668: emit failure analytics so the "Is Snyk OK?" dashboard's
+		// IDE panel reflects scan errors. Skip for reference baselines, matching
+		// the success path's IsReferenceScan guard (baseline scans aren't a
+		// user-meaningful "scan done" interaction).
+		if !scanData.IsReferenceScan || f.configResolver.GetBool(types.SettingScanNetNew, f.FolderConfigReadOnly()) {
+			go sendAnalytics(ctx, f.engine, f.configResolver, f.logger, &scanData)
+		}
 		return
 	}
 
@@ -459,11 +466,6 @@ func sendAnalytics(ctx context.Context, engine workflow.Engine, configResolver t
 		return
 	}
 
-	if data.Err != nil {
-		log.Debug().Err(data.Err).Msg("Skipping analytics for error")
-		return
-	}
-
 	// this information is not filled automatically, so we need to collect it
 	folderConfig := config.GetFolderConfigFromEngine(engine, configResolver, data.Path, logger)
 	categories := setupCategories(data, configResolver, engine, folderConfig)
@@ -485,10 +487,23 @@ func sendAnalytics(ctx context.Context, engine workflow.Engine, configResolver t
 		extension["scan_type"] = deltaScanType.String()
 	}
 
+	// IDE-1668: emit failure analytics events alongside success ones so the
+	// "Is Snyk OK?" dashboard's IDE panel reflects scan errors. categorizeError
+	// returns a low-cardinality catalog prefix (e.g. "SNYK-CLI") for rollups;
+	// errorCode adds the full code (e.g. "SNYK-CLI-0008") for drill-downs.
+	status := gafanalytics.Success
+	if data.Err != nil {
+		status = gafanalytics.Failure
+		extension["error_category"] = categorizeError(data.Err)
+		if code := errorCode(data.Err); code != "" {
+			extension["error_code"] = code
+		}
+	}
+
 	param := types.AnalyticsEventParam{
 		InteractionType: "Scan done",
 		Category:        categories,
-		Status:          string(gafanalytics.Success),
+		Status:          string(status),
 		TargetId:        targetId,
 		TimestampMs:     data.TimestampFinished.UnixMilli(),
 		DurationMs:      int64(data.Duration),
--- a/domain/ide/workspace/folder.go
+++ b/domain/ide/workspace/folder.go
@@ -359,11 +359,10 @@ func (f *Folder) scan(ctx context.Context, path types.FilePath) {
 func (f *Folder) ProcessResults(ctx context.Context, scanData types.ScanData) {
 	if scanData.Err != nil {
 		f.sendScanError(scanData.Product, scanData.Err)
-		// IDE-1668: emit failure analytics so the "Is Snyk OK?" dashboard's
-		// IDE panel reflects scan errors. Mirror sendScanError's filter — auth-not-set
-		// and "product disabled" outcomes are user state, not failures, and would
-		// inflate the failure rate one event per enabled product per scan.
-		// Reference scans short-circuit inside sendAnalytics via SendAnalytics:false.
+		// Mirror sendScanError's filter — auth-not-set and "product disabled" outcomes
+		// are user state, not failures, and would inflate the dashboard failure rate one
+		// event per enabled product per scan. Reference scans short-circuit inside
+		// sendAnalytics via SendAnalytics:false.
 		if !utils.IsNonFailingScanError(scanData.Err.Error()) {
 			go sendAnalytics(ctx, f.engine, f.configResolver, f.logger, &scanData)
 		}
--- a/domain/ide/workspace/folder.go
+++ b/domain/ide/workspace/folder.go
@@ -23,6 +23,7 @@ import (
 	"errors"
 	"fmt"
 	"sync"
+	"time"
 
 	"github.com/rs/zerolog"
 	"github.com/samber/lo"
@@ -359,10 +360,11 @@ func (f *Folder) ProcessResults(ctx context.Context, scanData types.ScanData) {
 	if scanData.Err != nil {
 		f.sendScanError(scanData.Product, scanData.Err)
 		// IDE-1668: emit failure analytics so the "Is Snyk OK?" dashboard's
-		// IDE panel reflects scan errors. Skip for reference baselines, matching
-		// the success path's IsReferenceScan guard (baseline scans aren't a
-		// user-meaningful "scan done" interaction).
-		if !scanData.IsReferenceScan || f.configResolver.GetBool(types.SettingScanNetNew, f.FolderConfigReadOnly()) {
+		// IDE panel reflects scan errors. Mirror sendScanError's filter — auth-not-set
+		// and "product disabled" outcomes are user state, not failures, and would
+		// inflate the failure rate one event per enabled product per scan.
+		// Reference scans short-circuit inside sendAnalytics via SendAnalytics:false.
+		if !utils.IsNonFailingScanError(scanData.Err.Error()) {
 			go sendAnalytics(ctx, f.engine, f.configResolver, f.logger, &scanData)
 		}
 		return
@@ -487,17 +489,35 @@ func sendAnalytics(ctx context.Context, engine workflow.Engine, configResolver t
 		extension["scan_type"] = deltaScanType.String()
 	}
 
+	// IDE-1668: routine cancellations (credential rotation, user abort via
+	// scanner.cancelFunc) surface as context.Canceled / DeadlineExceeded. These
+	// are not scan failures — skip emission entirely so they don't show up as
+	// either status on the "Is Snyk OK?" rollup.
+	if data.Err != nil && (errors.Is(data.Err, context.Canceled) || errors.Is(data.Err, context.DeadlineExceeded)) {
+		log.Debug().Err(data.Err).Msg("skipping analytics emission for cancelled scan")
+		return
+	}
+
 	// IDE-1668: emit failure analytics events alongside success ones so the
-	// "Is Snyk OK?" dashboard's IDE panel reflects scan errors. categorizeError
-	// returns a low-cardinality catalog prefix (e.g. "SNYK-CLI") for rollups;
-	// errorCode adds the full code (e.g. "SNYK-CLI-0008") for drill-downs.
+	// "Is Snyk OK?" dashboard's IDE panel reflects scan errors. classifyError
+	// returns (category, code): the category is a low-cardinality catalog
+	// prefix (e.g. "SNYK-CLI") for rollups, the code is the full catalog code
+	// (e.g. "SNYK-CLI-0008") for drill-down.
 	status := gafanalytics.Success
 	if data.Err != nil {
 		status = gafanalytics.Failure
-		extension["error_category"] = categorizeError(data.Err)
-		if code := errorCode(data.Err); code != "" {
+		category, code := classifyError(data.Err)
+		extension["error_category"] = category
+		if code != "" {
 			extension["error_code"] = code
 		}
+	}
+
+	// Failure paths may build ScanData without TimestampFinished; the zero
+	// time.Time produces a negative epoch via UnixMilli(). Default to now.
+	timestampFinished := data.TimestampFinished
+	if timestampFinished.IsZero() {
+		timestampFinished = time.Now().UTC()
 	}
 
 	param := types.AnalyticsEventParam{
@@ -505,7 +525,7 @@ func sendAnalytics(ctx context.Context, engine workflow.Engine, configResolver t
 		Category:        categories,
 		Status:          string(status),
 		TargetId:        targetId,
-		TimestampMs:     data.TimestampFinished.UnixMilli(),
+		TimestampMs:     timestampFinished.UnixMilli(),
 		DurationMs:      int64(data.Duration),
 		Extension:       extension,
 	}
--- a/domain/ide/workspace/folder.go
+++ b/domain/ide/workspace/folder.go
@@ -358,7 +358,13 @@ func (f *Folder) scan(ctx context.Context, path types.FilePath) {
 
 func (f *Folder) ProcessResults(ctx context.Context, scanData types.ScanData) {
 	if scanData.Err != nil {
-		f.sendScanError(scanData.Product, scanData.Err, scanData.UserNotified)
+		// UserNotified means the caller has already surfaced the error (toast +
+		// any diagnostic) and only routed through here for analytics emission.
+		// Skip sendScanError entirely so we don't fire a duplicate SendError or
+		// an unwanted SendErrorDiagnostic.
+		if !scanData.UserNotified {
+			f.sendScanError(scanData.Product, scanData.Err)
+		}
 		if shouldEmitAnalytics(&scanData) {
 			go sendAnalytics(ctx, f.engine, f.configResolver, f.logger, &scanData)
 		}
@@ -387,10 +393,8 @@ func (f *Folder) ProcessResults(ctx context.Context, scanData types.ScanData) {
 	f.FilterAndPublishDiagnostics(scanData.Product)
 }
 
-func (f *Folder) sendScanError(product product.Product, err error, userAlreadyNotified bool) {
-	if !userAlreadyNotified {
-		f.scanNotifier.SendError(product, f.path, err.Error())
-	}
+func (f *Folder) sendScanError(product product.Product, err error) {
+	f.scanNotifier.SendError(product, f.path, err.Error())
 
 	if utils.IsNonFailingScanError(err.Error()) {
 		f.logger.Debug().
@@ -847,7 +851,7 @@ func (f *Folder) publishDiagnostics(p product.Product, issuesToSendByProduct sny
 	f.sendDiagnostics(issuesToSendByProduct.AggregateFromAllProducts(p))
 	scanErr := f.scanStateAggregator.GetScanErr(f.path, p, f.IsDeltaFindingsEnabled())
 	if scanErr != nil {
-		f.sendScanError(p, scanErr, false)
+		f.sendScanError(p, scanErr)
 	} else {
 		f.sendSuccess(p)
 	}
--- a/domain/ide/workspace/folder.go
+++ b/domain/ide/workspace/folder.go
@@ -358,12 +358,8 @@ func (f *Folder) scan(ctx context.Context, path types.FilePath) {
 
 func (f *Folder) ProcessResults(ctx context.Context, scanData types.ScanData) {
 	if scanData.Err != nil {
-		f.sendScanError(scanData.Product, scanData.Err)
-		// Mirror sendScanError's filter — auth-not-set and "product disabled" outcomes
-		// are user state, not failures, and would inflate the dashboard failure rate one
-		// event per enabled product per scan. Reference scans short-circuit inside
-		// sendAnalytics via SendAnalytics:false.
-		if !utils.IsNonFailingScanError(scanData.Err.Error()) {
+		f.sendScanError(scanData.Product, scanData.Err, scanData.UserNotified)
+		if shouldEmitAnalytics(&scanData) {
 			go sendAnalytics(ctx, f.engine, f.configResolver, f.logger, &scanData)
 		}
 		return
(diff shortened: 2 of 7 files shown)


domain/ide/workspace/folder.go
```diff
@@ -358,6 +358,13 @@ func (f *Folder) scan(ctx context.Context, path types.FilePath) {
 func (f *Folder) ProcessResults(ctx context.Context, scanData types.ScanData) {
 	if scanData.Err != nil {
 		f.sendScanError(scanData.Product, scanData.Err)
+		// IDE-1668: emit failure analytics so the "Is Snyk OK?" dashboard's
+		// IDE panel reflects scan errors. Skip for reference baselines, matching
+		// the success path's IsReferenceScan guard (baseline scans aren't a
+		// user-meaningful "scan done" interaction).
+		if !scanData.IsReferenceScan || f.configResolver.GetBool(types.SettingScanNetNew, f.FolderConfigReadOnly()) {
+			go sendAnalytics(ctx, f.engine, f.configResolver, f.logger, &scanData)
```
bastiandoetsch, line 369: [Critical] This emits failure analytics whenever `scanData.Err != nil`, but `sendScanError` just above treats a set of errors as non-failures via `utils.IsNonFailingScanError` (not-authenticated, "Snyk Code/OSS/IaC/Secrets not enabled for this folder/org"). So an unauthenticated user emits a `status:"failure"` event — one per enabled product — on every scan, inflating the same "Is Snyk OK?" failure rate this feature is meant to make trustworthy.

Fix: gate emission on the same check `sendScanError` already uses.
```go
if scanData.Err != nil && !utils.IsNonFailingScanError(scanData.Err.Error()) {
    go sendAnalytics(ctx, f.engine, f.configResolver, f.logger, &scanData)
}
```
Add a test with `ScanData{Err: errors.New(utils.MsgNotAuthenticatedNoScan), SendAnalytics: true}` asserting the analytics workflow is invoked `Times(0)` — it fails today.

— AI review

bastiandoetsch (commented):
Automated verification review of IDE-1668 — see inline comments. Two findings that fall outside the diff (the pre-scan-command failure path in scanner.go and the pre-existing sendAnalytics size) are posted as separate top-level comments. — AI review

domain/ide/workspace/folder.go
```diff
@@ -358,6 +358,13 @@ func (f *Folder) scan(ctx context.Context, path types.FilePath) {
 func (f *Folder) ProcessResults(ctx context.Context, scanData types.ScanData) {
 	if scanData.Err != nil {
 		f.sendScanError(scanData.Product, scanData.Err)
+		// IDE-1668: emit failure analytics so the "Is Snyk OK?" dashboard's
+		// IDE panel reflects scan errors. Skip for reference baselines, matching
+		// the success path's IsReferenceScan guard (baseline scans aren't a
+		// user-meaningful "scan done" interaction).
+		if !scanData.IsReferenceScan || f.configResolver.GetBool(types.SettingScanNetNew, f.FolderConfigReadOnly()) {
```
bastiandoetsch: [Should Fix] This `IsReferenceScan`/`SettingScanNetNew` guard is the manual opposite (the same rule with every boolean flipped) of the success-path guard at folder.go:382. Keeping one policy written two ways in opposite boolean forms means the two copies can drift out of sync. It is also likely dead: a reference scan builds `ScanData` with `SendAnalytics:false` and no `Err` (scanner.go:330-338), so an error-bearing reference scan never reaches this branch, and `sendAnalytics` already returns early when `!SendAnalytics`.

Fix: drop the guard from the error branch and rely on the existing `SendAnalytics:false` short-circuit, or pull the reference-scan decision into one helper that both branches call.

— AI review

domain/ide/workspace/folder.go
```diff
@@ -485,10 +487,23 @@ func sendAnalytics(ctx context.Context, engine workflow.Engine, configResolver t
 		extension["scan_type"] = deltaScanType.String()
 	}
 
+	// IDE-1668: emit failure analytics events alongside success ones so the
+	// "Is Snyk OK?" dashboard's IDE panel reflects scan errors. categorizeError
+	// returns a low-cardinality catalog prefix (e.g. "SNYK-CLI") for rollups;
+	// errorCode adds the full code (e.g. "SNYK-CLI-0008") for drill-downs.
+	status := gafanalytics.Success
+	if data.Err != nil {
+		status = gafanalytics.Failure
```
bastiandoetsch, line 498: [Should Fix] A cancelled scan is recorded as a failure. When a scan is cancelled — credential rotation, or the user aborting (`cancelFunc()` in scanner.go) — `data.Err` is `context.Canceled`, and this code emits it as `status:"failure"` with `error_category:"unknown"` (cancellation has no catalog code). Routine cancellations shouldn't count against the failure rate.

Fix: skip the failure status when the error is a cancellation.
```go
if data.Err != nil && !errors.Is(data.Err, context.Canceled) && !errors.Is(data.Err, context.DeadlineExceeded) {
    status = gafanalytics.Failure
    // ...
}
```

— AI review

domain/ide/workspace/folder.go
```diff
@@ -485,10 +487,23 @@ func sendAnalytics(ctx context.Context, engine workflow.Engine, configResolver t
...
+		}
+	}
+
 	param := types.AnalyticsEventParam{
 		InteractionType: "Scan done",
 		Category:        categories,
-		Status:          string(gafanalytics.Success),
+		Status:          string(status),
 		TargetId:        targetId,
 		TimestampMs:     data.TimestampFinished.UnixMilli(),
```
bastiandoetsch: [Suggestion] On the failure path, `ScanData` is built without `TimestampFinished`, so the zero `time.Time` makes `UnixMilli()` produce a nonsensical negative epoch (`-62135596800000`, visible in the T6 test payload). It doesn't bite today because the internalScan path sets the timestamp, but this path has no fallback.

Fix: default to now when the timestamp is unset.
```go
ts := data.TimestampFinished
if ts.IsZero() {
    ts = time.Now().UTC()
}
// TimestampMs: ts.UnixMilli()
```

— AI review

domain/ide/workspace/analytics_errors.go
```diff
@@ -0,0 +1,66 @@
...
+// categorizeError returns a low-cardinality product prefix derived from the
+// Snyk error catalog code on err (e.g. "SNYK-CLI", "SNYK-OS"). For errors that
+// do not carry a catalog entry it returns "unknown". For a nil error it
+// returns the empty string.
+//
+// Why a prefix and not the full ErrorCode: ErrorCodes are ~hundreds across the
+// catalog and would inflate analytics cardinality without adding signal to the
+// default "Is Snyk OK?" dashboard rollup. The full code is emitted alongside
+// via errorCode() for drill-downs.
+func categorizeError(err error) string {
```
bastiandoetsch: [Suggestion] `categorizeError` and `errorCode` each run their own `stderrors.As(err, &snykErr)` on the same error for every failure, and their two return values have to stay consistent (the category must be a prefix of the code). Editing one later could let it drift from the other.

Fix (optional): combine them into one helper that unwraps the error once and returns both.
```go
func classifyError(err error) (category, code string) { /* one errors.As */ }
```

— AI review

domain/ide/workspace/analytics_errors.go
```diff
@@ -0,0 +1,66 @@
...
+	if err == nil {
+		return ""
+	}
+	var snykErr snyk_errors.Error
+	if stderrors.As(err, &snykErr) && snykErr.ErrorCode != "" {
+		parts := strings.SplitN(snykErr.ErrorCode, "-", 3)
```
bastiandoetsch: [Suggestion] `SplitN(code, "-", 3)` on a malformed trailing-dash code like `"SNYK-"` returns the category `"SNYK-"`. This is cosmetic only — no real catalog code has this shape — so change it only if convenient (e.g. require both split parts to be non-empty before using them).

— AI review

domain/ide/workspace/folder_failure_analytics_test.go
```diff
@@ -0,0 +1,250 @@
...
+//
+// These tests cover the contract: sendAnalytics() must produce a
+// `interaction.status:"Failure"` event with `error_category` (catalog prefix,
+// e.g. "SNYK-CLI") and, when available, `error_code` (full catalog code).
+// The empty-product and SendAnalytics=false guards must continue to suppress
+// emission even on the failure path.
+
+// T1 — failed scan whose error wraps a snyk catalog entry produces a Failure
+// analytics event with the full error code and its product prefix.
+func Test_processResults_FailedScan_EmitsFailureAnalytics_WithCatalogError(t *testing.T) {
```
bastiandoetsch, line 58: [Should Fix] All six tests call `f.ProcessResults` directly with hand-built `ScanData`, so none of them exercise the real `scanner.Scan -> ProcessResults` wiring. That is why the non-failing-error emission, the context-cancellation mis-classification, and the pre-scan-failure gap are all invisible to this suite. The acceptance criteria also ask for an end-to-end integration test.

Fix: add a test that drives a non-failing error (e.g. `ErrSnykOssNotEnabledForFolder`) through the scanner path asserting `Times(0)`, plus an `application/server` integration test asserting a `status:"failure"` payload reaches the analytics workflow.

— AI review

domain/ide/workspace/folder_failure_analytics_test.go
```diff
@@ -0,0 +1,250 @@
...
+		Err:               stderrors.New("boom"),
+	}
+
+	// Times(0) — gomock will fail the test if the analytics workflow gets invoked.
+	_ = testutil.MockAndCaptureWorkflowInvocation(t, engineMock, localworkflows.WORKFLOWID_REPORT_ANALYTICS, 0)
+
+	f.ProcessResults(t.Context(), data)
+
+	// Small settle window so any (incorrect) async emission would have time to land.
+	time.Sleep(50 * time.Millisecond)
```
bastiandoetsch: [Suggestion] The negative assertions (here and in the `SendAnalyticsFalse` test) rely on `time.Sleep(50ms)` followed by `Times(0)`, which only proves "nothing emitted within 50ms", not "nothing will emit" — a slow CI box could pass a case that should fail.

Fix: wait on a deterministic signal instead of a fixed sleep (e.g. a channel the mock closes, or a `WaitGroup`).

— AI review

bastiandoetsch:
[Should Fix] Pre-scan-command failures emit no analytics at all. In `domain/snyk/scanner/scanner.go:282-287`, when `executePreScanCommand` returns an error the goroutine sends an error notification, calls `SetScanDone(..., err)`, and `return`s before `processResults` runs — so this real scan-blocking failure is never counted on the dashboard. (This line is outside the PR diff, hence a top-level comment.)

Fix: route the error through `processResults` instead of returning, e.g.
```go
if err != nil {
    scanLogger.Err(err).Send()
    sc.scanStateAggregator.SetScanDone(folderPath, scanner.Product(), false, err)
    processResults(span.Context(), types.ScanData{
        Product: s.Product(), Err: err, SendAnalytics: true,
        TimestampFinished: time.Now().UTC(), Path: folderPath,
    })
    return
}
```
Or explicitly scope this path out in the PR description if intentional.

— AI review

bastiandoetsch:
[Suggestion] `sendAnalytics` (`domain/ide/workspace/folder.go:459`) is now ~80 lines with 5 parameters after this change. Pre-existing and the diff is appropriately surgical, so no action needed here — noting it for a future refactor. (Outside the diff hunks, hence a top-level comment.)

— AI review

domain/ide/workspace/folder.go
acke: Fixed in 76d8aa0b. Gated the goroutine launch on `\!utils.IsNonFailingScanError(scanData.Err.Error())` so the documented non-failing outcomes (not-authenticated, product-disabled-for-folder/org) no longer emit failure analytics. Added a table-driven test (`Test_processResults_FailedScan_NonFailingError_NoEmission`) covering MsgNotAuthenticatedNoScan and the four `*NotEnabledForFolder` strings.

domain/ide/workspace/folder.go
acke: Fixed in 76d8aa0b. Dropped the inverted `IsReferenceScan/SettingScanNetNew` guard from the error branch. Reference scans build ScanData with `SendAnalytics:false`, which `sendAnalytics` already short-circuits on at the top — so the explicit branch-level check was redundant.

domain/ide/workspace/folder.go
acke: Fixed in 76d8aa0b. Added an early return in `sendAnalytics` for `errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded)`. Went with skip-emission rather than the suggested skip-failure-status because a cancelled scan as a Success event would still misrepresent the rollup. Added `Test_processResults_FailedScan_Cancellation_NoEmission` covering Canceled, DeadlineExceeded, and a wrapped Canceled (via `errors.Join`).

domain/ide/workspace/folder.go
acke: Fixed in 76d8aa0b. `sendAnalytics` now defaults `TimestampFinished` to `time.Now().UTC()` when zero before computing `timestampMs`. T6 (partial ScanData) now asserts the payload does not contain `"timestampMs":-` — the regression guard for the negative-epoch case.

domain/ide/workspace/analytics_errors.go
acke: Fixed in 76d8aa0b. Consolidated into a single `classifyError(err) (category, code string)` that runs `errors.As` once and returns both. `sendAnalytics` is now `category, code := classifyError(data.Err)` so the two values can't drift.

domain/ide/workspace/analytics_errors.go
acke: Fixed in 76d8aa0b. The split now requires `parts[0] \!= "" && parts[1] \!= ""` before joining; otherwise falls back to the full code. Added `Test_classifyError_CatalogError_TrailingDashIsHardened` for `SNYK-`.

domain/ide/workspace/folder_failure_analytics_test.go
acke: Partially addressed in 76d8aa0b. Added `Test_processResults_FailedScan_NonFailingError_NoEmission` (5 cases) and `Test_processResults_FailedScan_Cancellation_NoEmission` (3 cases) driving `ProcessResults` directly with each non-failing / cancellation error — `Times(0)` proves the failure gate holds. Not adding the application/server end-to-end test here to keep the scope of the rip-out PR tight; happy to follow up in a separate ticket if you want belt-and-braces coverage at the server layer.

domain/ide/workspace/folder_failure_analytics_test.go
acke: Fixed in 76d8aa0b. Added `testsupport.RequireNeverReceive` (negative analog of `RequireEventuallyReceive`) and replaced the `time.Sleep(50ms)` in T4/T5 (and reused it in the new T7/T8 negative tests). `gomock.Times(0)` already fails immediately on any incorrect emission, and the helper now polls instead of sleeping a fixed window.

acke:
Re https://github.com/snyk/snyk-ls/pull/1308#issuecomment-4592462279 — fixed in 76d8aa0b. `scanner.go` now routes the `executePreScanCommand` failure through `processResults` with `SendAnalytics: true` and `TimestampFinished: time.Now().UTC()`, and drops the explicit `scanNotifier.SendError` call (handled by `sendScanError` inside `ProcessResults`). The pre-scan-command failure now counts on the dashboard and goes through the same non-failing/cancellation filters.

domain/snyk/scanner/scanner.go
```diff
@@ -282,8 +282,17 @@ func (sc *DelegatingConcurrentScanner) Scan(ctx context.Context, pathToScan type
 			err := sc.executePreScanCommand(span.Context(), sc.engine, s.Product(), workspaceFolderConfig, folderPath, true)
 			if err != nil {
 				scanLogger.Err(err).Send()
-				sc.scanNotifier.SendError(scanner.Product(), folderPath, err.Error())
 				sc.scanStateAggregator.SetScanDone(folderPath, scanner.Product(), false, err)
+				// IDE-1668: route the pre-scan-command failure through processResults so
```
bastiandoetsch: remove the jira markers, pls

domain/ide/workspace/folder.go
```diff
@@ -358,6 +359,14 @@ func (f *Folder) scan(ctx context.Context, path types.FilePath) {
 func (f *Folder) ProcessResults(ctx context.Context, scanData types.ScanData) {
 	if scanData.Err != nil {
 		f.sendScanError(scanData.Product, scanData.Err)
+		// IDE-1668: emit failure analytics so the "Is Snyk OK?" dashboard's
```
bastiandoetsch: we shouldn't leave ticket ids in the code comments :)

bastiandoetsch (commented):
Re-verification of `ab07a14` — all prior findings verified fixed; CI green; verdict PASS. The inline notes below are optional, non-blocking suggestions for a future pass. — AI review

domain/snyk/scanner/scanner.go
```diff
@@ -282,8 +282,17 @@ func (sc *DelegatingConcurrentScanner) Scan(ctx context.Context, pathToScan type
...
+				// it surfaces on the "Is Snyk OK?" dashboard alongside other scan failures.
+				// processResults handles the scanNotifier.SendError side-effect via sendScanError.
+				processResults(span.Context(), types.ScanData{
```
bastiandoetsch, line 290: [Suggestion] The user-facing `SendError` is now produced as a side-effect of `processResults` rather than called directly here, and `SetScanDone` now runs before it (previously `SendError` was called explicitly, then `SetScanDone`). This is correct because the only production `ScanResultProcessor` is `f.ProcessResults`, which calls `sendScanError` — but that side-effect is now an implicit, undocumented contract on the `types.ScanResultProcessor` func type, so a different processor or test double would silently drop the user-facing error notification.

Fix: document the SendError side-effect on the `ScanResultProcessor` type, or keep an explicit `sc.scanNotifier.SendError(...)` on this path so the notification does not depend on the processor implementation.

— AI review

domain/ide/workspace/folder.go
```diff
@@ -358,6 +359,14 @@ func (f *Folder) scan(ctx context.Context, path types.FilePath) {
...
+		// IDE panel reflects scan errors. Mirror sendScanError's filter — auth-not-set
+		// and "product disabled" outcomes are user state, not failures, and would
+		// inflate the failure rate one event per enabled product per scan.
+		// Reference scans short-circuit inside sendAnalytics via SendAnalytics:false.
+		if !utils.IsNonFailingScanError(scanData.Err.Error()) {
```
bastiandoetsch: [Suggestion] `IsNonFailingScanError` does an exact whole-string lookup on `scanData.Err.Error()`. It is correct today because the producers return these via unwrapped `errors.New(constant)`, but if any scanner later wraps one (e.g. `fmt.Errorf("oss: %w", err)`), the lookup silently misses and a non-failing outcome (auth-not-set, product-disabled) gets counted as a dashboard failure — the exact inflation this gate exists to prevent.

Fix (future): back the filter with typed sentinel errors and match via `errors.Is` instead of string equality.

— AI review

domain/ide/workspace/folder.go
```diff
@@ -358,6 +359,14 @@ func (f *Folder) scan(ctx context.Context, path types.FilePath) {
...
+			go sendAnalytics(ctx, f.engine, f.configResolver, f.logger, &scanData)
```
bastiandoetsch, line 369: [Suggestion] The "what counts as a dashboard failure" policy is now split across three places: this `IsNonFailingScanError` gate, the `isCancellationError` early-return inside `sendAnalytics` (folder.go:496), and the `SendAnalytics`/empty-product guards in `sendAnalytics`. They can drift independently. Cancellations also pass this gate and spawn the goroutine, which does config/category/summary work before bailing at the cancellation check.

Fix: consolidate into one `shouldEmitFailureAnalytics(data) bool` predicate checked here, so the policy lives in one place and cancelled scans skip the goroutine entirely:
```go
if shouldEmitFailureAnalytics(scanData) {
    go sendAnalytics(ctx, f.engine, f.configResolver, f.logger, &scanData)
}
```

— AI review

bastiandoetsch: I'd actually classify that as a strong `Should do`.

domain/snyk/scanner/scanner.go
bastiandoetsch: This is also a change that I would like to have handled - and not by a comment, but by keeping an explicit `sc.scanNotifier.SendError(...)` on this path.

domain/ide/workspace/folder_failure_analytics_test.go
```diff
@@ -0,0 +1,354 @@
...
+
+	"github.com/snyk/snyk-ls/infrastructure/utils"
+	"github.com/snyk/snyk-ls/internal/notification"
+	"github.com/snyk/snyk-ls/internal/product"
+	"github.com/snyk/snyk-ls/internal/testsupport"
+	"github.com/snyk/snyk-ls/internal/testutil"
+	"github.com/snyk/snyk-ls/internal/types"
+)
+
+// IDE-1668: emit failure analytics when a scan returns an error.
```
bastiandoetsch: please remove the jira ticket id

domain/ide/workspace/folder.go
acke: Removed — dropped the IDE-1668 prefix from the comment.

domain/snyk/scanner/scanner.go
acke: Removed — dropped the IDE-1668 prefix.

domain/ide/workspace/folder_failure_analytics_test.go
acke: Removed — dropped the IDE-1668 prefix from the test file header.

domain/snyk/scanner/scanner.go
acke: Done — restored the explicit `sc.scanNotifier.SendError(scanner.Product(), folderPath, err.Error())` call on the pre-scan-command failure path. Added a `UserNotified` bool to `types.ScanData` so `sendScanError` skips its own `scanNotifier.SendError` when the caller has already notified the user; the analytics path through `processResults` is unchanged. Existing call sites pass `UserNotified: false` (default behavior preserved).

domain/ide/workspace/folder.go
acke: Filed as follow-up [IDE-2098](https://snyksec.atlassian.net/browse/IDE-2098). The whole-string lookup is correct today (unwrapped `errors.New(constant)` producers); switching to typed sentinels + `errors.Is` touches the producers in `scannercommon`/`secrets`/`code` and the `ErrorConfig` UI lookups, so it's better as its own change rather than dragging this PR's scope.

domain/ide/workspace/folder.go
acke: Done — added `shouldEmitAnalytics(data *types.ScanData) bool` in `domain/ide/workspace/analytics_errors.go` covering SendAnalytics, empty Product, IsNonFailingScanError, and isCancellationError. Both call sites in `ProcessResults` now gate the goroutine spawn on the predicate, and the duplicate guards inside `sendAnalytics` are removed (function comment now states the contract that callers must gate). Cancelled / non-failing scans skip the goroutine entirely. All existing failure-analytics tests still pass unchanged.

domain/snyk/scanner/scanner.go
```diff
@@ -284,6 +284,16 @@ func (sc *DelegatingConcurrentScanner) Scan(ctx context.Context, pathToScan type
...
 				sc.scanStateAggregator.SetScanDone(folderPath, scanner.Product(), false, err)
+				// Route through processResults purely for failure analytics; UserNotified
+				// suppresses the duplicate SendError that sendScanError would otherwise emit.
+				processResults(span.Context(), types.ScanData{
+					Product:           s.Product(),
+					Err:               err,
+					TimestampFinished: time.Now().UTC(),
+					Path:              folderPath,
+					SendAnalytics:     true,
+					UserNotified:      true,
```
bastiandoetsch, line 297: [Should Fix] `sendScanError` has a second side-effect beyond the notification `SendError`: it calls `f.notifier.SendErrorDiagnostic(f.path, err)` for any error that does not match `IsNonFailingScanError`. Before this PR, pre-scan errors never reached `sendScanError`, so `SendErrorDiagnostic` was never called for them. After this PR, a failed pre-scan shell command sends an unexpected `textDocument/publishDiagnostics` error to the IDE — a red squiggle in the editor that was not present before.

Fix: either add a `sendDiagnostic bool` guard to `sendScanError`, or call `shouldEmitAnalytics` first and then directly invoke the analytics path without routing through `sendScanError`.

— AI review

bastiandoetsch (commented):

domain/snyk/scanner/scanner.go
bastiandoetsch, line 297: [Should Fix] `sendScanError` has a second side-effect beyond the user notification: it calls `f.notifier.SendErrorDiagnostic(f.path, err)` for any error that does not match `IsNonFailingScanError`. Before this PR, pre-scan errors never reached `sendScanError`, so `SendErrorDiagnostic` was never called for them. After this PR, a failed pre-scan shell command sends an unexpected `textDocument/publishDiagnostics` error to the IDE — a red squiggle in the editor that was not present before.

Fix: either add a `sendDiagnostic bool` guard to `sendScanError`, or call `shouldEmitAnalytics` first and invoke the analytics path directly without routing through `sendScanError`.

— AI review

bastiandoetsch (commented):

domain/snyk/scanner/scanner.go
bastiandoetsch, line 297: [Should Fix] `sendScanError` has a second side-effect beyond the user notification: it calls `f.notifier.SendErrorDiagnostic(f.path, err)` for any error that does not match `IsNonFailingScanError`. Before this PR, pre-scan errors never reached `sendScanError`, so `SendErrorDiagnostic` was never called for them. After this PR, a failed pre-scan shell command sends an unexpected `textDocument/publishDiagnostics` error to the IDE — a red squiggle in the editor that was not present before.

Fix: either add a `sendDiagnostic bool` guard to `sendScanError`, or call `shouldEmitAnalytics` first and invoke the analytics path directly without routing through `sendScanError`.

— AI review

domain/ide/workspace/folder_failure_analytics_test.go
```diff
@@ -0,0 +1,352 @@
...
+	captured := testsupport.RequireEventuallyReceive(t, capturedCh, time.Second, 10*time.Millisecond, "failure analytics should still be emitted on partial scan data")
+	require.Len(t, captured.Input, 1)
+	payload := string(captured.Input[0].GetPayload().([]byte))
+
+	assert.Contains(t, payload, `"status":"failure"`)
+	assert.Contains(t, payload, `"error_category":"SNYK-OS"`)
+	assert.Contains(t, payload, `"error_code":"SNYK-OS-7001"`)
+	// Zero TimestampFinished must be defaulted to "now" — never emitted as the
+	// negative-epoch UnixMilli of time.Time{}.
+	assert.NotContains(t, payload, `"timestampMs":-`, "zero TimestampFinished must be defaulted, not emitted as a negative epoch")
```
bastiandoetsch: [Suggestion] `assert.NotContains(t, payload, "\"timestampMs\":-")` searches opaque GAF V2 JSON for a specific key name. If the GAF schema serialises the timestamp under a different key (the value is set via `ic.SetTimestamp(...)`, not written directly), the assertion passes vacuously and provides no protection against a negative-epoch timestamp reaching the backend.

Fix: assert on the value directly before marshalling — e.g. `assert.Positive(t, data.TimestampFinished.UnixMilli())`.

— AI review

domain/snyk/scanner/scanner.go
```diff
@@ -284,6 +284,16 @@ func (sc *DelegatingConcurrentScanner) Scan(ctx context.Context, pathToScan type
 				scanLogger.Err(err).Send()
 				sc.scanNotifier.SendError(scanner.Product(), folderPath, err.Error())
 				sc.scanStateAggregator.SetScanDone(folderPath, scanner.Product(), false, err)
+				// Route through processResults purely for failure analytics; UserNotified
+				// suppresses the duplicate SendError that sendScanError would otherwise emit.
+				processResults(span.Context(), types.ScanData{
```
bastiandoetsch, line 290: [Suggestion] `ScanData.Duration` is not set in this pre-scan error path, so every pre-scan failure event carries `durationMs: 0`. Dashboard consumers cannot distinguish a pre-scan command error from a genuine sub-millisecond scan.

Fix: record the start time before `executePreScanCommand` and set `Duration: time.Since(scanStart)` when building this `ScanData`.

— AI review

domain/snyk/scanner/scanner.go
acke: Done — restructured per option 2: `ProcessResults` now skips `sendScanError` entirely when `UserNotified` is true, keeping `sendScanError`'s single responsibility (both SendError and SendErrorDiagnostic surface to the user; both are caller's job when UserNotified). Pre-PR behavior for pre-scan-command errors is restored — no SendErrorDiagnostic / no red squiggle. Added regression test `Test_processResults_FailedScan_UserNotified_SkipsSendScanError` asserting ErrorCalls=0, SendErrorDiagnosticCount=0, analytics still emitted. Updated `UserNotified` doc to make the broader contract explicit.

acke: Duplicate of #3339347512 — fixed there.

acke: Duplicate of #3339347512 — fixed there.

domain/ide/workspace/folder_failure_analytics_test.go
acke: Done — replaced the brittle string check with JSON-parse + `assert.Positive` on `data.attributes.interaction.timestamp_ms`. The previous `assert.NotContains(payload, "\"timestampMs\":-")` was actually vacuous: the GAF schema serialises the key as snake_case `timestamp_ms`, so the camelCase substring would never match regardless of value.

domain/snyk/scanner/scanner.go
acke: Done — record `preScanStart := time.Now()` before `executePreScanCommand` and set `Duration: time.Since(preScanStart)` on the failure ScanData. Pre-scan failures now carry the actual command runtime so dashboard consumers can distinguish them from sub-millisecond scans.

domain/snyk/scanner/scanner.go
bastiandoetsch: Hm. Isn't sending the diagnostic the better solution? I think I misread this comment a few times, and I would expect an error diagnostic for ANY error that happens.

domain/ide/workspace/folder.go
bastiandoetsch: Fixed at analytics_errors.go:34 — `shouldEmitAnalytics` gates on `IsNonFailingScanError` before spawning the goroutine. No further action needed. — AI review

domain/ide/workspace/folder.go
bastiandoetsch: Fixed — the inverted `IsReferenceScan/SettingScanNetNew` guard is gone; reference scans set `SendAnalytics: false`, which `shouldEmitAnalytics` catches at analytics_errors.go:36. No further action needed. — AI review

domain/ide/workspace/folder.go
bastiandoetsch: Fixed at analytics_errors.go:52 — `isCancellationError` short-circuits `shouldEmitAnalytics` for `context.Canceled` and `context.DeadlineExceeded`, and the new T8 test covers the wrapped case via `errors.Join`. No further action needed. — AI review

domain/ide/workspace/folder.go
bastiandoetsch: Fixed at folder.go:508 — `sendAnalytics` defaults `TimestampFinished` to `time.Now().UTC()` when zero before computing the timestamp. No further action needed. — AI review

domain/ide/workspace/analytics_errors.go
bastiandoetsch: Fixed at analytics_errors.go:66 — `classifyError` runs `errors.As` once and returns both `(category, code)`. No further action needed. — AI review

domain/ide/workspace/analytics_errors.go
bastiandoetsch: Fixed at analytics_errors.go:76 — split now requires `parts[0] != "" && parts[1] != ""` before forming the category prefix, with test coverage for the `"SNYK-"` edge case. No further action needed. — AI review

domain/ide/workspace/folder_failure_analytics_test.go
bastiandoetsch: Acknowledged — partially addressed with `Test_processResults_FailedScan_NonFailingError_NoEmission` and `Test_processResults_FailedScan_Cancellation_NoEmission`. The end-to-end `scanner.Scan` wiring test is intentionally deferred (IDE-2098 or similar follow-up). Finding closed. — AI review

domain/ide/workspace/folder_failure_analytics_test.go
bastiandoetsch: Fixed at testsupport/channels.go:72 — `RequireNeverReceive` replaces the `time.Sleep(50ms)` pattern with a poll-until-timeout approach. No further action needed. — AI review

domain/snyk/scanner/scanner.go
bastiandoetsch: Fixed — `sc.scanNotifier.SendError` is retained at scanner.go:287 and `UserNotified: true` tells `sendScanError` in `ProcessResults` to skip its own `SendError` call, preventing the duplicate. No further action needed. — AI review

domain/ide/workspace/folder.go
bastiandoetsch: Acknowledged — follow-up filed as IDE-2098 to switch producers to typed sentinel errors and `errors.Is` matching. Finding closed; the current whole-string lookup is correct for all existing callers. — AI review

domain/ide/workspace/folder.go
bastiandoetsch: Fixed at analytics_errors.go:34 — `shouldEmitAnalytics` is now the single gate in `ProcessResults` before the goroutine spawn, replacing the scattered inline guards. Duplicate policy points removed. No further action needed. — AI review

bastiandoetsch:
[Should Fix] `stopOnce` is scoped per handler registration inside `withContext`, not per server. Each of the 20+ handler closures registered in `initHandlers` gets its own independent `sync.Once` — so two concurrently panicking handlers can each spawn a `srv.Stop()` goroutine. `jrpc2.Server.Stop()` is safe to call twice but the comment says "one Stop() goroutine per handler registration" when the intent is one per server.

Fix: hoist `stopOnce` out of `withContext` and into `initHandlers`, passing or closing over a single shared instance across all handler closures.

— AI review

bastiandoetsch:
[Should Fix] `NewLspInitializedChannel` in `initializeHandler` overwrites the channel stored in config if `initialize` is received twice (IDE reconnect without server restart). Any goroutine blocked on the old channel via `<-ch` in the notifier callback will block forever — the old channel is never closed and its reference is lost after the overwrite.

Fix: guard creation with a check for an existing open channel (e.g. try a non-blocking receive on the existing channel; if it's already closed, create a new one), or use a `sync.Once`.

— AI review

bastiandoetsch:
[Should Fix] `logError` in `notification.go:35` passes `nil` as the error reporter — an active regression from the DI context-injection refactor. The previous code called `di.ErrorReporter().CaptureError(err)` unconditionally; the new code silently drops error capture for notifier errors even though the reporter is now injectable from context.

Fix: obtain the error reporter via `mustErrorReporterFromContext(ctx)` (or equivalent) in `registerNotifier` and pass it through to `logError`.

— AI review

bastiandoetsch:
[Should Fix] `validateMandatoryDeps` is called on every jrpc2 request but is purely startup validation — the `di.Dependencies` struct is value-captured at `initHandlers` time and never changes after `di.Init` completes. If deps are non-nil at startup, every per-request check always passes; if any are nil, every request always fails. The per-request invocation adds hot-path overhead for zero runtime safety benefit, and the nil-guards in `injectCoreServicesIntoMap` contradict the guarantee the validator just made.

Fix: call `validateMandatoryDeps` once during `Start`, before `srv.Serve()`, and return an error that kills the process. Remove the per-request call and the redundant nil-guards on mandatory deps in the inject functions.

— AI review

bastiandoetsch (approved):
Analytics core (IDE-1668) is correct and ready to ship. All prior review findings addressed. Five Should-Fix issues noted in general comments relate to the DI refactor (IDE-1898, PR #1302 already merged) — not blockers for this PR.

bastiandoetsch:
The four Should-Fix findings from the DI refactor (IDE-1898 / PR #1302) that surfaced during this review have been filed as **[IDE-2102](https://snyksec.atlassian.net/browse/IDE-2102)**:

1. `stopOnce` scoped per handler registration instead of per server
2. `validateMandatoryDeps` called on every request instead of once at startup
3. `NewLspInitializedChannel` goroutine leak on double `initialize` (reconnect)
4. `logError` passes `nil` error reporter in `notification.go:35` (regression from DI migration)

None block this PR — they're in already-merged code from IDE-1898.

— AI review

[IDE-2102]: https://snyksec.atlassian.net/browse/IDE-2102?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1310, opened 2026-05-29
fix(tree-view): prevent tree jump on manual issue selection [IDE-2077]

nick-y-snyk:
### Description

Fixes [IDE-2077](https://snyksec.atlassian.net/browse/IDE-2077).

Clicking an issue in the tree caused the tree to jump unexpectedly. The user clicks a visible row, the IDE sends `snyk.navigateToRange` to the LS, the LS responds with a `showDocument` callback, the IDE calls `window.__selectTreeNode__` — which unconditionally called `scrollIntoView(false)`, scrolling the tree even when the row was already in view.

**Fix:** replace with `scrollIntoView({block:'nearest', inline:'nearest'})` + restore `container.scrollLeft`. `block:nearest` is a no-op when the row is already visible, scrolls minimally when it isn't. `scrollLeft` restore ensures horizontal position never changes.

No IDE plugin changes required — `__selectTreeNode__(issueId)` signature unchanged.

### Checklist

[IDE-2077]: https://snyksec.atlassian.net/browse/IDE-2077?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/tree-view.md
+++ b/docs/tree-view.md
@@ -172,7 +172,7 @@ window.__selectTreeNode__(issueId)
 
 This finds the node by `data-issue-id`, expands any collapsed ancestors so the node is visible, applies the `.selected` class, and removes it from any previously selected node.
 
-**Scroll behavior on programmatic selection:** if the selected row is already fully visible vertically within `#treeContainer`, the scroll position is preserved unchanged. If the row is off-screen, the container scrolls vertically to center the row in the viewport. The horizontal scroll position (`scrollLeft`) is never changed.
+**Scroll behavior on programmatic selection:** the container scrolls the minimum amount needed to make the row visible (`block: 'nearest'`); if the row is already fully visible, no scroll occurs. The horizontal scroll position (`scrollLeft`) is never changed.
 
 ### Issue Badges
 
--- a/docs/tree-view.md
+++ b/docs/tree-view.md
@@ -172,7 +172,7 @@ window.__selectTreeNode__(issueId)
 
 This finds the node by `data-issue-id`, expands any collapsed ancestors so the node is visible, applies the `.selected` class, and removes it from any previously selected node.
 
-**Scroll behavior on programmatic selection:** if the selected row is already fully visible vertically within `#treeContainer`, the scroll position is preserved unchanged. If the row is off-screen, the container scrolls vertically to center the row. The horizontal scroll position (`scrollLeft`) is never changed.
+**Scroll behavior on programmatic selection:** if the selected row is already fully visible vertically within `#treeContainer`, the scroll position is preserved unchanged. If the row is off-screen, the container scrolls vertically to center the row in the viewport. The horizontal scroll position (`scrollLeft`) is never changed.
 
 ### Issue Badges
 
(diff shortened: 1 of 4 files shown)


domain/ide/treeview/template/tree.js
```diff
@@ -51,6 +51,18 @@
...
 
+  function scrollRowIntoViewVerticalOnly(row) {
+    if (!row || !container) return;
+    if (!row.scrollIntoView) return;
+    // Preserve horizontal scroll — scrollIntoView may shift it even with inline:'nearest'
+    // if the row is horizontally out of view. Capture and restore unconditionally.
+    var savedScrollLeft = container.scrollLeft;
+    // block:'nearest' scrolls vertically only when the row is not already visible.
+    // inline:'nearest' minimises horizontal movement, but we restore scrollLeft anyway.
+    row.scrollIntoView({ block: 'nearest', inline: 'nearest' });
```
bastiandoetsch, line 63: Visual Studio has migrated its webview to Chromium, so IE11 is no longer a supported target. The options-object form of scrollIntoView is correct. Updating both docs to reflect the dropped IE11 constraint.

domain/ide/treeview/template/tree.js
```diff
@@ -51,6 +51,18 @@
...
+    container.scrollLeft = savedScrollLeft;
```
bastiandoetsch, line 66: Added a comment at the restore line documenting that scroll-behavior: smooth must not be set on #treeContainer.

domain/ide/treeview/template/tree.js
```diff
@@ -51,6 +51,18 @@
     }
   }
 
+  function scrollRowIntoViewVerticalOnly(row) {
+    if (!row || !container) return;
```
bastiandoetsch: Removed the dead !container check — container is guaranteed non-null by the IIFE guard at line 41.

docs/tree-view.md
```diff
@@ -162,15 +162,17 @@ Disabled products (not in `SupportedIssueTypes`) are rendered with `opacity: 0.5
...
 
 ```javascript
-window.__selectTreeNode__(nodeId)
+window.__selectTreeNode__(issueId)
 ```
 
-This finds the node by `data-node-id`, applies the `.selected` class, and removes it from any previously selected node.
+This finds the node by `data-issue-id`, expands any collapsed ancestors so the node is visible, applies the `.selected` class, and removes it from any previously selected node.
+
+**Scroll behavior on programmatic selection:** if the selected row is already fully visible vertically within `#treeContainer`, the scroll position is preserved unchanged. If the row is off-screen, the container scrolls vertically to center the row. The horizontal scroll position (`scrollLeft`) is never changed.
```
bastiandoetsch: Changed block:'nearest' to block:'center' so programmatic selection always centers the row in the viewport. Updated docs/tree-view.md and the JS test accordingly.

bastiandoetsch (commented):

bastiandoetsch:
[Should Fix] `docs/tree-view.md:264-266` (IE11 Compatibility section) still states that `scrollIntoView(false)` (boolean form) must be used because IE11 (Visual Studio's webview) does not support `ScrollIntoViewOptions`. The new code now uses the options-object form, directly contradicting this documented constraint. This section was not updated in the PR.

Fix: either update the IE11 section to say Visual Studio support is dropped and explain the new browser requirements, or use the feature-detection approach in the code so IE11 still works correctly.

— AI review

domain/ide/treeview/template/tree.js
bastiandoetsch: I do think, that we can drop IE11, as Andrew's VS fix is already merged.

bastiandoetsch (approved):

PR #1311, opened 2026-05-30
refactor(oss): delegate C/C++ auto-detect to CLI extension [IDE-2089]

acke:
## Summary

This PR was originally cut to ship a snyk-ls–side C/C++ auto-detect prompt + per-folder unmanaged toggle. After review, the detection + decision was moved into the CLI's `cli-extension-os-flows` instead. snyk-ls now only sets one env var; everything else is the extension's responsibility.

**Net diff against `main`: 4 files, +62 / -0.**

The intermediate work (detector, prompt, per-folder setting, command, panel toggle, re-arm logic, etc.) was added and then removed in subsequent commits on this branch, so it does not appear in the cumulative file diff. See commit history for the steps.

## What snyk-ls actually changes here

- `infrastructure/oss/cli_scanner.go` — sets `SNYK_AUTODETECT_OSS=1` in the env passed to every OSS CLI invocation. The CLI's os-flows extension then decides per input directory whether to also run an unmanaged scan.
- `infrastructure/oss/cli_scanner_test.go` — new test asserting the env-var contract (`TestCLIScanner_prepareScanCommand_SetsAutodetectEnv`).
- `docs/requirements/architecture.md` — ADR describing the new delegation, including the limitation (see below).
- `docs/requirements/requirements.md` — 3 updated IDE-2089 requirements reflecting the new boundary.

## How the auto-detect now works (CLI side, see [cli-extension-os-flows#251](https://github.com/snyk/cli-extension-os-flows/pull/251))

```
snyk test … (with SNYK_AUTODETECT_OSS=1 from snyk-ls)
   │
   └─► OSWorkflow
         ├─► managed scan (existing path)
         └─► detect.HasCPPArtefacts(inputDir) for each input dir
                  │
                  └─► any C/C++ artefacts?
                          ├─ no  → return managed data
                          └─ yes → invoke legacycli --unmanaged
                                   append its workflow.Data to managed output
```

The extension keeps the previous `--unmanaged` flag's semantics for direct CLI users; the env var is the only new switch.

## Limitation

A native Go `unmanaged.test` workflow does not yet exist; the extension currently invokes the legacy TypeScript CLI to perform the unmanaged scan when C/C++ is detected. The legacy CLI's `workflow.Data` carries its own content type, so for mixed manifest + C/C++ folders the unmanaged and managed results render as **two separate sections** rather than as one unified report. When the native workflow lands, the extension swap is local — no snyk-ls change needed.

## Why this shape

- **No IDE prompt and no per-folder toggle.** Detection happens unconditionally on every OSS scan; the user never has to opt in. The C/C++ walker is bounded (max 5000 entries, depth 6, skipping `node_modules`/`vendor`/`cmake-build-*`/VCS dirs) and short-circuits on the first hit, so the cost is minimal.
- **One source of truth.** Detection rules now live with the rest of the OSS routing in `cli-extension-os-flows/pkg/unmanaged/detect`. Future plugin work (VS Code, Eclipse, Visual Studio) gets the same behaviour automatically since they all invoke the same CLI.
- **Backwards-compatible for non-IDE CLI users.** The auto-detect is off unless `SNYK_AUTODETECT_OSS=1` is set; direct `snyk test` invocations behave exactly as before.

## Depends on / related

- [cli-extension-os-flows#251](https://github.com/snyk/cli-extension-os-flows/pull/251) — adds the detector, `OSWorkflow` orchestration, and the env-var gate. Must merge before this snyk-ls PR has any user-visible effect.
- [snyk-intellij-plugin#834](https://github.com/snyk/snyk-intellij-plugin/pull/834) — closed; the per-folder `snyk_oss_unmanaged_enabled` setting it forwarded no longer exists.

Jira: [IDE-2089](https://snyk.atlassian.net/browse/IDE-2089)

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./internal/types/... ./infrastructure/oss/... ./infrastructure/configuration/...` — green
- [x] `make test-js` — 98/98 pass against regenerated `config-page.html` fixture
- [x] `make build-debug` in `/cli` against this branch — clean (binary built, LS commit pinned)
- [ ] Manual end-to-end: requires cli-extension-os-flows#251 merged + CLI dep bumped. Open a folder with both `package.json` and `.cpp` files in the IDE and confirm both managed and unmanaged scan results appear.

[IDE-2089]: https://snyksec.atlassian.net/browse/IDE-2089?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


infrastructure/oss/cli_scanner.go
```diff
@@ -392,6 +393,14 @@ func (cliScanner *CLIScanner) prepareScanCommand(args []string, parameterBlackli
...
 		args = append(args, params...)
 	}
+	// Opt the CLI's os-flows extension into auto-detecting C/C++ projects:
+	// when set, it runs an extra unmanaged scan alongside the managed scan
+	// for folders that look unmanaged-eligible, so the LS doesn't need to
+	// prompt or expose a per-folder toggle.
+	if env == nil {
+		env = gotenv.Env{}
+	}
+	env["SNYK_AUTODETECT_OSS"] = "1"
```
bastiandoetsch, line 403: [Suggestion] `getCommand` (infrastructure/cli/cli.go) treats a non-nil `env` as the *complete* subprocess environment and skips loading `os.Environ()`. Today that's safe because `updateArgs`→`GetEnvFromSystemAndConfiguration` always returns a fully OS-populated map. But the new `if env == nil { env = gotenv.Env{} }` guard codifies a path that, if ever reached, would hand `getCommand` a map containing only `SNYK_AUTODETECT_OSS=1` — stripping PATH and the rest of the OS env from the CLI subprocess. Consider a one-line comment naming that invariant, or setting the var via the existing `AppendCliEnvironmentVariables` merge path. Non-blocking.

— AI review

bastiandoetsch:
[Suggestion · non-blocking] `SNYK_AUTODETECT_OSS=1` is set unconditionally, but the LS router (`shouldUseLegacyScan`) sends a default `snyk test` (no `UseOsTest`/`UseExperimentalRiskScoreInCLI` flag) through the **legacy** CLI path, whereas the architecture doc's diagram routes this var into the os-flows `OSWorkflow` (the ostest path). Worth confirming whether the legacy CLI also honours the var, or stating explicitly that the feature is gated behind the `UseOsTest` rollout — and reconciling the mermaid diagram so default-population behaviour is clear. The env var degrades gracefully (ignored if unrecognised), so this is a doc/contract clarification, not a defect.

— AI review

bastiandoetsch (approved):
Reviewed with the full verification pass (semantic, adversarial, security, code-review). The snyk-ls side is a correct, minimal change — set `SNYK_AUTODETECT_OSS=1` at the single command-builder choke point so the env reaches both the legacy and ostest paths, with a meaningful test and a genuinely good architecture decision record. No correctness or security issues. Two non-blocking suggestions left inline/below (env-completeness invariant; confirm the var is honoured on the default legacy route). Approving.

— AI review

infrastructure/oss/cli_scanner.go
basti-snyk, line 403: **Suggestion (non-blocking).** The string literal `"SNYK_AUTODETECT_OSS"` + value `"1"` is the entire contract between snyk-ls and `cli-extension-os-flows`, which is the sole arbiter of whether an unmanaged C/C++ scan also runs. That boundary direction is correct (LS only opts in). Two small notes:

- Consider anchoring this literal with a comment pointing at the extension's gate logic, mirroring the existing precedent for the hard-coded workflow ID in `ostest_scan.go`.
- The `if env == nil { env = gotenv.Env{} }` guard just above is effectively dead in production — the SDK env path (`UpdateEnvironmentAndReturnAdditionalParams` → `GetEnvFromSystemAndConfiguration`) always returns a non-nil map; it only fires under test doubles. Harmless, keep or drop.

Verified the feared interactions are non-issues: setting this unconditionally alongside a user-supplied `--unmanaged` does not double-scan (`--unmanaged` forces the legacy path where this var is inert), and forcing `env` non-nil does not strip `os.Environ()`/PATH (env was already non-nil before this change).

— AI review

PR #1314, opened 2026-06-01
fix(treeview): pair hover background and foreground for readability [IDE-2078]

acke:
## Summary

- Tree-node hover text becomes unreadable in IntelliJ light mode (and is fragile in dark) because `.tree-node-row:hover` only sets `background-color` — the foreground stays inherited and clashes with IntelliJ's opaque hover background.
- Fix: introduce `--list-hover-foreground` (sourced from `--vscode-list-hoverForeground`, falling back to `--list-active-selection-foreground`) and apply it alongside the existing hover background. Adds a 1px `outline` using `--focus-border` to satisfy the "border" AC without shifting layout.
- Removed redundant `.tree-node-issue .tree-node-row:hover` / `.tree-node-location .tree-node-row:hover` rules — they only repeated the base hover background, which still matches via the base selector.

Ticket: [IDE-2078](https://snyksec.atlassian.net/browse/IDE-2078)

IntelliJ
image
image

Visual Studio Code
image

## Scope

**In scope:** `domain/ide/treeview/template/styles.css` (CSS-only). One Go test added in `domain/ide/treeview/tree_html_test.go`. 4 atomic requirements appended to `docs/requirements/requirements.md`.

**Out of scope (intentionally untouched):** `.filter-btn:hover`, `.action-btn:hover`, `.ref-picker-*:hover`, `.error-overlay-close:hover`, and the config-page CSS — different hover sites, not what the ticket reports.

## Test plan

- [x] `go test ./domain/ide/treeview/...` — 28 pass (new `TestTreeHtmlRenderer_TreeHoverRow_PairsBackgroundAndForeground` included)
- [ ] Manual IDE matrix (still pending — required before un-drafting):
  - [ ] IntelliJ light (IntelliJ Light + macOS Light)
  - [ ] IntelliJ dark (Darcula + High Contrast)
  - [ ] VS Code light + dark
  - [ ] Eclipse default
- [ ] Before/after screenshots from IntelliJ light to be attached

## Rollback

Trivial. The change is additive (one new CSS custom property + a `color` line on the existing `:hover` block). `git revert <sha>` is safe — no schema, protocol, or migration.

[IDE-2078]: https://snyksec.atlassian.net/browse/IDE-2078?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

___

### **PR Type**
Bug fix, Documentation

___

### **Description**
- Fixes tree-view hover readability by pairing colors.

- Introduces theme variables for hover foreground.

- Documents hover styling requirements for IDEs.

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["CSS Theme Variables"] -- "define hover foreground" --> B["Treeview CSS Rules"]
  B -- "apply background & color" --> C["Treeview UI Readability"]
  D["Documentation"] -- "capture requirements" --> B
```

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>styles.css</strong><dd><code>Enhance treeview hover color pairing with theme variables</code></dd></summary>
<hr>

domain/ide/treeview/template/styles.css

<ul><li>Defines <code>--list-hover-foreground</code> variable sourced from theme variables.<br> <li> Applies the new foreground color to the <code>.tree-node-row:hover</code> state.<br> <li> Removes redundant specific hover rules for issue and location rows.<br> <li> Ensures <code>.tree-node-row-info:hover</code> has inherited color.</ul>

</details>

  </td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1314/changes#diff-60771d9f4d356e11613ba1f9e235b885aa5f0e120e46f8c2b10e6497c6429b6a">+4/-9</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
</table></td></tr><tr><td><strong>Documentation</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>requirements.md</strong><dd><code>Document tree-view hover readability requirements</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

docs/requirements/requirements.md

<ul><li>Appends four new requirements related to tree-view hover styling.<br> <li> Requirements cover theme variable usage, fallbacks, and IDE injection.</ul>

</details>

  </td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1314/changes#diff-3d7fa82db2df9bfc78cb214da6015ec71b61922b3c7dfa9aedd58489833216ed">+4/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

</details>

___

### DIFF
--- a/domain/ide/treeview/template/styles.css
+++ b/domain/ide/treeview/template/styles.css
@@ -19,6 +19,7 @@
   --info-background: var(--vscode-inputValidation-infoBackground, #e6f3ff);
 
   --list-hover-background: var(--vscode-list-hoverBackground, #8080801a);
+  --list-hover-foreground: var(--vscode-list-hoverForeground, var(--list-active-selection-foreground));
   --list-active-selection-background: var(--vscode-list-activeSelectionBackground, #0066cc1f);
   --list-active-selection-foreground: var(--vscode-list-activeSelectionForeground, #0066cc);
 
@@ -77,6 +78,9 @@ body {
 
 .tree-node-row:hover {
   background-color: var(--list-hover-background);
+  color: var(--list-hover-foreground);
+  outline: 1px solid var(--focus-border);
+  outline-offset: -1px;
 }
 
 .tree-node-row.selected {
@@ -278,10 +282,6 @@ body {
   cursor: pointer;
 }
 
-.tree-node-issue .tree-node-row:hover {
-  background-color: var(--list-hover-background);
-}
-
 /* Issue group node (multi-location container) — expand/collapse like file node */
 .tree-node-issue-group .tree-node-row {
   cursor: pointer;
@@ -290,10 +290,6 @@ body {
 /* Location node — clickable like issue leaf */
 .tree-node-location .tree-node-row {
   cursor: pointer;
-}
-
-.tree-node-location .tree-node-row:hover {
-  background-color: var(--list-hover-background);
 }
 
 /* Filter toolbar */
--- a/domain/ide/treeview/template/styles.css
+++ b/domain/ide/treeview/template/styles.css
@@ -21,7 +21,7 @@
   --list-hover-background: var(--vscode-list-hoverBackground, #8080801a);
   --list-hover-foreground: var(--vscode-list-hoverForeground, var(--list-active-selection-foreground));
   --list-active-selection-background: var(--vscode-list-activeSelectionBackground, #0066cc1f);
-  --list-active-selection-foreground: var(--vscode-list-activeSelectionForeground, #0066cc);
+  --list-active-selection-foreground: var(--vscode-list-activeSelectionForeground, inherit);
 
   --tree-indent-guides-stroke: var(--vscode-tree-indentGuidesStroke, #80808099);
 
@@ -79,8 +79,6 @@ body {
 .tree-node-row:hover {
   background-color: var(--list-hover-background);
   color: var(--list-hover-foreground);
-  outline: 1px solid var(--focus-border);
-  outline-offset: -1px;
 }
 
 .tree-node-row.selected {
@@ -410,6 +408,7 @@ body {
 
 .tree-node-row-info:hover {
   background-color: transparent;
+  color: inherit;
 }
 
 /* Disabled product nodes — greyed out */
--- a/domain/ide/treeview/tree_html_test.go
+++ b/domain/ide/treeview/tree_html_test.go
@@ -709,7 +709,7 @@ func TestTreeHtmlRenderer_TreeHoverRow_PairsBackgroundAndForeground(t *testing.T
 	html := renderer.RenderTreeView(TreeViewData{})
 
 	// The embedded CSS must define the hover-foreground variable so that hover
-	// text colour is paired with the hover background across IDE themes (IDE-2078).
+	// text color is paired with the hover background across IDE themes (IDE-2078).
 	assert.Contains(t, html, "--list-hover-foreground", "CSS must define --list-hover-foreground variable")
 
 	// The .tree-node-row:hover block must set BOTH background-color AND color so
--- a/domain/ide/treeview/tree_html_test.go
+++ b/domain/ide/treeview/tree_html_test.go
@@ -700,27 +700,3 @@ func TestTreeHtmlRenderer_FileNode_EmptyFileIconHTML_RendersGenericSVG(t *testin
 	assert.NotContains(t, html, "📄", "empty FileIconHTML should not fall back to emoji")
 	assert.Contains(t, html, `<svg`, "empty FileIconHTML should render the generic file SVG")
 }
-
-func TestTreeHtmlRenderer_TreeHoverRow_PairsBackgroundAndForeground(t *testing.T) {
-	engine := testutil.UnitTest(t)
-	renderer, err := NewTreeHtmlRenderer(engine.GetLogger())
-	require.NoError(t, err)
-
-	html := renderer.RenderTreeView(TreeViewData{})
-
-	// The embedded CSS must define the hover-foreground variable so that hover
-	// text color is paired with the hover background across IDE themes (IDE-2078).
-	assert.Contains(t, html, "--list-hover-foreground", "CSS must define --list-hover-foreground variable")
-
-	// The .tree-node-row:hover block must set BOTH background-color AND color so
-	// that an opaque IDE-injected hover background does not leave text unreadable.
-	hoverBlockStart := strings.Index(html, ".tree-node-row:hover")
-	require.Greater(t, hoverBlockStart, 0, ".tree-node-row:hover rule must be present")
-	hoverBlock := html[hoverBlockStart:]
-	closingBrace := strings.Index(hoverBlock, "}")
-	require.Greater(t, closingBrace, 0, ".tree-node-row:hover block must have a closing brace")
-	hoverRule := hoverBlock[:closingBrace]
-
-	assert.Contains(t, hoverRule, "background-color", ".tree-node-row:hover must set background-color")
-	assert.Contains(t, hoverRule, "color", ".tree-node-row:hover must set color")
-}
--- a/domain/ide/treeview/tree_html_test.go
+++ b/domain/ide/treeview/tree_html_test.go
@@ -700,3 +700,27 @@ func TestTreeHtmlRenderer_FileNode_EmptyFileIconHTML_RendersGenericSVG(t *testin
 	assert.NotContains(t, html, "📄", "empty FileIconHTML should not fall back to emoji")
 	assert.Contains(t, html, `<svg`, "empty FileIconHTML should render the generic file SVG")
 }
+
+func TestTreeHtmlRenderer_TreeHoverRow_PairsBackgroundAndForeground(t *testing.T) {
+	engine := testutil.UnitTest(t)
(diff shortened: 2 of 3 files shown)


domain/ide/treeview/tree_html_test.go
rrama, line 1: Seems like a useless test tbh.

rrama (approved):
I don't think we really need the test, but other than that LGTM.

domain/ide/treeview/tree_html_test.go
```diff
@@ -700,3 +700,27 @@ func TestTreeHtmlRenderer_FileNode_EmptyFileIconHTML_RendersGenericSVG(t *testin
...
+	// that an opaque IDE-injected hover background does not leave text unreadable.
+	hoverBlockStart := strings.Index(html, ".tree-node-row:hover")
+	require.Greater(t, hoverBlockStart, 0, ".tree-node-row:hover rule must be present")
+	hoverBlock := html[hoverBlockStart:]
+	closingBrace := strings.Index(hoverBlock, "}")
+	require.Greater(t, closingBrace, 0, ".tree-node-row:hover block must have a closing brace")
+	hoverRule := hoverBlock[:closingBrace]
+
+	assert.Contains(t, hoverRule, "background-color", ".tree-node-row:hover must set background-color")
+	assert.Contains(t, hoverRule, "color", ".tree-node-row:hover must set color")
```
bastiandoetsch: [Should Fix] `assert.Contains(t, hoverRule, "color", ...)` does not verify that `color:` is set. The string `"color"` is a substring of `"background-color"`, which is already in the same block, so removing the new `color: var(--list-hover-foreground)` line from the CSS leaves the test green. The assertion gives a false guarantee for the core correctness of this fix.

Fix: assert on the full property value instead.
```go
assert.Contains(t, hoverRule, "color: var(--list-hover-foreground)", ".tree-node-row:hover must set color")
```

— AI review

domain/ide/treeview/template/styles.css
```diff
@@ -77,6 +78,9 @@ body {
 
 .tree-node-row:hover {
   background-color: var(--list-hover-background);
+  color: var(--list-hover-foreground);
```
bastiandoetsch, line 81: [Suggestion] `.tree-node-row-info` elements carry both classes, so `.tree-node-row:hover` now also changes their text color on hover, even though the `.tree-node-row-info:hover` rule already sets `background-color: transparent` to suppress the hover background and signal they are non-interactive. The result is that info rows get a foreground color shift without a background shift, which may look odd.

If info rows should be fully inert on hover, add `color: inherit` to the existing `.tree-node-row-info:hover` rule to cancel the color change.

— AI review

domain/ide/treeview/template/styles.css
```diff
@@ -77,6 +78,9 @@ body {
...
+  outline: 1px solid var(--focus-border);
```
bastiandoetsch: [Suggestion] `--focus-border` is the VS Code design-system variable for keyboard focus rings, not for pointer hover. Applying it here conflates two distinct interaction states; a screen-reader or accessibility audit tool may flag it. `box-shadow` achieves the same inset-border effect without the semantic mismatch and without needing `outline-offset`:
```css
box-shadow: inset 0 0 0 1px var(--focus-border);
```
Alternatively, if the outline is not in the stated requirements (requirements.md lists none), consider dropping it entirely.

— AI review

domain/ide/treeview/template/styles.css
```diff
@@ -19,6 +19,7 @@
   --info-background: var(--vscode-inputValidation-infoBackground, #e6f3ff);
 
   --list-hover-background: var(--vscode-list-hoverBackground, #8080801a);
+  --list-hover-foreground: var(--vscode-list-hoverForeground, var(--list-active-selection-foreground));
```
bastiandoetsch, line 22: [Suggestion] The fallback chain for `--list-hover-foreground` is: `--vscode-list-hoverForeground` → `--list-active-selection-foreground` → `#0066cc`. In a plain browser context with no IDE injection at all (useful for local HTML development and snapshot testing), hover text will render in the blue selection-accent color rather than the body text color, which is unlikely to be the intent.

If `inherit` were used as the final fallback for `--list-active-selection-foreground`, hover text would naturally adopt the ancestor text color and remain readable in any context.

— AI review

bastiandoetsch (approved):
Looks good — the core fix is correct, fallback chain is well-formed, removed rules are genuinely redundant, and CI is green.

Inline suggestions are posted on the diff. Worth picking up before merge:

1. **[Should Fix] Test false positive** () — `assert.Contains(t, hoverRule, "color", ...)` passes even if the `color:` property is removed, because `"color"` is a substring of `"background-color"`. Change the assertion to `assert.Contains(t, hoverRule, "color: var(--list-hover-foreground)", ...)` to make it load-bearing.

2. **[Suggestion] Info-row hover color** () — `.tree-node-row-info` elements now also receive the `color` change on hover. If they should remain fully inert, add `color: inherit` to `.tree-node-row-info:hover`.

3. **[Suggestion] `--focus-border` semantic** () — `--focus-border` is the VS Code focus-ring token; using it on pointer hover conflates focus and hover states. Consider `box-shadow: inset 0 0 0 1px var(--focus-border)` or dropping the outline altogether (it's not in the stated requirements).

4. **[Suggestion] Fallback bottoms out at blue** () — In a standalone browser with no IDE injection the fallback chain resolves to `#0066cc`. Using `inherit` as the ultimate fallback for `--list-active-selection-foreground` would give neutral behavior in that context.

bastiandoetsch (approved):
Looks good — the core fix is correct, fallback chain is well-formed, removed rules are genuinely redundant, and CI is green.

Inline suggestions are posted on the diff. Worth picking up before merge:

1. **[Should Fix] Test false positive** (`tree_html_test.go:725`) — `assert.Contains(t, hoverRule, "color", ...)` passes even if the `color:` property is removed, because `"color"` is a substring of `"background-color"`. Change the assertion to `assert.Contains(t, hoverRule, "color: var(--list-hover-foreground)", ...)` to make it load-bearing.

2. **[Suggestion] Info-row hover color** (`styles.css:81`) — `.tree-node-row-info` elements now also receive the `color` change on hover. If they should remain fully inert, add `color: inherit` to `.tree-node-row-info:hover`.

3. **[Suggestion] `--focus-border` semantic** (`styles.css:82`) — `--focus-border` is the VS Code focus-ring token; using it on pointer hover conflates focus and hover states. Consider `box-shadow: inset 0 0 0 1px var(--focus-border)` or dropping the outline altogether (it's not in the stated requirements).

4. **[Suggestion] Fallback bottoms out at blue** (`styles.css:22`) — In a standalone browser with no IDE injection the fallback chain resolves to `#0066cc`. Using `inherit` as the ultimate fallback for `--list-active-selection-foreground` would give neutral behavior in that context.

domain/ide/treeview/tree_html_test.go
acke: Removed the test entirely — @rrama also questioned its value and your "color substring of background-color" finding showed the assertion was not load-bearing. With the test gone the CSS change stands on its own.

domain/ide/treeview/tree_html_test.go
acke: Agreed — removed the test.

acke:
Re: https://github.com/snyk/snyk-ls/pull/1314#pullrequestreview-4416805718

Removed the test — thanks for the review.

domain/ide/treeview/template/styles.css
acke: Good catch — added `color: inherit` to `.tree-node-row-info:hover` so the foreground shift is neutralised alongside the background.

domain/ide/treeview/template/styles.css
acke: Dropped the outline entirely — it was not in the stated requirements, and the background + foreground pair already provides the hover signal. Avoids the focus-vs-hover token conflation.

domain/ide/treeview/template/styles.css
acke: Changed `--list-active-selection-foreground` fallback from `#0066cc` to `inherit`, so hover (and selected) text inherits the body colour in a plain browser. Also better aligns with the IDE-2078 requirement of "no hard-coded colour values; sensible neutral fallbacks".

acke:
Re: https://github.com/snyk/snyk-ls/pull/1314#pullrequestreview-4416888211 / https://github.com/snyk/snyk-ls/pull/1314#pullrequestreview-4416889324

Thanks for the thorough review. All four points addressed in commits `06eae161` and `a1830ef7`:

1. **Test false positive** — deleted the test entirely (also addressed @rrama's "useless test" feedback).
2. **Info-row hover color** — added `color: inherit` to `.tree-node-row-info:hover`.
3. **`--focus-border` semantic mismatch** — dropped the outline entirely; not in the stated requirements.
4. **Fallback bottoms out at blue** — changed `--list-active-selection-foreground` fallback from `#0066cc` to `inherit`, also better aligns with the IDE-2078 "no hard-coded colour values" requirement.

PR #1316, opened 2026-06-01
feat: update html tree view with UI changes for ga [IDE-2117]

andrewrobinsonhodges-snyk:
### Description

Brings the HTML tree view to GA with a batch of UI/UX fixes, and adds per-folder filters

- UI polish: updated tooltips and child-node text to match the prototype; tooltips now explain why scans aren't running.
- Chevrons: hidden when a node has no children; restored on file icons where they'd gone missing.
- Autofix: hide autofix messaging when autofix isn't available.
- Severity filters are now per-folder, and sync with the settings page in real-time.
- Toolbar renders a "mixed" (translucent) state when open folders disagree on a severity filter (MixedSeverity).

### Checklist


rrama (commented):
I haven't properly checked this, just wanted to share the bugs the AI thinks it found.

domain/ide/treeview/tree_builder.go
```diff
@@ -116,6 +121,18 @@ func (b *TreeBuilder) BuildTree(workspace types.Workspace) TreeViewData {
...
+			// Agent Fix is gated per-folder here (each folder's Code node reflects its
+			// own SAST settings). This intentionally differs from the summary panel,
+			// which is workspace-wide and uses any-folder enablement
+			// (scanstates.HtmlRenderer.isAutofixEnabledInAnyFolder); in a multi-root
+			// workspace with mixed settings the two surfaces can legitimately disagree.
+			// Absent/nil SAST settings deliberately fall through to AgentFixEnabled=false
+			// (unknown == hidden). This is safe: the fixable line only renders after a
+			// completed Code scan (see buildProductNodes' enabled/scanRegistered gate),
+			// which cannot succeed without SAST settings being populated first.
+			if sast := types.GetSastSettings(cfg.Conf(), cfg.FolderPath); sast != nil {
```
rrama: AI found bug (unchecked by me):

> **Missing nil check on cfg.Conf() before passing to GetSastSettings, risking nil pointer panic**
> 
> The new SAST settings code at `domain/ide/treeview/tree_builder.go:133` calls `cfg.Conf()` and passes the result directly to `types.GetSastSettings()` without checking for nil. The established pattern in the same `if cfg != nil` block — visible just 6 lines below at `domain/ide/treeview/tree_builder.go:137-138` — explicitly stores `cfg.Conf()` in a local variable and guards it with `if conf != nil` before use. If `cfg.Conf()` returns nil (which the existing guard proves is a possible state), `GetSastSettings` will dereference the nil `configuration.Configuration`, panicking and crashing the `renderLoop` goroutine in `TreeScanStateEmitter`. Once that goroutine dies, no further tree views are rendered for the rest of the session.

domain/ide/treeview/template/styles.css
```diff
@@ -411,11 +473,130 @@ body {
...
+  white-space: nowrap;
+}
+.tree-node-info--untrusted-folder .untrusted-learn-more:hover {
+  text-decoration: underline;
+}
+.tree-node-info--untrusted-folder .untrusted-folder-list-row {
+  flex-direction: column;
+  align-items: flex-start;
+  padding-top: 4px;
+  gap: 2px;
```
rrama, line 545: AI found bug (unchecked by me):

> **Missing `display: flex` makes flex properties inert on `.untrusted-folder-list-row`**
> 
> The CSS rule at `domain/ide/treeview/template/styles.css:504-509` sets `flex-direction: column`, `align-items: flex-start`, and `gap: 2px` on `.untrusted-folder-list-row`. However, this element inherits `display: block` from the base `.tree-node-row` rule (line 70) and no override sets `display: flex`. The flex properties are completely inert on block-level elements, so the intended column layout with spacing won't apply when the untrusted-folder feature is wired up.

domain/ide/treeview/template/tree.js
```diff
@@ -585,4 +597,238 @@
...
+
+  function measureWidth(text) {
+    return getMeasureCtx().measureText(text).width;
+  }
+
+  function truncateMiddleByWidth(text, maxWidth) {
+    if (maxWidth <= 0) return text;
+    if (measureWidth(text) <= maxWidth) return text;
+
+    var lastSlash = text.lastIndexOf('/');
```
rrama, line 658: AI found bug (unchecked by me):

> **Middle truncation ineffective on Windows due to hardcoded `/` separator**
> 
> The `truncateMiddleByWidth` function only searches for `/` via `text.lastIndexOf('/')` (line 643). On Windows, `computeRelativePath` (`domain/ide/treeview/tree_builder.go:741`) uses `filepath.Rel` which returns paths with `\` separators (e.g. `src\app\component.ts`). Since `lastIndexOf('/')` returns -1 for these paths, the function always returns the unmodified text and falls back to CSS end-truncation — which clips the filename (the most informative part). The middle-truncation feature is entirely non-functional on Windows.

application/server/configuration.go
```diff
@@ -717,6 +717,12 @@ func applySeverityFilter(conf configuration.Configuration, engine workflow.Engin
 		return
 	}
 
+	// This is the user-global settings path: it sets the workspace default only.
+	// Per-folder severity overrides are written independently via the per-folder
+	// folderConfigs channel (processFolderConfigs) for the specific folder being
+	// edited, and via the workspace-wide tree toggle (toggleTreeFilter). Keeping
+	// this path global-only is what stops the global default and per-folder
+	// overrides from moving in lockstep. (IDE-1996)
```
rrama: Nitpick: Such an AI comment...

domain/ide/treeview/tree_scan_emitter.go
```diff
@@ -124,14 +124,73 @@ func (e *TreeScanStateEmitter) renderPending() {
...
+			if firstFC == nil {
+				firstFC = fc
+			}
+			severities = append(severities, fc.ConfigResolver.FilterSeverityForFolder(fc))
+		}
+		if len(severities) > 0 {
+			severity, mixed = aggregateSeverityFilters(severities)
+		}
+		if firstFC != nil {
+			issueView = firstFC.ConfigResolver.IssueViewOptionsForFolder(firstFC)
```
rrama, line 166: AI nitpick:

> **IssueViewOptions in filterState only uses first folder, not aggregated across all folders like severity**
> 
> The `filterState` method at `domain/ide/treeview/tree_scan_emitter.go:161-162` reads `IssueViewOptions` from only the first folder's config (`firstFC`), while severity filters are aggregated across ALL folders with mixed-state detection (lines 146-160). In a multi-folder workspace with different IVO per folder, the toolbar's IVO state only reflects the first folder. This asymmetry is currently harmless because IVO filter buttons aren't rendered in the tree toolbar template (`domain/ide/treeview/template/tree.html` only has severity buttons at lines 33-36). However, if IVO buttons are added to the toolbar in the future, this will silently use only the first folder's state. Consider adding a `MixedIssueViewOptions` analogous to `MixedSeverity`, or at least adding a comment noting this gap.

A comment will keep it happy.

rrama (commented):
Leaving the AI comments before I review once you merge in Basti's changes.

domain/ide/command/toggle_tree_filter_test.go
rrama, line 1: AI found testing gap (unverified by me):

> notifyConfigurationChanged is entirely untested. Every test constructs toggleTreeFilter with only command and engine set — notifier and configResolver are always nil. So notifyConfigurationChanged (toggle_tree_filter.go:155-161) always short-circuits at the cmd.notifier == nil guard. The $/snyk.configuration push — a headline feature of this PR (IDE-1866) — has no test verifying the notification is actually sent or that the payload contains the right filter values. A test with a MockNotifier asserting Send was called would close this.

domain/ide/command/toggle_tree_filter_test.go
```diff
@@ -19,18 +19,75 @@ package command
...
+	scanNotifier := scanner.NewMockScanNotifier()
+	scanPersister := persistence.NewGitPersistenceProvider(engine.GetLogger(), engine.GetConfiguration())
+	scanStateAggregator := scanstates.NewNoopStateAggregator()
+	resolver := testutil.DefaultConfigResolver(engine)
+	folderPath := types.PathKey("dummy")
+	w := workspace.New(engine.GetConfiguration(), engine.GetLogger(), performance.NewInstrumentor(), sc, nil, scanNotifier, notification.NewMockNotifier(), scanPersister, scanStateAggregator, featureflag.NewFakeService(), resolver, engine)
+	folder := workspace.NewFolder(engine.GetConfiguration(), engine.GetLogger(), folderPath, "dummy", sc, nil, scanNotifier, notification.NewMockNotifier(), scanPersister, scanStateAggregator, featureflag.NewFakeService(), resolver, engine)
+	w.AddFolder(folder)
+	config.SetWorkspace(engine.GetConfiguration(), w)
+	return folderPath
```
rrama, line 46: Could just be:
```go
func setupToggleWorkspaceFolder(t *testing.T, engine workflow.Engine) types.FilePath {
	t.Helper()
	folderPath := types.PathKey("dummy")
	setupToggleWorkspaceFolders(t, engine, folderPath)
	return folderPath
}
```

domain/ide/treeview/template/tree.js
```diff
@@ -538,6 +538,18 @@
       var isActive = hasClass(btn, 'filter-active');
       var enabled = !isActive;
 
+      // Optimistically reflect the toggle in the button's class so the active
+      // state updates immediately, rather than waiting for the LS to re-render
+      // the tree in response to the command below. A "mixed" button (open folders
+      // disagree) counts as not-active, so the first click enables the severity
+      // for every folder, resolving the mismatch (IDE-1866).
```
rrama: Q: What if one folder is org locked? Would optimistic not be too optimistic? Just a comment update to address this case would do me.

AI put my thoughts into words (although I have not read it):

> Nit (org-locked edge case): the optimistic flip can be wrong when a folder has this severity org-locked.
> 
> The comment says "the first click enables the severity for every folder, resolving the mismatch" — but that isn't true when one of the open folders has the severity locked by an org admin. writeFilterToAllFolders (toggle_tree_filter.go) writes a UserFolderKey value to every folder, but per the folder-scope precedence (Locked Remote > User Folder Override > …) the locked folder's resolved value doesn't change — see TestConfigResolver_GetValue_OrgScope_Locked (config_resolver_test.go:412-418).
> 
> So if the locked value disagrees with the toggled value, the button flashes solid filter-active here, then filterState/aggregateSeverityFilters (tree_scan_emitter.go) re-marks it filter-mixed on the next LS re-render — a visible flicker that snaps back to mixed.
> 
> The resolved state ends up correct, so this is a UX nit rather than a correctness bug. If we want to avoid it, the toolbar would need a per-severity "locked in some folder" signal (not currently present in MixedSeverity / TreeViewFilterState in tree_node.go) so we can skip the optimistic flip there. Otherwise, worth a code comment noting the optimistic path doesn't account for locked folders.

domain/ide/treeview/template/tree.js
```diff
@@ -585,4 +597,238 @@
...
+    while (lo <= hi) {
+      var mid = (lo + hi) >> 1;
+      if (measureWidth(prefix.substring(0, mid)) <= availForPrefix) {
+        best = mid;
+        lo = mid + 1;
+      } else {
+        hi = mid - 1;
+      }
+    }
+    if (best === prefix.length) return text;
```
rrama, line 681: Very minor edge case (non-blocking): for a single-char folder prefix like `x/file.txt`, when `x` doesn't fit the binary search yields `best === 0`, so we return `…/file.txt`. That swaps a 1-char folder name for a 1-char ellipsis — no width saved, and the string is still too long, so we've dropped the `x` for nothing.

The loop only checks whether the kept prefix fits `availForPrefix`; it never checks whether truncating actually made the result narrower. Whenever the dropped portion is about as wide as `…` (short prefixes like `x/`), we lose info without gaining fit.

domain/ide/treeview/template/tree.js
```diff
@@ -585,4 +597,238 @@
...
+
+  if (typeof ResizeObserver !== 'undefined') {
+    new ResizeObserver(scheduleRetruncate).observe(container);
+  } else {
+    window.addEventListener('resize', scheduleRetruncate);
+  }
+
+  // Expand/collapse mutates .expanded on .tree-node elements, which changes
+  // which file rows are laid out. Watch class changes on tree-node elements
+  // and re-run; selection toggling on .tree-node-row is filtered out below.
```
rrama, line 739: Consider adding:

> // No fallback when MutationObserver is unavailable: unlike ResizeObserver
  // (which falls back to the 'resize' event), there is no DOM event for
  // "a node's class changed", so expand/collapse re-truncation is simply
  // skipped.

domain/ide/treeview/template/tree.js
```diff
@@ -585,4 +597,238 @@
...
+
+  scheduleRetruncate();
+
+  // Tooltips.
+  // Native `title` tooltips render inconsistently inside IDE webviews, so we
+  // draw our own from the same `title` attributes (IDE-1864 disabled/errored
+  // scanner hints, full file paths from the truncation pass above, untrusted
+  // folder paths). While our tooltip is shown the title is moved to
+  // data-tooltip so the native one can't also fire, and restored on mouse-out
+  // so it stays available for accessibility and any non-webview host.
```
rrama: Nitpick: Doing our own tooltips in 2026 is mental. I know Bootstrap is outdated and should be replaced for the HTML settings page, but I would have preferred it. Oh well, we can change it later.

domain/ide/treeview/tree_builder.go
```diff
@@ -444,19 +533,15 @@ func productDescription(p product.Product, totalIssues int, counts *SeverityCoun
...
-func productCountWord(p product.Product, count int) string {
-	if p == product.ProductOpenSource {
-		if count == 1 {
-			return "unique vulnerability"
-		}
-		return "unique vulnerabilities"
-	}
+// countWord returns the singular/plural noun for an issue count. All products
+// use "issue"/"issues" for consistency across scanner rows (OSS previously said
+// "unique vulnerabilities").
```
rrama: This sort of comment explaining the change (and why) should be in the PR description and Jira ticket, not really needed in the code.

infrastructure/configuration/template/js/features/filter-sync.js
rrama, line 1: I've assumed this and related changes go away with Basti's HTML refresh code.

domain/ide/treeview/tree_builder_test.go
rrama, line 1: AI found test coverage gap (unverified by me):

> showFixableLine for Code when agentFixEnabled=true but fixableCount=0. TestBuildTree_Code_AgentFixEnabled_ShowsFixableLine (:660) uses a fixable issue. There's no test for Code + agentFix-enabled + zero fixable issues, which would hit the "There are no issues automatically fixable." branch at tree_builder.go:420. The OSS equivalent is covered (TestBuildTree_ProductNode_ZeroFixable_ShowsNoFixableMessage at :627), but the Code path is new behavior in this PR and goes untested.

domain/ide/treeview/tree_node.go
```diff
@@ -45,38 +45,63 @@ type SeverityCounts struct {
...
+	// errored scanners hint that the row is clickable for details.
+	Tooltip string `json:"tooltip,omitempty"`
+	// InfoVariant selects an alternate rendering for info nodes. "" is the plain
+	// inline info row; "filter-empty" (IDE-1869) is hidden until the client
+	// reveals it when severity filters hide every issue; "untrusted-folder"
+	// (IDE-1882) is the workspace-trust banner with a folder list + Trust button.
+	InfoVariant string `json:"infoVariant,omitempty"`
+	// FolderPaths lists the untrusted folder paths shown in the "untrusted-folder"
+	// info banner (IDE-1882).
+	FolderPaths         []string `json:"folderPaths,omitempty"`
```
rrama: Nit: Bad field name.

domain/ide/treeview/tree_node.go
rrama, line 1: AI found test coverage gaps:

> 1. MixedSeverity toolbar rendering — the headline IDE-1866 template change (...tree.html:33-36: filter-mixed class + the "Open folders use different ... severity filters" tooltip) has no test here. The aggregation logic is tested in ...tree_scan_emitter_test.go (TestAggregateSeverityFilters), but nothing asserts the template actually emits filter-mixed or the mixed tooltip when FilterState.MixedSeverity.X is true. RenderTreeView accepts a TreeViewFilterState directly, so this would be straightforward to test.
> 
> 2. Scanner Tooltip → title attribute (...tree.html:70) — the IDE-1864 tooltip is rendered by this template but not asserted here. (The content of tooltips is tested via BuildTree in ...tree_builder_test.go:TestBuildTree_DisabledScanner_TooltipVariesByReason, but the template wiring of .Tooltip into title= is untested.)
> 
> 3. data-product-id attribute (...tree.html:69) — new attribute, no assertion.
> 
> 4. product-icon--abbreviated class (...tree.html:72) — new icon class, no assertion.
> 
> 5. InfoVariant blocks (...tree.html:147-169) — neither "untrusted-folder" (banner, folder paths, Learn-more link, Trust button) nor "filter-empty" (the hidden attribute) is rendered by any test. As the lifeguard analyses note, the builder never sets InfoVariant yet, so these are dead template paths — but they're shipping renderable code with zero coverage. A direct RenderTreeView test (which doesn't depend on the builder) could cover them now.

NGL, I didn't really understand 3 or 4 and just copied them in. For 5, I am hoping this will be done on the follow-up PR for actual untrusted support.

domain/ide/treeview/tree_scan_emitter_test.go
rrama, line 1: AI found coverage gaps:

> 1. filterState mixed-severity path is never tested end-to-end. TestAggregateSeverityFilters (:212) unit-tests aggregateSeverityFilters directly, but no test exercises filterState (tree_scan_emitter.go:140-171) with multiple folders that have differing per-folder severity filters. So the wiring — iterating ws.Folders(), calling FilterSeverityForFolder, and feeding the result into aggregateSeverityFilters — and whether the resulting MixedSeverity actually reaches the rendered HTML (filter-mixed class) is untested. This is the headline IDE-1866 feature.
> 
> 2. filterState fallback branches untested. No test covers:
> 
>     1. ws == nil → falls back to global config.GetFilterSeverity (:141-142)
>     2. a folder whose FolderConfigReadOnly() is nil or ConfigResolver is nil (:150-151) → skipped
>     3. all folders skipped (len(severities) == 0) → global fallback retained
> 
> 4. IssueViewOptions "first folder only" behavior untested. The known asymmetry flagged in the review (filterState reads IVO from firstFC only, :161-163, vs aggregated severity) has no test pinning the current behavior. The existing IVO tests (:243, :308) use a single folder, so they can't distinguish "first folder" from "aggregate."
> 
> 5. aggregateSeverityFilters uses filters[0] as the baseline, so when folders disagree the returned agg value for a mixed severity is "whatever folder 0 had." The code comment (:176-177) says this is fine because the template checks MixedSeverity first. That template precedence is not verified by any test in this file — it's an untested coupling (also called out in the lifeguard analysis).

internal/context/context.go
```diff
@@ -134,6 +134,22 @@ func DependenciesFromContext(ctx context.Context) (map[string]any, bool) {
 	return d, ok
 }
 
+// copyDependenciesFromContext returns a shallow copy of the dependencies map
+// stored in ctx, or an empty map if none is present. Merge helpers use this so
+// that enriching a context never mutates the map shared with the parent context.
+// Without the copy, deriving multiple contexts concurrently from a single parent
+// (e.g. one scan goroutine per folder) races on the shared map.
+func copyDependenciesFromContext(ctx context.Context) map[string]any {
```
rrama: FYI, cli_scanner.go `enrichContext` has the same mutation bug. As a follow up, best to check all uses of `DependenciesFromContext`. Maybe we shouldn't export `DependenciesFromContext` at all.

rrama (approved):
While I had a differing opinion on how the setting should work when you click one of the buttons, I do think this way is good enough.
My prefered approach would have been that writes should unset at the folder level and set at the global level.

I have made the assumption that the changes in the settings page go away with Basti's change, but if they do not then let me know.

rrama:
Oh and add screenshots to your PR description please.

domain/ide/treeview/tree_builder.go
andrewrobinsonhodges-snyk: Wrapped both the SAST and Delta blocks in a single `if conf := cfg.Conf(); conf != nil` guard, so `GetSastSettings` is no longer called with a possibly-nil conf and the block matches the pattern below it. (Note: `GetSastSettings` already nil-checks internally, so there was no actual panic risk — but the guard makes it consistent and explicit.)

domain/ide/command/toggle_tree_filter_test.go
andrewrobinsonhodges-snyk: Added `TestToggleTreeFilter_Execute_SendsConfigurationNotification` — it wires in a real `MockNotifier` + `configResolver` (as `command_factory.go` does in production) and asserts an `LspConfigurationParam` is sent whose folder config carries the toggled value (`severity_filter_high=false`). The push is synchronous in `Execute`, and the workspace's async `HandleConfigChange` uses a separate notifier, so the assertion is race-free.

domain/ide/treeview/template/tree.js
andrewrobinsonhodges-snyk: Agreed it's not ideal. The hand-rolled tooltip exists because native `title` tooltips render inconsistently inside IDE webviews (noted in the comment above the block) — so this isn't really a Bootstrap-vs-custom choice for the tree view. Happy to revisit when the settings page moves off Bootstrap. Leaving as-is for now.

infrastructure/configuration/template/js/features/filter-sync.js
andrewrobinsonhodges-snyk: Keep as is for now and I'll address when we merge Basti's changes

domain/ide/treeview/template/tree.js
andrewrobinsonhodges-snyk: Fixed in 59ac69fd — `computeRelativePath` now returns `filepath.ToSlash(rel)` so the display label uses `/` on all platforms, restoring middle-truncation on Windows (relPath is display-only; the path id is unchanged).

domain/ide/treeview/template/tree.js
andrewrobinsonhodges-snyk: Fixed in 59ac69fd — `truncateMiddleByWidth` now returns the original text when `best === 0`, so we don't swap a short prefix for a same-width `…`; CSS end-truncates instead.

domain/ide/treeview/template/styles.css
andrewrobinsonhodges-snyk: Fixed in 23d95e0d — added `display: flex` to `.untrusted-folder-list-row` so the flex-direction/align/gap rules apply.

internal/context/context.go
andrewrobinsonhodges-snyk: Fixed in 9cb1d32f — `enrichContext` no longer mutates the shared deps map; exported `ctx2.CopyDependenciesFromContext` (mirrors the internal copy helper) and copy-before-mutate. Read-only callers untouched; the broader audit / unexport of `DependenciesFromContext` is a follow-up.

domain/ide/treeview/tree_node.go
andrewrobinsonhodges-snyk: Done in 068e426a — renamed `FolderPaths` → `UntrustedFolderPaths` (field, json tag, `WithUntrustedFolderPaths`, template).

domain/ide/treeview/tree_builder.go
andrewrobinsonhodges-snyk: Done in f9d85e3d — trimmed the change-history out of the `countWord` doc comment.

domain/ide/treeview/tree_scan_emitter.go
andrewrobinsonhodges-snyk: Done in f9d85e3d — added a TODO noting IVO reads from the first folder only; will aggregate into a `MixedIssueViewOptions` (analogous to `MixedSeverity`) when the IVO toolbar buttons land (separate branch, same release).

domain/ide/treeview/template/tree.js
andrewrobinsonhodges-snyk: Done in 59ac69fd — added the comment noting there's no fallback when `MutationObserver` is unavailable (expand/collapse re-truncation is simply skipped).

domain/ide/treeview/template/tree.js
andrewrobinsonhodges-snyk: Done in 59ac69fd — extended the comment to note the optimistic flip doesn't account for org-locked folders (the button flips active then snaps back to filter-mixed on the next LS re-render — brief flicker; resolved state stays correct).

domain/ide/treeview/tree_builder_test.go
andrewrobinsonhodges-snyk: Done in a97bac84 — added a test for Code + Agent Fix enabled + zero fixable, asserting the "no issues automatically fixable" line (covers tree_builder.go:420).

domain/ide/treeview/tree_node.go
andrewrobinsonhodges-snyk: Done in a97bac84 — added RenderTreeView template tests for MixedSeverity (filter-mixed class + mixed tooltip), scanner Tooltip→title, data-product-id, and product-icon--abbreviated. Item 5 (InfoVariant untrusted-folder/filter-empty) is deferred to the follow-up untrusted-support PR.

domain/ide/treeview/tree_scan_emitter_test.go
andrewrobinsonhodges-snyk: Done in a97bac84 — added filterState tests: multi-folder differing filters → MixedSeverity → filter-mixed (end-to-end), the three fallback branches (ws==nil, nil folder config/resolver skipped, all-skipped global fallback), current IVO first-folder behavior (pinned — will change with the IVO branch), and the filters[0]/MixedSeverity precedence coupling.

PR #1318, opened 2026-06-03
fix: eliminate data race in fflags.LoadFeatureFlags() [IDE-2103]

bastiandoetsch:
## Summary

- Replaces the mutable package-level `featureFlag` var with `sync.Once` + `cached`/`errOnce` vars so the embedded JSON is parsed exactly once and concurrent callers cannot race on shared mutable state
- `LoadFeatureFlags()` now returns a copy of the cached struct (not a pointer into it), preventing callers from mutating shared state through the returned pointer
- Extracts an unexported `parseFeatureFlags([]byte)` helper so the broken-JSON test no longer needs to mutate the global `featuresEmbed` var (itself a latent race)
- Adds `TestLoadFeatureFlags_concurrent` — 20 goroutines calling `LoadFeatureFlags()` concurrently; fails on the old code under `go test -race`, passes cleanly after the fix

## Test plan

- [ ] `go test -race ./internal/fflags/...` passes (all three tests)
- [ ] `go test -race -run ^TestLoadFeatureFlags_concurrent$ ./internal/fflags/...` passes (exercises first-initialisation race in isolation)
- [ ] `make test` passes

Fixes IDE-2103

___

### **PR Type**
Bug fix, Enhancement

___

### **Description**
- Eliminates data race in `LoadFeatureFlags`.

- Ensures feature flags are parsed only once using `sync.Once`.

- Prevents callers from mutating shared cached state.

- Adds comprehensive concurrent execution tests.

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A[LoadFeatureFlags] --> B{sync.Once.Do};
  B --> C{parseFeatureFlags};
  C --> D[cached FeatureFlag];
  C --> E[errCached error];
  D -- return copy --> F[Caller];
  E -- return error --> F;
  B -- already done --> G[Read cached];
  G --> D;
  D -- return copy --> F;
```

<tr>
  <td>
    <details>
      <summary><strong>features.go</strong><dd><code>Implement concurrent-safe feature flag loading</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

internal/fflags/features.go

<ul><li>Introduced <code>sync.Once</code>, <code>cached</code>, and <code>errCached</code> for safe, single <br>initialization.<br> <li> Added <code>parseFeatureFlags</code> helper for unmarshalling logic.<br> <li> Modified <code>LoadFeatureFlags</code> to use <code>once.Do</code> and return a shallow copy of <br>cached flags.</ul>

</details>

  </td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1318/changes#diff-d3173c12dbaffb92cf68cdf7d369c30dd6e601c917b3c0b97082cd5407e2176e">+26/-5</a>&nbsp; &nbsp; </td>

</tr>
<tr>
  <td>
    <details>
      <summary><strong>features_test.go</strong><dd><code>Add concurrent test and refactor broken JSON test</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

internal/fflags/features_test.go

<ul><li>Added <code>TestLoadFeatureFlags_concurrent</code> to verify thread-safe access.<br> <li> Refactored <code>TestLoadFeatureFlags_brokenJSON</code> to use the new <br><code>parseFeatureFlags</code> helper.<br> <li> Added necessary imports for <code>sync</code> and <code>require</code>.</ul>

</details>

  </td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1318/changes#diff-98a4808e5476b117b2cf06124eac9aef43f3c9654da967158e8da7f3783aef38">+38/-3</a>&nbsp; &nbsp; </td>

</tr>

</details>

___

### DIFF
--- a/internal/fflags/features.go
+++ b/internal/fflags/features.go
@@ -20,6 +20,7 @@ package fflags
 import (
 	_ "embed"
 	"encoding/json"
+	"sync"
 )
 
 type FeatureFlag struct {
@@ -29,13 +30,33 @@ type FeatureFlag struct {
 var (
 	//go:embed features.json
 	featuresEmbed []byte
-	featureFlag   FeatureFlag
+
+	once      sync.Once
+	cached    FeatureFlag
+	errCached error
 )
 
-func LoadFeatureFlags() (*FeatureFlag, error) {
-	err := json.Unmarshal(featuresEmbed, &featureFlag)
-	if err != nil {
-		return nil, err
+func parseFeatureFlags(data []byte) (FeatureFlag, error) {
+	var ff FeatureFlag
+	if err := json.Unmarshal(data, &ff); err != nil {
+		return FeatureFlag{}, err
 	}
-	return &featureFlag, nil
+	return ff, nil
+}
+
+// LoadFeatureFlags returns the parsed feature flags. The embedded JSON is
+// unmarshalled exactly once; errCached is non-nil only if the embedded asset
+// is malformed at build time, which TestLoadFeatureFlags would also catch.
+func LoadFeatureFlags() (*FeatureFlag, error) {
+	once.Do(func() {
+		cached, errCached = parseFeatureFlags(featuresEmbed)
+	})
+	if errCached != nil {
+		return nil, errCached
+	}
+	// Shallow copy is safe because FeatureFlag contains only value-typed fields
+	// (strings). If a slice, map, or pointer is ever added, replace this with
+	// an explicit deep copy to avoid sharing mutable state between callers.
+	cp := cached
+	return &cp, nil
 }
--- a/internal/fflags/features_test.go
+++ b/internal/fflags/features_test.go
@@ -17,9 +17,11 @@
 package fflags
 
 import (
+	"sync"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestLoadFeatureFlags(t *testing.T) {
@@ -29,9 +31,42 @@ func TestLoadFeatureFlags(t *testing.T) {
 	assert.Equal(t, "test-feature with value 555", ff.TestFeature)
 }
 
+// TestLoadFeatureFlags_brokenJSON calls parseFeatureFlags directly to verify
+// JSON parse errors are propagated. The errCached branch in LoadFeatureFlags
+// is unreachable in practice: the embedded asset is compiled in at build time,
+// so a malformed features.json would be caught by TestLoadFeatureFlags first.
 func TestLoadFeatureFlags_brokenJSON(t *testing.T) {
-	featuresEmbed = []byte("{{{broken-json")
-
-	_, err := LoadFeatureFlags()
+	_, err := parseFeatureFlags([]byte("{{{broken-json"))
 	assert.Error(t, err)
+}
+
+// TestLoadFeatureFlags_concurrent exercises concurrent calls to LoadFeatureFlags.
+// Package-level state is reset before launching goroutines so the test covers
+// the once.Do initialisation path regardless of which tests ran before it.
+func TestLoadFeatureFlags_concurrent(t *testing.T) {
+	// Reset package-level state so this test exercises the once.Do init path,
+	// not just the cached-read fast path.
+	once = sync.Once{}
+	cached = FeatureFlag{}
+	errCached = nil
+
+	const goroutines = 20
+
+	var wg sync.WaitGroup
+	results := make([]*FeatureFlag, goroutines)
+	errs := make([]error, goroutines)
+
+	for i := 0; i < goroutines; i++ {
+		wg.Add(1)
+		go func(idx int) {
+			defer wg.Done()
+			results[idx], errs[idx] = LoadFeatureFlags()
+		}(i)
+	}
+	wg.Wait()
+
+	for i := 0; i < goroutines; i++ {
+		require.NoError(t, errs[i])
+		assert.Equal(t, "test-feature with value 555", results[i].TestFeature)
+	}
 }
--- a/.gitignore
+++ b/.gitignore
@@ -102,3 +102,4 @@ brain/
 .verify-range
 .verify-agent-progress
 .pr-review-*.md
+.verify-metrics-pending


internal/fflags/features_test.go
```diff
@@ -30,8 +32,32 @@ func TestLoadFeatureFlags(t *testing.T) {
...
-	_, err := LoadFeatureFlags()
+	_, err := parseFeatureFlags([]byte("{{{broken-json"))
 	assert.Error(t, err)
 }
+
+// TestLoadFeatureFlags_concurrent is meaningful under go test -race.
+// When run as part of the full suite, once.Do has already fired (TestLoadFeatureFlags
+// runs first), so this test exercises concurrent reads of the cached path.
+// Run with -run ^TestLoadFeatureFlags_concurrent$ -race to cover concurrent initialisation.
+func TestLoadFeatureFlags_concurrent(t *testing.T) {
```
acke, line 46: 🟡 **Medium**

**[features_test.go:43]** `TestLoadFeatureFlags_concurrent` is mislabelled. When run under `go test ./...`, `TestLoadFeatureFlags` fires `once.Do` first, so the 20 goroutines all hit the cached-read fast path — the test does not exercise the once.Do initialisation race it claims to cover. Your own comment on line 39-42 acknowledges this and recommends `-run ^TestLoadFeatureFlags_concurrent$`, but CI typically runs the full suite, so race-protection coverage relies on a developer remembering to run it standalone.

**Fix:**
Reset package state at the top of the test before launching goroutines, then run under `-race`:
```go
once = sync.Once{}
cached = FeatureFlag{}
errOnce = nil
```
This makes the test self-contained and meaningful as part of the full suite.

— AI review comment

internal/fflags/features_test.go
```diff
@@ -30,8 +32,32 @@ func TestLoadFeatureFlags(t *testing.T) {
 }
 
 func TestLoadFeatureFlags_brokenJSON(t *testing.T) {
-	featuresEmbed = []byte("{{{broken-json")
-
-	_, err := LoadFeatureFlags()
+	_, err := parseFeatureFlags([]byte("{{{broken-json"))
```
acke, line 39: 🟡 **Medium**

**[features_test.go:35]** `TestLoadFeatureFlags_brokenJSON` now calls `parseFeatureFlags` directly instead of going through `LoadFeatureFlags`. The `if errOnce \!= nil { return nil, errOnce }` branch of the public function is no longer covered by any test. Practical impact is low (the embedded asset is build-time immutable), but it is a real coverage regression on the public API.

**Fix:**
Either (a) inject the parser via a package-level function var so the test can swap in a broken-parser stub and exercise `LoadFeatureFlags`'s error branch, or (b) accept the gap and add a comment noting that the error path is unreachable in practice because `TestLoadFeatureFlags` would catch a malformed embed at build time.

— AI review comment

internal/fflags/features.go
```diff
@@ -29,13 +30,30 @@ type FeatureFlag struct {
...
+func parseFeatureFlags(data []byte) (FeatureFlag, error) {
+	var ff FeatureFlag
+	if err := json.Unmarshal(data, &ff); err != nil {
+		return FeatureFlag{}, err
+	}
+	return ff, nil
+}
+
+// LoadFeatureFlags returns the parsed feature flags. The embedded JSON is
+// unmarshalled exactly once; cachedErr is non-nil only if the embedded asset
```
acke: 🔵 **Low**

**[features.go:48]** Doc comment refers to `cachedErr is non-nil only if ...`, but the actual variable is named `errOnce`. Stale identifier in the comment.

**Fix:**
Either rename the variable to `cachedErr` (matches `cached`, reads better, fixes the comment in one shot) or update the comment to say `errOnce`. Renaming is preferable — `errOnce` reads like it's a `sync.Once`-typed thing, which it isn't.

— AI review comment

internal/fflags/features.go
```diff
@@ -29,13 +30,30 @@ type FeatureFlag struct {
...
-	if err != nil {
-		return nil, err
+	once.Do(func() {
+		cached, errOnce = parseFeatureFlags(featuresEmbed)
+	})
+	if errOnce != nil {
+		return nil, errOnce
 	}
-	return &featureFlag, nil
+	cp := cached
```
acke, line 60: 🔵 **Low**

**[features.go:57]** `cp := cached` works today because `FeatureFlag` contains only a `string` field (Go strings are immutable and pass by value). If a future contributor adds a `[]string`, `map[...]...`, or any pointer/slice field to `FeatureFlag`, this shallow copy will silently become a shared-state leak — callers will be able to mutate cached slice/map backing arrays through the returned pointer.

**Fix:**
Add a code comment above the struct or near `cp := cached` explicitly stating the invariant: "FeatureFlag must contain only value-typed fields; if you add a slice/map/pointer, replace `cp := cached` with an explicit deep copy." Cheap insurance against silent regression.

— AI review comment

acke (commented):
## Orchestrated review (review-orchestrator)

**Verdict: APPROVE WITH NITS**

The `sync.Once` + cached-value pattern is correct. `sync.Once.Do` establishes happens-before for the reads of `cached`/`errOnce` outside the closure, so the new pattern is sound under the Go memory model. No critical or high issues. Four inline comments cover test quality and a doc nit (2 Medium, 2 Low). The Info-level observations below don't need their own threads.

### ℹ️ Info — included in body to avoid inline noise

ℹ️ **Info**

**[features.go]** Behaviour change worth a line in the PR body: callers used to share one pointer into the global and could mutate cached state through it; they now get a fresh per-call copy and mutations don't persist. Observable impact today is zero (`FeatureFlag` is a single immutable string and there are no non-test callers in-repo — see below), but external importers of `github.com/snyk/snyk-ls/internal/fflags` may exist.

**Fix:**
Add a one-liner to the PR description noting the behavioural contract change. No code change required.

— AI review comment

ℹ️ **Info**

**[features.go]** `gh code search` across `snyk/snyk-ls` finds zero non-test callers of `LoadFeatureFlags()`. The race fix is theoretical for in-repo use; only reachable for external importers. Consider whether the package should be unexported (or moved internal-only) in a follow-up — out of scope for this PR.

**Fix:**
No change needed in this PR. Optional follow-up ticket to audit whether `fflags` needs to remain exported.

— AI review comment

ℹ️ **Info**

**[features.go:54-56]** `errOnce` is sticky for process lifetime by construction (no retry path). This is correct — `featuresEmbed` is built in at compile time, so a parse failure cannot self-heal at runtime. A one-line code comment would prevent a future reader from filing this as a bug.

**Fix:**
Add: `// errOnce is sticky: featuresEmbed is built-in at compile time, so a parse failure cannot self-heal at runtime.`

— AI review comment

### Verified correct (no finding)

- `sync.Once.Do` happens-before guarantee for reads of `cached`/`errOnce` outside the closure — confirmed under the Go memory model.
- The defensive `cp := cached` copy is meaningful today (prevents callers from mutating cached state through the returned pointer), even though it is shallow.
- `parseFeatureFlags` extraction removes the previous `featuresEmbed = ...` mutation in tests — that was an actual latent race in the old code, now eliminated.

### Summary

| Lens          | Findings        |
|---------------|-----------------|
| Logic         | 0 (verified correct) |
| Security      | 0               |
| Tests         | 2 (Medium) — inline |
| Contracts     | 1 (Info) — above |
| AI-patterns   | 2 (Low, Info) — inline + above |
| Verification  | 7/7 candidates confirmed (Phase B trace + Phase D re-verify) |

Tools: 5 lenses (logic, security, tests, contracts, ai-patterns) inline due to small diff; verification subagent ran Phase B + D. Adversarial-review skipped (diff 53 LOC < 200 threshold). Snyk-rules skipped (no local checkout).

acke (approved):
APPROVE WITH NITS

internal/fflags/features_test.go
bastiandoetsch: Fixed — `TestLoadFeatureFlags_concurrent` now resets `once`, `cached`, and `cachedErr` at the top before launching goroutines, making it self-contained and exercising the actual `once.Do` initialisation race in any run order.

internal/fflags/features_test.go
bastiandoetsch: Added a comment to `TestLoadFeatureFlags_brokenJSON` explaining that the `cachedErr` branch in `LoadFeatureFlags` is unreachable in practice — the embedded asset is compiled in at build time, so a malformed embed would be caught by `TestLoadFeatureFlags` before reaching this path.

internal/fflags/features.go
bastiandoetsch: Fixed — renamed `errOnce` → `cachedErr` throughout `features.go` and `features_test.go`. The doc comment now matches the variable name.

internal/fflags/features.go
bastiandoetsch: Added an inline comment near `cp := cached` stating that `FeatureFlag` must contain only value-typed fields — if a slice, map, or pointer is ever added, the shallow copy must be replaced with an explicit deep copy.

bastiandoetsch:
Fixed — renamed `errOnce` → `cachedErr` throughout. The doc comment and variable name now match.

PR #1321, opened 2026-06-04
feat: block unauthenticated Snyk API requests via GAF middleware [IDE-1988]

nick-y-snyk:
### Description

Enables `STOP_REQUESTS_WITHOUT_AUTH` on the GAF engine so requests to Snyk
API URLs without an `Authorization` header are intercepted in-memory with a
synthetic 401 — no wire call, no latency. This eliminates the N failing
round-trips fired during unauthenticated sessions (slow settings page,
log/error-reporter spam).

Uses a local `go.mod` replace directive pointing at the GAF feature branch
(`feat/IDE-1988_stop-requests-without-auth`) until GAF PR #620 merges to main.
Remove the replace directive and bump the GAF version after that merge.

Also switches the ggshield pre-commit hook from the isolated virtualenv install
to the system binary (`repo: local`, `language: system`) so the hook uses the
already-authenticated system ggshield.

Jira: [IDE-1988](https://snyk.atlassian.net/browse/IDE-1988)

### Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,12 +24,14 @@ repos:
         stages: [pre-push]
         verbose: true
         pass_filenames: false
-  - repo: https://github.com/gitguardian/ggshield
-    rev: v1.43.0
+  - repo: local
     hooks:
       - id: ggshield
-        language_version: python3
+        name: ggshield (pre-commit)
+        entry: ggshield secret scan pre-commit
+        language: system
         stages: [pre-commit]
+        pass_filenames: false
   - repo: https://github.com/alessandrojcm/commitlint-pre-commit-hook
     rev: v9.22.0
     hooks:
--- a/application/di/init.go
+++ b/application/di/init.go
@@ -161,6 +161,7 @@ func initDomain(tokenService types.TokenService, conf configuration.Configuratio
 
 func initInfrastructure(tokenService types.TokenService, conf configuration.Configuration, engine workflow.Engine, logger *zerolog.Logger) {
 	gafConfiguration := conf
+	gafConfiguration.Set(configuration.STOP_REQUESTS_WITHOUT_AUTH, true)
 
 	fs := pflag.NewFlagSet("snyk-ls-config", pflag.ContinueOnError)
 	types.RegisterAllConfigurations(fs)
--- a/application/server/authentication_flows_e2e_test.go
+++ b/application/server/authentication_flows_e2e_test.go
@@ -351,7 +351,8 @@ func (s *startupAuthRequestLdxSyncService) RefreshConfigFromLdxSync(ctx context.
 	require.True(s.t, ok)
 
 	req := httptest.NewRequest(http.MethodGet, "https://api.snyk.io/rest/self", nil).WithContext(ctx)
-	require.NoError(s.t, provider.Authenticator().AddAuthenticationHeader(req))
+	_, err := provider.Authenticator().AddAuthenticationHeader(req)
+	require.NoError(s.t, err)
 
 	s.mu.Lock()
 	defer s.mu.Unlock()
--- a/application/server/authentication_flows_e2e_test.go
+++ b/application/server/authentication_flows_e2e_test.go
@@ -351,7 +351,7 @@ func (s *startupAuthRequestLdxSyncService) RefreshConfigFromLdxSync(ctx context.
 	require.True(s.t, ok)
 
 	req := httptest.NewRequest(http.MethodGet, "https://api.snyk.io/rest/self", nil).WithContext(ctx)
-	_, err := provider.Authenticator().AddAuthenticationHeader(req)
+	err := provider.Authenticator().AddAuthenticationHeader(req)
 	require.NoError(s.t, err)
 
 	s.mu.Lock()
--- a/go.mod
+++ b/go.mod
@@ -141,7 +141,7 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
 
-//replace github.com/snyk/go-application-framework => ../go-application-framework
+replace github.com/snyk/go-application-framework => ../go-application-framework
 
 //replace github.com/snyk/code-client-go => ../code-client-go
 
--- a/infrastructure/authentication/auth_service_impl.go
+++ b/infrastructure/authentication/auth_service_impl.go
@@ -155,6 +155,8 @@ func (a *AuthenticationServiceImpl) Shutdown() {
 	}
 }
 
+var _ AuthenticationService = (*AuthenticationServiceImpl)(nil)
+
 func (a *AuthenticationServiceImpl) Provider() AuthenticationProvider {
 	a.m.RLock()
 	defer a.m.RUnlock()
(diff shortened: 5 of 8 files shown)


go.mod
```diff
@@ -141,7 +141,7 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
 
-//replace github.com/snyk/go-application-framework => ../go-application-framework
+replace github.com/snyk/go-application-framework => ../go-application-framework
```
bastiandoetsch, line 144: [Critical] This uncommented `replace` points the module at a sibling `../go-application-framework` checkout that only exists on your machine. Combined with the removed `go-application-framework` checksum lines in `go.sum`, the build resolves the published GAF version `v0.1.1-0.20260601153725-33af46ffdfa5`, which does not contain `configuration.STOP_REQUESTS_WITHOUT_AUTH`. CI and any clean checkout fail to compile (`application/di/init.go:164:37: undefined: configuration.STOP_REQUESTS_WITHOUT_AUTH`), which is why every job on this PR is red. It builds locally only because the sibling directory is present.

Fix: publish and tag the GAF change containing the constant and the blocking middleware, bump `go.mod` to that version, re-comment (remove) this `replace` directive, and run `go mod tidy` to restore the `go.sum` checksums.

— AI review

application/server/authentication_flows_e2e_test.go
```diff
@@ -351,7 +351,8 @@ func (s *startupAuthRequestLdxSyncService) RefreshConfigFromLdxSync(ctx context.
 	require.True(s.t, ok)
 
 	req := httptest.NewRequest(http.MethodGet, "https://api.snyk.io/rest/self", nil).WithContext(ctx)
-	require.NoError(s.t, provider.Authenticator().AddAuthenticationHeader(req))
+	err := provider.Authenticator().AddAuthenticationHeader(req)
```
bastiandoetsch, line 354: [Should Fix] The feature's core claim — that a Snyk API request without an Authorization header is blocked with a synthetic 401 — has no test. This change only splits an existing assertion that checks the positive path (the header IS added), so the new blocking behavior is unverified from this repo's side.

Fix: add a test that issues a Snyk API request with no Authorization header and asserts it is blocked (synthetic 401), plus one confirming an authenticated request still passes. (Splitting the original one-line `require.NoError(provider.Authenticator().AddAuthenticationHeader(req))` into two lines is otherwise equivalent — pure churn.)

— AI review

.pre-commit-config.yaml
```diff
@@ -24,12 +24,14 @@ repos:
         stages: [pre-push]
         verbose: true
         pass_filenames: false
-  - repo: https://github.com/gitguardian/ggshield
-    rev: v1.43.0
+  - repo: local
     hooks:
       - id: ggshield
-        language_version: python3
+        name: ggshield (pre-commit)
```
bastiandoetsch, line 30: [Should Fix] Switching ggshield from the version-pinned upstream repo (`gitguardian/ggshield rev: v1.43.0`) to a `language: system` hook drops both version pinning and managed installation. Contributors who don't have `ggshield` on `$PATH` get a failing (or silently skipped) secret scan, and scan behavior drifts with whatever local version each developer has. This is also unrelated to the auth-gating ticket.

Fix: keep the pinned upstream hook, or if a system hook is required, pin a minimum version and document the install step in CONTRIBUTING. Ideally split this into its own change with justification.

— AI review

application/di/init.go
```diff
@@ -161,6 +161,7 @@ func initDomain(tokenService types.TokenService, conf configuration.Configuratio
 
 func initInfrastructure(tokenService types.TokenService, conf configuration.Configuration, engine workflow.Engine, logger *zerolog.Logger) {
 	gafConfiguration := conf
+	gafConfiguration.Set(configuration.STOP_REQUESTS_WITHOUT_AUTH, true)
```
bastiandoetsch, line 164: [Should Fix] `STOP_REQUESTS_WITHOUT_AUTH` is enabled unconditionally with no feature gate or kill-switch. The scoping is sound — verified that the gate only affects Snyk API URLs on the authorized client, so `GetUnauthorizedHttpClient` and non-Snyk learn/installer calls are unaffected — so there is no false-positive risk for legitimate pre-auth traffic.

The residual concern is operational: an always-on security policy has no way to disable it short of a code change and redeploy if an unforeseen blocking case appears. Consider gating it behind the existing feature-flag mechanism for safe rollout, or confirm that always-on with no kill-switch is the intended product decision.

— AI review

infrastructure/authentication/auth_service_impl.go
```diff
@@ -724,6 +705,33 @@ func (a *AuthenticationServiceImpl) doAuthCheck(conf configuration.Configuration
 	return true
 }
 
+// handleTransientAuthError sends a deduplicated balloon notification for
```
bastiandoetsch, line 708: [Suggestion] The extraction of `handleTransientAuthError` is a faithful, behavior-preserving move (same lock ordering, 30s dedup window, message, and log line) — nice. One small note: the doc comment frames it as handling "transient (non-logout) auth failures", but it also serves the offline case where `err == nil` (user empty, no error). The `if err != nil` guard handles that correctly; the comment just slightly understates that path. Optional wording tweak.

— AI review

bastiandoetsch:
> ℹ️ This finding is in a file not changed by this PR. It is non-blocking for this review — shared for awareness.

[Suggestion · non-blocking] OAuth vs token-auth asymmetry in how unauthenticated Snyk API calls are blocked. In GAF `auth_header.go`, the `STOP_REQUESTS_WITHOUT_AUTH` gate fires for the token authenticator with an empty token (no header set, nil returned → synthetic 401). For the OAuth authenticator with no token (the default IDE case), `AddAuthenticationHeader` returns an *error* earlier, so `RoundTrip` returns before the gate. The end goal (outbound call not sent) holds for both, but callers and retry/response middleware see different result types (error vs 401).

Fix: add a one-line comment in GAF noting the gate is a backstop for the token-auth path, since OAuth is already covered by the error path.

— AI review

bastiandoetsch:
[Suggestion · non-blocking] Scope creep: this PR bundles the security flag with an unrelated `handleTransientAuthError` extraction, a `var _ AuthenticationService` interface assertion, the ggshield pre-commit rewrite, and a test-assertion split. None are required to block unauthenticated requests.

Fix: keep the security change to the flag + its test; move the refactors and the ggshield change to separate commits/PRs. This shrinks review surface and makes the security-relevant change easy to revert in isolation.

— AI review

bastiandoetsch:
> ℹ️ These items are process/dependency notes, non-blocking for this review — shared for awareness.

[Should Fix · non-blocking] **Requirements & ticket scope (IDE-1988):** no requirements are captured for IDE-1988 in the central store, and the ticket's Acceptance also lists three items not addressed here — consolidating scattered auth guards, removing PR #1245's `IsSet` gate + prime calls, and wiring `server.go:485 startOfflineDetection` into the state source. If this PR is one slice of a multi-PR stack, please mark those as explicit follow-ups on the ticket so the Acceptance isn't silently left open.

[Suggestion · non-blocking] **Pre-existing dependency vulns** (not introduced by this diff, no fix available): `golang.org/x/crypto/ssh@0.51.0` (3 high — improper authentication, missing release of resource), plus medium/low in `golang.org/x/net/html@0.54.0` and `go-git/v5`. Tracking only.

— AI review

go.mod
bastiandoetsch, line 144: [Critical] This activates a local `replace` pointing at `../go-application-framework`, a sibling directory that doesn't exist in CI or any other clone, and the published GAF checksums were removed from `go.sum`. The build fails everywhere but the author's machine — CI is fully red, and `STOP_REQUESTS_WITHOUT_AUTH` only resolves against the local GAF checkout. All 4 reviewers flagged this. The commit message confirms it's a stopgap until GAF #620 merges.

This PR is not mergeable in this state. Once GAF #620 is published: bump the `require` version to the real release, restore the `go.sum` entries, and re-comment this `replace`. Keep the PR a draft until then.

— AI review

application/di/init.go
bastiandoetsch, line 164: [Should Fix] Three concerns on enabling this flag:

1. No test covers the behavior. The only test change is a cosmetic happy-path edit; nothing asserts an unauthenticated Snyk API request is blocked with the synthetic 401, nor that legitimately-unauthenticated flows still work.
2. Global-scope risk. The flag is Set on the shared engine configuration, so it applies to every consumer. The CLI installer/download, Learn service, and Sentry deliberately use `GetUnauthorizedHttpClient()`. If GAF scopes the block by request URL rather than by client, those flows break. This is unverifiable here (GAF source unresolvable) and untested.
3. Ordering. The flag is set after `engine.GetConfiguration()`, and clients/NetworkAccess are built shortly after. If GAF reads the flag at client-build time rather than per-request, the Set may be a no-op or need to precede client construction.

Fix: add an e2e/integration test for (a) a no-auth Snyk API request → 401, no wire call, and (b) an unauthorized-client flow (e.g. CLI download) still succeeding; and confirm GAF's flag scoping/ordering.

— AI review

.pre-commit-config.yaml
```diff
@@ -24,12 +24,14 @@ repos:
...
+        entry: ggshield secret scan pre-commit
+        language: system
```
bastiandoetsch, line 32: [Should Fix] Switching ggshield from the pinned managed hook (`rev: v1.43.0`, isolated virtualenv) to `language: system` drops version pinning and requires every developer to have `ggshield` on PATH — without it the secret-scan hook silently no-ops or errors. This is also unrelated to the PR's stated purpose (scope creep in a security PR). Keep the pinned managed hook, or split this out with its own rationale.

— AI review

bastiandoetsch:
[Suggestion · non-blocking] This security PR bundles three unrelated changes: the auth flag, a behaviour-preserving extraction of `handleTransientAuthError`, and a compile-time interface assertion (plus the ggshield hook change). Since the go.mod/go.sum state forces a likely revert/rebase, the entangled refactor makes a clean cherry-pick harder. Consider splitting the refactor and the pre-commit change into their own PRs. Also worth a one-line comment at the `Set(STOP_REQUESTS_WITHOUT_AUTH, true)` call documenting the security invariant and IDE-1988, so it isn't silently removed later.

— AI review

PR #1324, opened 2026-06-05
fix: reset summary panel on stop-scan [IDE-1035]

acke:
## Summary
- New `resetSummaryPanelOnStopScan` helper reuses the existing `resetSummaryPanelForOrgChange` reset path.
- `windowWorkDoneProgressCancelHandler` invokes it after `progress.Cancel`, so pressing the stop-scan button now clears the summary back to its initial state instead of leaving stale counts on screen.
- Three unit tests: happy path, nil aggregator, no folders.

## Why
When the user pressed stop-scan, the summary panel kept showing the previous scan's counts even though the scan was cancelled. The org-change flow already had a helper to reset the panel to initial state — this PR wires the same helper into the stop-scan LSP handler (`window/workDoneProgress/cancel`).

## Test plan
- [x] `TestResetSummaryPanelOnStopScan_CallsInit`
- [x] `TestResetSummaryPanelOnStopScan_NilAgg_DoesNotPanic`
- [x] `TestResetSummaryPanelOnStopScan_NoFolders_DoesNotCallInit`
- [ ] Manual: start a scan, press stop, observe the summary returns to initial / empty.

Jira: [IDE-1035](https://snyksec.atlassian.net/browse/IDE-1035)
Triage: [Buglist](https://snyksec.atlassian.net/wiki/spaces/IDE/pages/4860444692/Buglist)

[IDE-1035]: https://snyksec.atlassian.net/browse/IDE-1035?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

___

### **PR Type**
Bug fix, Enhancement

___

### **Description**
- Reset the summary panel state when a scan is stopped.

- Ensure scan cancellations do not race with scan result updates.

- Distinguish scan progress tokens from generic progress notifications.

___

### Diagram Walkthrough

```mermaid
flowchart LR
  User[User clicks Stop Scan] --> RPC[LSP RPC Handler]
  RPC -->|calls `window/workDoneProgress/cancel`| Handler{handleWindowWorkDoneProgressCancel}
  Handler -->|checks `IsScanToken`| IsScan[progress.IsScanToken]
  IsScan -->|No| Handler
  IsScan -->|Yes| Handler
  Handler -->|defers `progress.Cancel`| DeferredCancel[progress.Cancel (Deferred)]
  Handler -->|registers callback| Register[Scanner.RegisterCancelCallback]
  Scanner[Scanner.Scan] -->|starts goroutines| ScanGoroutines[Scan Goroutines]
  ScanGoroutines -->|writes results| AggregatorSetDone[Scan Aggregator (SetScanDone)]
  ScanGoroutines -->|context cancelled| ScannerScanWait[Scanner.Scan WaitGroups]
  ScannerScanWait -->|after WaitGroups| Consume[Scanner.consumeCancelCallback]
  Consume -->|invokes callback| AggregatorInit[ScanAggregator.Init (Reset)]
  DeferredCancel -->|eventually runs| ActualCancel[progress.Cancel]
```

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Tests</strong></td><td><details><summary>5 files</summary><table>
<tr>
  <td><strong>cancel_handler_test.go</strong><dd><code>Add tests for scan token cancellation handling</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1324/changes#diff-910540ce871778228b8b8d3486094db926daae7ee2abdf3d9567db358089d850">+165/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>configuration_test.go</strong><dd><code>Update tests for reset summary panel function rename</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1324/changes#diff-e141faafb3f0507956b85e92ac3b584c7c706bd96e7f428c7ef95d0d8b2df564">+13/-10</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>notification_test.go</strong><dd><code>Test non-scan token cancellation does not reset aggregator</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1324/changes#diff-09b80ca28f0c9f8db60c941c24e92dfc65ded48c06249e70e2004faec30b9891">+34/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>scanner_test.go</strong><dd><code>Add tests for ordered cancel callback execution</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1324/changes#diff-ca51a6c8a5aee5d9347b6425bca1629ea179a6722ad1bb0eae8fb52fdb490327">+178/-11</a></td>

</tr>

<tr>
  <td><strong>progress_test.go</strong><dd><code>Add tests for scan token identification and folder association</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1324/changes#diff-82673a4178c5a088aeedb0642c91ddb54a7b254b6ad40eeb4a27e55ace3e7980">+48/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Refactoring</strong></td><td><details><summary>1 files</summary><table>
<tr>
  <td><strong>configuration.go</strong><dd><code>Rename and update summary panel reset function</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1324/changes#diff-eb3fe6fd1956591838595b08979653ec23c0b0c16f5340e9e89a05c30e921607">+9/-8</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Enhancement</strong></td><td><details><summary>8 files</summary><table>
<tr>
  <td><strong>server.go</strong><dd><code>Implement gated scan cancellation reset handler</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1324/changes#diff-1b5a200b7dc3f37ed67632d0c2682eeb53762813f96edb15cb7660e076fa6031">+56/-4</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>scanner.go</strong><dd><code>Add cancel callback registration and consumption to scanner</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1324/changes#diff-ca9273debc8b1b659b1a630df729fe1be56f139b823c0b6d4d5a296d55ba47cb">+48/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>test_scanner.go</strong><dd><code>Add no-op cancel callback to test scanner</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1324/changes#diff-8cbcd4aa66d49d14534ace6605aff5020b7c031e495e8348548cd3471b771307">+4/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>code.go</strong><dd><code>Tag code scanner progress as scan token</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1324/changes#diff-19cf169a3cc81ebe57b31aaa2c780a2fcb7986bdf656981a4bb42ac3b8dc6845">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>code_tracker.go</strong><dd><code>Ensure code tracker uses non-scan token</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1324/changes#diff-817c4b4f3a18faab6639560da9543bc42035a6d32363d2d23e015e10519307f8">+5/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>iac.go</strong><dd><code>Tag IaC scanner progress as scan token</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1324/changes#diff-ececb95448061db2f05addd42eb0f500c604a1bce72216b3e1815b0771407b39">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>cli_scanner.go</strong><dd><code>Tag OSS scanner progress as scan token</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1324/changes#diff-6303c702b1a7d4e9fb6a9f930cfbe2546fa60cced2ba801246cb356bd2a047a6">+4/-4</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>progress.go</strong><dd><code>Introduce scan token type and folder association</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1324/changes#diff-66d4349586a1dea4bd4de639d9a4a0439f4e011ca6b4f8667b1852baef6cf7c6">+63/-4</a>&nbsp; &nbsp; </td>

</tr>

</details>

___


application/server/server.go
```diff
@@ -1176,10 +1176,12 @@ func textDocumentHover() jrpc2.Handler {
 	})
 }
 
-func windowWorkDoneProgressCancelHandler() jrpc2.Handler {
+func windowWorkDoneProgressCancelHandler(conf configuration.Configuration) jrpc2.Handler {
 	return handler.New(func(ctx context.Context, params types.WorkdoneProgressCancelParams) (any, error) {
 		ctx2.LoggerFromContext(ctx).Debug().Str("method", "WindowWorkDoneProgressCancelHandler").Interface("params", params).Msg("RECEIVING")
 		progress.Cancel(params.Token)
+		scanAgg, _ := scanStateAggregatorFromContext(ctx)
```
bastiandoetsch: [Should Fix] `window/workDoneProgress/cancel` is a generic LSP notification, but this unconditionally resets the scan summary panel for every cancel — including non-scan progress. CLI install/download uses cancellable trackers (`infrastructure/cli/install/downloader.go`), so cancelling a download (or any non-scan progress) while scan results are displayed wipes the panel back to 'NotStarted', losing valid counts. Flagged by 3 of 4 reviewers.

Fix: gate the reset on whether `params.Token` belongs to a scan tracker, so only scan cancels reset the panel.

— AI review

bastiandoetsch: [Should Fix] Ordering race against the in-flight scan. `progress.Cancel` only signals cancellation over a channel; scanner goroutines (`domain/snyk/scanner/scanner.go`) keep running and call `SetScanInProgress`/`SetScanDone` on the same aggregator. A late `SetScanDone` can land *after* this synchronous `Init` reset, re-populating the panel with the cancelled scan's state (or causing a visible empty→refill flash). The shared mutex prevents a data race but not this logical clobber, and no test exercises a concurrent/just-cancelled scan. (Note: a parallel reviewer checked only the data-race and deemed it safe — the logical ordering race is the actual concern.)

Fix: reset after the scan goroutines observe cancellation and stop writing (e.g. converge aggregator state in the scan lifecycle on cancel), and add an outcome-level test asserting the final snapshot is the initial state.

— AI review

application/server/server.go
```diff
@@ -1176,10 +1176,12 @@ func textDocumentHover() jrpc2.Handler {
 	})
 }
 
-func windowWorkDoneProgressCancelHandler() jrpc2.Handler {
+func windowWorkDoneProgressCancelHandler(conf configuration.Configuration) jrpc2.Handler {
 	return handler.New(func(ctx context.Context, params types.WorkdoneProgressCancelParams) (any, error) {
 		ctx2.LoggerFromContext(ctx).Debug().Str("method", "WindowWorkDoneProgressCancelHandler").Interface("params", params).Msg("RECEIVING")
 		progress.Cancel(params.Token)
```
bastiandoetsch: [Suggestion] The `ok` from `scanStateAggregatorFromContext` is discarded, so a missing aggregator silently skips the reset. The analogous org-change path uses `mustScanStateAggregatorFromContext` (panics on absence). Consider logging a debug/warning when `ok` is false so a genuine wiring failure is observable rather than silently swallowed.

— AI review

application/server/configuration.go
```diff
@@ -876,6 +876,15 @@ func resetSummaryPanelForOrgChange(scanAgg scanstates.Aggregator, folderPaths []
 	scanAgg.Init(folderPaths)
 }
 
+// resetSummaryPanelOnStopScan resets the summary panel to its initial state
+// when the user cancels a scan via the IDE stop-scan button (IDE-1035).
+// It resolves current workspace folder paths from conf and delegates to
+// resetSummaryPanelForOrgChange, which guards against nil aggregator and
+// empty folder list.
+func resetSummaryPanelOnStopScan(scanAgg scanstates.Aggregator, conf configuration.Configuration) {
```
bastiandoetsch: [Suggestion] `resetSummaryPanelForOrgChange` is now shared by two semantically different triggers (org change and stop-scan). A maintainer on the stop-scan path lands on a function whose name asserts org-change semantics. Consider renaming the shared helper neutrally (e.g. `resetSummaryPanel`) with the two callers as the only specialization.

— AI review

application/server/server.go
acke: Fixed — gated the panel reset on `progress.IsScanToken(params.Token)`. Scan trackers are now created via a new `NewScanTracker` constructor (used by OSS, Code, IaC scanners and the code-client tracker factory); download/install trackers keep using `NewTracker` so their cancel events no longer wipe the panel. New `TestCancelProgress_NonScanToken_DoesNotResetAggregator` covers the negative case.

acke: Fixed — moved the reset into the scan lifecycle. The LSP cancel handler now registers a per-folder reset callback on `DelegatingConcurrentScanner` (via new `RegisterCancelCallback`); `Scan()` consumes the callback (`LoadAndDelete`) only **after** both `waitGroup` and `referenceBranchScanWaitGroup` return, so all `SetScanDone` writes are complete before the aggregator `Init`. New `TestScan_CancelCallback_CalledAfterGoroutinesFinish` asserts `SetScanDone` precedes `Init` in the call sequence.

One edge case worth noting: if user cancels mid-scan when only some folders are in-flight, the handler still registers callbacks for all workspace folders, so a stale callback could fire on a later `Scan()` for a folder that wasn't actively scanning. Easy follow-up — happy to clear stale callbacks at `Scan()` entry if you'd like a v2.

application/server/server.go
acke: Fixed — the handler now logs a `Warn` when the aggregator is missing from context ("scan state aggregator not found in context; summary panel will not be reset"). Chose Warn over panic to avoid bringing down an LSP notification handler over a wiring miss.

application/server/configuration.go
acke: Fixed — renamed `resetSummaryPanelForOrgChange` → `resetSummaryPanel`; all callers, comments, and tests updated. The narrower `resetSummaryPanelOnStopScan` wrapper stays as the call site for the stop-scan flow.

application/server/server.go
```diff
@@ -1176,10 +1176,46 @@ func textDocumentHover() jrpc2.Handler {
...
+			// Register the reset callback on the scanner so it fires AFTER all
+			// per-product goroutines have finished their SetScanDone writes (E,
+			// IDE-1035). The reset is keyed to each workspace folder so that
+			// concurrent folder scans reset independently.
+			sc, scOk := scannerFromContext(ctx)
+			if scOk {
+				if dcs, ok := sc.(*scanner2.DelegatingConcurrentScanner); ok {
+					folderPaths := workspaceFolderPaths(conf)
+					for _, fp := range folderPaths {
+						dcs.RegisterCancelCallback(fp, func() {
```
bastiandoetsch: [Critical] The reset callback is registered *after* `progress.Cancel(params.Token)` (line 1187). `Cancel` unblocks the in-flight scan goroutines, so `Scan()` can drain both WaitGroups and run `consumeCancelCallback` (scanner.go:399) *before* this `RegisterCancelCallback` stores the callback. When that happens the reset is silently dropped — the exact bug this PR fixes — and the late `Store` leaves an orphaned callback that fires at the end of the folder's *next* scan, wiping a valid result. The `LoadAndDelete`-after-WaitGroups ordering only guards against late `SetScanDone`; it does not guard register-vs-consume.

Fix: register the callback *before* calling `progress.Cancel`, so the handoff has a happens-before guarantee:

```go
sc, scOk := scannerFromContext(ctx)
if dcs, ok := sc.(*scanner2.DelegatingConcurrentScanner); scOk && ok {
    for _, fp := range workspaceFolderPaths(conf) {
        dcs.RegisterCancelCallback(fp, func() { resetSummaryPanel(scanAgg, []types.FilePath{fp}) })
    }
}
progress.Cancel(params.Token)
```

Better still, move reset-on-cancel inside the scanner so no handler-injected closure or downcast is needed. Confirmed by 3 of 4 reviewers.

— AI review

application/server/server.go
```diff
@@ -1176,10 +1176,46 @@ func textDocumentHover() jrpc2.Handler {
...
+				return nil, nil
+			}
+			// Register the reset callback on the scanner so it fires AFTER all
+			// per-product goroutines have finished their SetScanDone writes (E,
+			// IDE-1035). The reset is keyed to each workspace folder so that
+			// concurrent folder scans reset independently.
+			sc, scOk := scannerFromContext(ctx)
+			if scOk {
+				if dcs, ok := sc.(*scanner2.DelegatingConcurrentScanner); ok {
+					folderPaths := workspaceFolderPaths(conf)
```
bastiandoetsch: [Should Fix] A stop-scan cancels one product token, but this loop registers a reset callback for *every* workspace folder, and `consumeCancelCallback` fires unconditionally at the end of *every* `Scan()` (scanner.go:399) — it is not correlated to the canceled token or folder. Consequences: (1) other products still scanning that folder get their panel wiped; (2) folders that weren't scanning keep a pending callback that fires on their next scan, resetting a valid result. You flagged this same case in the thread above ("a stale callback could fire on a later Scan() for a folder that wasn't actively scanning") — it is a correctness issue, not just a v2 nicety.

Fix: scope the callback to the folder/token actually being canceled, and only fire it for that scan (key by token, or check `ctx.Err()` is a user cancel inside `Scan` before resetting).

— AI review

application/server/configuration.go
```diff
@@ -866,16 +866,25 @@ func applyOrganization(conf configuration.Configuration, engine workflow.Engine,
...
 		return
 	}
 	scanAgg.Init(folderPaths)
 }
 
+// resetSummaryPanelOnStopScan resets the summary panel to its initial state
+// when the user cancels a scan via the IDE stop-scan button (IDE-1035).
+// It resolves current workspace folder paths from conf and delegates to
+// resetSummaryPanel, which guards against nil aggregator and empty folder list.
+func resetSummaryPanelOnStopScan(scanAgg scanstates.Aggregator, conf configuration.Configuration) {
```
bastiandoetsch: [Should Fix] `resetSummaryPanelOnStopScan` has no production callers — `rg` shows only the three `TestResetSummaryPanelOnStopScan_*` tests call it. The actual stop-scan handler (server.go) calls `resetSummaryPanel` / `RegisterCancelCallback` directly. The thread above states this wrapper "stays as the call site for the stop-scan flow," but it is not wired into that flow. Its three dedicated tests give false confidence that the stop-scan path is covered while exercising code that never runs in production.

Fix: either wire the handler through this wrapper, or delete the wrapper and its three tests and cover the real handler path instead.

— AI review

application/server/server.go
```diff
@@ -1176,10 +1176,46 @@ func textDocumentHover() jrpc2.Handler {
...
+					Msg("scan state aggregator not found in context; summary panel will not be reset")
+				return nil, nil
+			}
+			// Register the reset callback on the scanner so it fires AFTER all
+			// per-product goroutines have finished their SetScanDone writes (E,
+			// IDE-1035). The reset is keyed to each workspace folder so that
+			// concurrent folder scans reset independently.
+			sc, scOk := scannerFromContext(ctx)
+			if scOk {
+				if dcs, ok := sc.(*scanner2.DelegatingConcurrentScanner); ok {
```
bastiandoetsch: [Should Fix] Type-asserting the `Scanner` interface down to the concrete `*scanner2.DelegatingConcurrentScanner` couples application/server to a domain implementation detail — `RegisterCancelCallback` is not on the interface. Any other `Scanner` falls through to the synchronous fallback at line 1217, which resets *immediately* (racing in-flight `SetScanDone` writes — the very race the callback path exists to avoid) and resets *all* folders. A correct path and a knowingly-racy path selected by a runtime type assertion is fragile.

Fix: express the reset-on-cancel capability on the `Scanner` interface (or perform it inside the scanner), removing both the downcast and the racy fallback.

— AI review

domain/snyk/scanner/scanner_test.go
```diff
@@ -222,6 +223,137 @@ func Test_ScanStarted_TokenChanged_ScanCancelled(t *testing.T) {
...
+			callsMu.Unlock()
+		},
+	}
+
+	resolver := defaultResolver(t, engine)
+	sc, _ := setupScannerWithResolverAndAgg(t, engine, tokenService, resolver, agg, mockProductScanner)
+
+	folderPath := types.FilePath(t.TempDir())
+	resetCalled := make(chan struct{}, 1)
+	sc.(*DelegatingConcurrentScanner).RegisterCancelCallback(folderPath, func() {
```
bastiandoetsch, line 258: [Should Fix] This test registers the callback *before* starting the scan, then cancels — the opposite of production, where `progress.Cancel` runs before `RegisterCancelCallback` (server.go:1187 then 1205). It therefore cannot catch the register-vs-consume race flagged on server.go. No test drives the real `window/workDoneProgress/cancel` handler with a scan token while a scan is in flight, so the behavior the ticket asks for is unverified end-to-end.

Fix: add an integration test that invokes the cancel handler with a `NewScanTracker` token during an in-flight scan (Cancel→Register order) and asserts the panel is reset exactly once, after `SetScanDone`.

— AI review

internal/progress/progress.go
```diff
@@ -33,6 +33,15 @@ import (
 var trackersMutex sync.RWMutex
 var trackers = make(map[types.ProgressToken]*Tracker)
 var ToServerProgressChannel = make(chan types.ProgressParams, 1000)
+
+// scanTokensMutex guards scanTokens.
+var scanTokensMutex sync.RWMutex
+
+// scanTokens is the set of progress tokens that belong to scan operations.
```
bastiandoetsch: [Suggestion] `scanTokens` is a second registry parallel to `trackers`, keyed by the same token, that must be kept in sync in three places (`deleteTracker`, `Cancel`, and the bulk reset in `CleanupChannels`). Any future tracker-removal path that forgets to also touch `scanTokens` leaks a token or leaves `IsScanToken` stale.

Fix: model "is scan" as an `isScan bool` field on the `Tracker` already held in `trackers`. `IsScanToken` becomes a single lookup, and the separate map, mutex, and three-way cleanup all go away.

— AI review

domain/snyk/scanner/scanner.go
```diff
@@ -38,6 +38,12 @@ import (
 	"github.com/snyk/snyk-ls/internal/vcs"
 )
 
+// cancelCallbacks maps folderPath → reset function registered by the LSP
+// cancel handler (IDE-1035). Callbacks are consumed (and removed) by Scan()
+// after all per-product goroutines have finished, ensuring the reset happens
```
bastiandoetsch: [Suggestion] `type cancelCallbackMap = sync.Map` adds a name but no type safety — keys/values stay `any`, so `consumeCancelCallback` needs the defensive `v.(func())` assertion. Access is one register and one consume per folder, so `sync.Map`'s concurrency machinery isn't needed.

Fix: a plain `map[types.FilePath]func()` guarded by a small mutex is simpler and type-safe. (Moot if the registry is removed per the reset-in-scanner suggestion.)

— AI review

domain/snyk/scanner/scanner_test.go
```diff
@@ -222,6 +223,137 @@ func Test_ScanStarted_TokenChanged_ScanCancelled(t *testing.T) {
...
+}
+
+func (r *recordingOrderAggregator) SetScanDone(fp types.FilePath, p product.Product, ref bool, err error) {
+	r.NoopStateAggregator.SetScanDone(fp, p, ref, err)
+	if r.SetScanDoneFn != nil {
+		r.SetScanDoneFn()
+	}
+}
+
+func setupScannerWithResolverAndAgg(t *testing.T, engine workflow.Engine, tokenService types.TokenService, configResolver types.ConfigResolverInterface, agg scanstates.Aggregator, testProductScanners ...types.ProductScanner) (
```
bastiandoetsch, line 375: [Suggestion] `setupScannerWithResolverAndAgg` duplicates the full `NewDelegatingScanner` construction from `setupScannerWithResolver`; the only difference is the explicit `agg` parameter.

Fix: have `setupScannerWithResolver` delegate to `setupScannerWithResolverAndAgg` with a default aggregator.

— AI review

application/server/server.go
bastiandoetsch: Re-reviewed — this isn't fully closed. The callback is consumed after the WaitGroups (good for the late-`SetScanDone` race), but it's **registered after** `progress.Cancel` runs, so a fast scan can consume before the callback is stored — the reset is dropped and the orphaned callback then fires on the next scan. See the [Critical] on server.go:1205. The all-folders / next-scan case you noted is also a correctness issue, not just a v2 — flagged as [Should Fix] on server.go:1203.

— AI review

application/server/configuration.go
bastiandoetsch: The rename is fine, but `resetSummaryPanelOnStopScan` is **not** the call site for the stop-scan flow — `rg` shows only the three tests call it; the handler uses `resetSummaryPanel` directly. So the wrapper is dead production code with three tests covering an unused path. See [Should Fix] on configuration.go:884.

— AI review

application/server/server.go
acke: Fixed the register-vs-consume race: `progress.Cancel(params.Token)` is now `defer`-ed, so any `RegisterCancelCallback` runs first and the happens-before guarantee holds. The deferred Cancel runs once on every return path, so the warning and fallback branches still cancel correctly.

You're right that the broader architectural concerns (downcast in #3382206417, per-folder over-fire in #3382206240, parallel `scanTokens` registry in #3382206602, `sync.Map` alias in #3382206707) are real. Moving reset-on-cancel into the scanner needs per-product token correlation (each `ProductScanner` creates its own scan tracker — e.g. `infrastructure/code/code.go:267` — so the LSP cancel propagates per-product, not per-folder), which is more involved than the original fix shape. Tracking that as a follow-up; this PR ships the race fix only.

application/server/configuration.go
acke: Fixed — deleted `resetSummaryPanelOnStopScan` and its three `TestResetSummaryPanelOnStopScan_*` tests. The real stop-scan path in `windowWorkDoneProgressCancelHandler` is what now exercises `resetSummaryPanel` directly, so removing the never-called wrapper drops the false-coverage signal without losing any behaviour. `initRecordingAggregator` stays — it's still used elsewhere in the file and by `notification_test.go`.

application/server/server.go
acke: Fixed — `RegisterCancelCallback(folderPath, fn)` is now on the `Scanner` interface (`domain/snyk/scanner.Scanner`), so the cancel handler calls it via the interface directly. The `*scanner2.DelegatingConcurrentScanner` downcast and the racy synchronous fallback at line 1217 are both gone; if the scanner is missing from context (tests / early startup) we now log debug and skip the reset rather than racing in-flight `SetScanDone` writes. `TestScanner` got a no-op implementation since it doesn't run the cancel-drain cycle.

The broader per-folder over-fire concern in #3382206240 still needs the bigger redesign (per-product token correlation) and is tracked separately.

application/server/server.go
acke: Won't fix in this PR — same root cause as the architectural concerns acknowledged on #3382206124. Per-product token correlation (each `ProductScanner` creates its own scan tracker, e.g. `infrastructure/code/code.go:267`, so cancel propagates per-product, not per-folder) is the missing piece. Tracking that as a follow-up issue alongside the downcast (already addressed), parallel `scanTokens` registry (#3382206602), and `sync.Map` alias (#3382206707) so the redesign can land coherently rather than piecemeal.

domain/snyk/scanner/scanner.go
acke: Fixed — `cancelCallbackMap` type alias is gone. The field is now a plain `map[types.FilePath]func()` guarded by a `sync.Mutex`, so the defensive `v.(func())` cast in `consumeCancelCallback` is no longer needed and the values stay typed. Tests pass under `go test -race ./domain/snyk/scanner/...`.

domain/snyk/scanner/scanner_test.go
acke: Fixed — added handler-level tests in `application/server/cancel_handler_test.go` exercising `windowWorkDoneProgressCancelHandler` directly:

- `ScanToken_RegistersBeforeCancel` — uses `progress.NewScanTracker(...)`, asserts a callback is registered for each workspace folder, then asserts the deferred `progress.Cancel` has run by handler return (`!IsScanToken(token)`) — that's the register-vs-consume ordering guarantee end-to-end.
- `NonScanToken_NoRegistration` — verifies a plain (non-scan) tracker token cancels but never registers a reset, so generic progress can't wipe scan results.
- `ScanToken_NoScanner_NoSyncFallback` — proves the racy synchronous fallback is gone; missing scanner now logs-and-skips instead.

To keep the production code untouched I extracted the handler body into a free function `handleWindowWorkDoneProgressCancel(ctx, params, conf)` that the jrpc2 closure now delegates to and the tests drive directly.

The companion scanner-layer test `TestScan_CancelCallback_CalledAfterGoroutinesFinish` (already on the branch) covers the in-flight drain-then-reset ordering — `SetScanDone` precedes the registered callback, asserted across a real `DelegatingConcurrentScanner` with a blocking `ProductScanner`. Together the two layers cover the reviewer's ask without spinning up a full LSP server inside the handler test.

domain/snyk/scanner/scanner_test.go
acke: Fixed — `setupScannerWithResolver` now delegates to `setupScannerWithResolverAndAgg` with `scanstates.NewNoopStateAggregator()` as the default. The duplicated `NewDelegatingScanner` construction is gone.

internal/progress/progress.go
acke: Fixed — replaced the parallel `scanTokens` map (plus its mutex) with an `isScan bool` field on `Tracker`. `IsScanToken` is now a single `trackers` lookup under `trackersMutex.RLock`; `deleteTracker`, `Cancel`, and `CleanupChannels` no longer carry a second mutation, so a future tracker-removal path can't desync. `go test -race ./internal/progress/...` is green.

application/server/cancel_handler_test.go
```diff
@@ -0,0 +1,169 @@
...
+	require.NoError(t, err)
+
+	// Both workspace folders must have had a callback registered.
+	got := scanner.registered()
+	assert.Contains(t, got, folderA, "callback must be registered for folder A")
+	assert.Contains(t, got, folderB, "callback must be registered for folder B")
+	assert.Len(t, got, 2, "exactly one callback per folder")
+
+	// The deferred progress.Cancel must have run by the time the handler returns,
+	// so the scan token is no longer recognised. This is the ordering guarantee
```
bastiandoetsch: [Should Fix] `recognised` is a misspelling of `recognized` — golangci-lint's misspell checker will fail CI on this line.

Fix: change `recognised` to `recognized` in the comment.

— AI review

internal/progress/progress.go
```diff
@@ -33,13 +33,20 @@ import (
...
 	channel              chan types.ProgressParams
 	cancelChannel        chan bool
 	token                types.ProgressToken
 	cancellable          bool
+	// isScan distinguishes scan-operation trackers (NewScanTracker) from
+	// generic progress trackers so window/workDoneProgress/cancel can scope
+	// the summary-panel reset to scan cancellations only (IDE-1035).
+	// Written under trackersMutex when the tracker is registered; only ever
+	// read via IsScanToken under the same lock.
+	isScan               bool
```
bastiandoetsch: [Should Fix] `gofmt` reports this file has misaligned struct fields — CI will fail the formatting check.

`gofmt -l internal/progress/progress.go` produces a diff that collapses the extra spaces on fields in the `Tracker` struct (both the pre-existing `channel`/`cancelChannel`/`token`/`cancellable` fields and the newly added `isScan` field use manual multi-space alignment rather than gofmt's single-tab standard).

Fix: run `gofmt -w internal/progress/progress.go` (or `make fmt`).

— AI review

application/server/cancel_handler_test.go
acke: Fixed in c7161833 — `recognised` → `recognized` in the comment so the misspell linter is happy.

internal/progress/progress.go
acke: Fixed in c7161833 — ran `gofmt -w internal/progress/progress.go`. The new `isScan` field's preceding comment split the Tracker struct's alignment group; reflowed the pre-comment fields (`channel`/`cancelChannel`/`token`/`cancellable`) to the gofmt-canonical narrow alignment so the formatting check passes.

application/server/server.go
```diff
@@ -1176,14 +1176,58 @@ func textDocumentHover() jrpc2.Handler {
...
+	// per-product goroutines have finished their SetScanDone writes
+	// (IDE-1035). The reset is keyed to each workspace folder so that
+	// concurrent folder scans reset independently.
+	sc, scOk := scannerFromContext(ctx)
+	if !scOk {
+		logger.Debug().Str("method", "WindowWorkDoneProgressCancelHandler").
+			Msg("scanner not in context; summary panel will not be reset")
+		return nil, nil
+	}
+	for _, fp := range workspaceFolderPaths(conf) {
```
basti-snyk: **Should fix — cancel resets every folder, not the cancelled one.** This loop registers a reset callback for *every* workspace folder, but `params.Token` identifies a single per-product tracker belonging to one folder. In a multi-folder workspace, cancelling folder A's scan also registers reset callbacks on folders B, C… Those callbacks are not consumed now (no in-flight scan there) and instead fire on each of those folders' *next* `Scan()` completion (`scanner.go` `consumeCancelCallback`), wiping valid results the user never asked to clear. Consider mapping the cancelled token back to its owning folder and registering the reset only for that folder.

(Three of the four reviewers independently flagged this as the root cause of a cluster of data-loss bugs.)

— AI review

domain/snyk/scanner/scanner.go
```diff
@@ -358,6 +399,13 @@ func (sc *DelegatingConcurrentScanner) Scan(ctx context.Context, pathToScan type
 	waitGroup.Wait()
 	referenceBranchScanWaitGroup.Wait()
 
+	// After all per-product goroutines have exited (including their SetScanDone
+	// writes), invoke any cancel callback registered by the LSP cancel handler
+	// for this folder (IDE-1035). Consuming the callback here — after the
+	// WaitGroups — guarantees the aggregator reset cannot race with in-flight
+	// SetScanDone calls.
+	sc.consumeCancelCallback(folderPath)
```
basti-snyk, line 407: **Should fix — consume is unconditional and unreachable on early returns.** Two related issues:

1. `consumeCancelCallback(folderPath)` fires at the end of *every* `Scan()` with no check that *this* scan was cancelled. A callback registered during a cancel therefore also resets the panel after the folder's next *successful* scan — turning "reset on cancel" into "reset on next scan completion", wiping valid results.

2. This call sits after the WaitGroups, past several early returns in `Scan()` (missing FolderConfig, offline, not authenticated, already-cancelled ctx). If the next thing that happens to the folder hits one of those guards, the registered callback is neither fired nor cleared — it lingers and fires against a still-later scan.

Consider draining/clearing the folder's callback in a `defer` at the top of `Scan()` (after `folderPath` is known) so every exit path consumes it exactly once, and gate the reset on whether the scan was actually cancelled.

— AI review

application/server/server.go
nick-y-snyk: Fixed in 4be7b67e — the cancel handler no longer loops over every workspace folder. `Tracker` now records its owning folder at creation (`NewScanTracker(cancellable, logger, folderPath)`), and a new `progress.FolderForScanToken(token)` resolves the cancelled token back to that single folder. The handler registers exactly one `RegisterCancelCallback` for it; if the token isn't found (already consumed, or a race with an extremely fast scan), it skips the reset entirely rather than falling back to resetting anything broader.

Handler test rewritten to assert exactly one registration for the cancelled folder and that an untouched second folder is left alone (`TestHandleWindowWorkDoneProgressCancel_ScanToken_RegistersBeforeCancel`). New scanner-level test `TestScan_CancelCallbackForFolderA_DoesNotAffectFolderB` proves a folder-A cancel-reset never surfaces after folder B's own successful scan.

domain/snyk/scanner/scanner.go
nick-y-snyk: Partially addressed by 4be7b67e, but not closed. That commit fixes the multi-folder over-registration (#3474144036) — a cancel now registers exactly one callback, for the token's own folder, via the new `progress.FolderForScanToken`. That closes the specific case where an untouched folder B ends up with a dangling callback because folder A was cancelled.

What's unchanged: `consumeCancelCallback` is still consumed unconditionally at the end of every `Scan()` for a folder, and the early-return paths you listed (missing FolderConfig, offline, not authenticated, already-cancelled ctx) still bypass that consume point. So the same-folder version of this — a cancel-armed callback surviving past an early-return and firing on a later, unrelated scan of that same folder — is still possible. This needs the callback's consume point re-plumbed through every `Scan()` exit path (likely a `defer` right after `folderPath` is resolved), which is a larger, concurrency-sensitive change in its own right.

Tracking this alongside the per-product blast-radius gap (#3388243621) as a follow-up rather than bundling it into this fix — will file a Jira ticket referencing this PR before merge.

application/server/server.go
nick-y-snyk: Filed follow-up: IDE-2248 tracks the remaining product-level and same-folder consume-timing gaps referenced above.

application/server/server.go
```diff
@@ -1181,12 +1181,64 @@ func textDocumentHover() jrpc2.Handler {
...
-		ctx2.LoggerFromContext(ctx).Debug().Str("method", "WindowWorkDoneProgressCancelHandler").Interface("params", params).Msg("RECEIVING")
-		progress.Cancel(params.Token)
+		return handleWindowWorkDoneProgressCancel(ctx, params, conf)
+	})
+}
+
+// handleWindowWorkDoneProgressCancel is the testable body of
+// windowWorkDoneProgressCancelHandler. Kept as a free function so integration
+// tests can drive it without going through jrpc2.Handler indirection.
+func handleWindowWorkDoneProgressCancel(ctx context.Context, params types.WorkdoneProgressCancelParams, conf configuration.Configuration) (any, error) {
```
basti-snyk: **Should fix (new since last review) — dead `conf` parameter threaded through the handler.** `handleWindowWorkDoneProgressCancel(ctx, params, conf)` never reads `conf` — the aggregator and scanner both come from `ctx` (`scanStateAggregatorFromContext`/`scannerFromContext`), and no other line touches it. It's plumbed through `windowWorkDoneProgressCancelHandler(conf)` (server.go:1184/1186) and three test call sites for no effect, and `unparam`/`revive` will likely flag it. It also misleads a reader into thinking configuration influences the reset decision. Drop the parameter (revert the `windowWorkDoneProgressCancelHandler()` signature and the three test args), or use it.

— AI review

infrastructure/code/code_tracker.go
```diff
@@ -37,7 +37,12 @@ func NewCodeTrackerFactory(logger *zerolog.Logger) codeClientScan.TrackerFactory
 }
 
 func (t trackerFactory) GenerateTracker() codeClientScan.Tracker {
-	newTracker := progress.NewTracker(true, t.logger)
+	// No folder path in scope here, and it wouldn't matter: newCodeTracker
+	// below mints its own token and reports Cancellable: false, so this
+	// NewScanTracker's token/folder is never the one the IDE echoes back on
+	// cancel. This call site cannot be the target of a cancellation today;
+	// it only needs to keep compiling.
+	newTracker := progress.NewScanTracker(true, t.logger, "")
```
basti-snyk: **Suggestion (new since last review) — `NewScanTracker(true, …, "")` plants a latent `Init([""])` foot-gun.** This tags the tracker `isScan=true` with an empty `folderPath` and registers it in the global `trackers` map. `FolderForScanToken` returns `(folderPath, true)` for any `isScan` tracker, so for this one it returns `("", true)` — no empty-path guard. The comment correctly notes this token is never IDE-cancellable *today* (the code client mints its own `Cancellable:false` token), so it's unreachable now. But if any future path ever made such a token cancellable, the handler would pass `folderOk=true` with `folderPath=""` into `resetSummaryPanel` → `agg.Init([""])`, injecting a bogus empty-folder key. Cheap hardening: keep `NewTracker` here (it already compiled and needs no scan tagging), or make `FolderForScanToken` return `false` for an empty `folderPath`.

— AI review

application/server/server.go
nick-y-snyk: Fixed in d527b0b6 — dropped `conf` from `handleWindowWorkDoneProgressCancel`'s signature and updated the one caller plus the three test call sites.

infrastructure/code/code_tracker.go
nick-y-snyk: Fixed in be228d33 — swapped `NewScanTracker` for plain `NewTracker` at this call site, since this token is never IDE-cancellable and doesn't need scan classification.

nick-y-snyk (approved):

PR #1333, opened 2026-06-09
feat: block unauthenticated Snyk API requests via GAF middleware [IDE-1988]

nick-y-snyk:
### Description

Bumps GAF to v0.3.0 which includes the opt-in auth-header middleware and enables `STOP_REQUESTS_WITHOUT_AUTH` on the GAF engine.

Requests to Snyk API URLs without an `Authorization` header are now intercepted in-memory with a synthetic 401 — no wire call, no latency, no log/error-reporter spam during unauthenticated sessions.

Jira: [IDE-1988](https://snyk.atlassian.net/browse/IDE-1988)

### Checklist

### DIFF
--- a/application/di/init.go
+++ b/application/di/init.go
@@ -161,6 +161,7 @@ func initDomain(tokenService types.TokenService, conf configuration.Configuratio
 
 func initInfrastructure(tokenService types.TokenService, conf configuration.Configuration, engine workflow.Engine, logger *zerolog.Logger) {
 	gafConfiguration := conf
+	gafConfiguration.Set(configuration.STOP_REQUESTS_WITHOUT_AUTH, true)
 
 	fs := pflag.NewFlagSet("snyk-ls-config", pflag.ContinueOnError)
 	types.RegisterAllConfigurations(fs)
--- a/go.mod
+++ b/go.mod
@@ -34,7 +34,7 @@ require (
 	github.com/snyk/cli-extension-secrets v0.0.0-20260305092220-defe1129df99
 	github.com/snyk/code-client-go v1.26.1
 	github.com/snyk/error-catalog-golang-public v0.0.0-20260410094451-50af33399e90
-	github.com/snyk/go-application-framework v0.1.1-0.20260601153725-33af46ffdfa5
+	github.com/snyk/go-application-framework v0.3.0
 	github.com/snyk/studio-mcp v1.6.1
 	github.com/sourcegraph/go-lsp v0.0.0-20240223163137-f80c5dd31dfd
 	github.com/spf13/pflag v1.0.10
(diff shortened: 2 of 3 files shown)


rrama (approved):

application/di/init.go
```diff
@@ -161,6 +161,7 @@ func initDomain(tokenService types.TokenService, conf configuration.Configuratio
 
 func initInfrastructure(tokenService types.TokenService, conf configuration.Configuration, engine workflow.Engine, logger *zerolog.Logger) {
 	gafConfiguration := conf
+	gafConfiguration.Set(configuration.STOP_REQUESTS_WITHOUT_AUTH, true)
```
bastiandoetsch, line 164: **Should Fix:** `gafConfiguration` (line 163) is an interface alias for `conf`, not a clone — both point to the same `*extendedViper`. `Set()` here mutates the shared engine config, which the networking layer reads lazily at request time, so the feature works. However, the variable name implies isolation that doesn't exist: if a future engineer adds `.Clone()` to line 163 (a natural refactor given the name), the networking layer stops seeing this flag with no test failure to catch the regression.

Please add a comment explaining the shared-reference semantics:
```go
// gafConfiguration is an alias for conf (same *extendedViper) — Set() propagates
// to the networking layer's config at request time via shared reference. IDE-1988.
gafConfiguration.Set(configuration.STOP_REQUESTS_WITHOUT_AUTH, true)
```

Also: `TestInit` in `test_init.go` does not set this flag, so integration tests run with blocking disabled while production blocks. Add the same `Set()` call to `TestInit` and add at least one test asserting an unauthenticated Snyk API call returns 401 via the authorised client. — AI review

bastiandoetsch:
## AI Review Summary — snyk/snyk-ls #1333

**Verdict: Changes Requested** — 2 Should Fix, 2 Suggestions.

### Should Fix
1. **Alias fragility + missing comment** (`init.go:164`) — `gafConfiguration` is an interface alias for `conf`, not a clone. The feature works (shared `*extendedViper` propagates to the networking layer), but the variable name implies isolation. A future `.Clone()` refactor silently breaks the 401 gate with no test failure. Add a comment explaining the shared-reference intent. (inline comment posted)

2. **`TestInit` missing the flag — test/prod divergence** (`test_init.go:60`) — `TestInit` builds the DI graph without setting `STOP_REQUESTS_WITHOUT_AUTH`, so integration tests run with blocking disabled while production blocks. Add the `Set()` call to `TestInit` and add at least one test asserting an unauthenticated Snyk API call via `authorizedClient` returns a synthetic 401.

### Suggestions
3. **Stale `go.sum` entries** — `go.sum` contains `v0.3.0` hashes (intermediate development step) alongside `v0.3.1`. Run `go mod tidy` to remove the orphaned entries.

4. **OAuth gap not documented** — For OAuth users with expired tokens, the `AuthHeaderMiddleware` logs and passes through (error path), so they get a real 401 from the API rather than the synthetic one. This is a known GAF design tradeoff but not documented at the call site. A one-line comment (`// OAuth expired-token path: auth error → passthrough, not synthetic 401`) would prevent future confusion.

### Security
Clean. The `Set()` call correctly enables the 401 gate on the GAF networking layer. No secrets, no IaC changes. Pre-existing SCA vulns in `golang.org/x/crypto` not introduced by this PR.

— AI review

PR #1334, opened 2026-06-09
fix(treeview): use folder-level IssueViewOptions for info node message [IDE-2123]

nick-y-snyk:
### Description

The "Open and Ignored issues are disabled!" info node in the tree view always reflected the **global** `IssueViewOptions`, ignoring folder-level overrides. Users who enabled open/ignored issues at the folder level still saw the disabled message, even though the issue list itself rendered correctly.

Root cause: `BuildTree` populated `fd.IssueViewOptions` from a builder-wide field (`b.issueViewOptions`) set once from global config. The folder-aware resolver (`IssueViewOptionsForFolder`) was never consulted for the message path.

Fix: expose `IssueViewOptions()` on the `Folder` interface (matching `IsDeltaFindingsEnabled` / `DisplayableIssueTypes` pattern), delegate to the existing `IssueViewOptionsForFolder` resolver, and use it in `BuildTree`. Remove the now-dead `SetIssueViewOptions` setter and `issueViewOptions` field from `TreeBuilder`.

Jira: https://snyksec.atlassian.net/browse/IDE-2123

### Checklist

### DIFF
--- a/domain/ide/treeview/tree_builder.go
+++ b/domain/ide/treeview/tree_builder.go
@@ -62,7 +62,6 @@ type TreeBuilder struct {
 	productScanStates map[types.FilePath]map[product.Product]bool
 	productScanErrors map[types.FilePath]map[product.Product]string
 	pendingAutoExpand map[string]bool // deferred auto-expand writes applied after build
-	issueViewOptions  types.IssueViewOptions
 }
 
 // NewTreeBuilder creates a new TreeBuilder with the given expand state.
@@ -83,11 +82,6 @@ func (b *TreeBuilder) SetProductScanStates(states map[types.FilePath]map[product
 // SetProductScanErrors sets the per-(folder, product) scan error messages.
 func (b *TreeBuilder) SetProductScanErrors(errors map[types.FilePath]map[product.Product]string) {
 	b.productScanErrors = errors
-}
-
-// SetIssueViewOptions sets the current issue view options for building info nodes.
-func (b *TreeBuilder) SetIssueViewOptions(opts types.IssueViewOptions) {
-	b.issueViewOptions = opts
 }
 
 // BuildTree constructs tree view data from a workspace.
@@ -114,7 +108,7 @@ func (b *TreeBuilder) BuildTree(workspace types.Workspace) TreeViewData {
 			AllIssues:           allIssues,
 			FilteredIssues:      filtered,
 			DeltaEnabled:        f.IsDeltaFindingsEnabled(),
-			IssueViewOptions:    b.issueViewOptions,
+			IssueViewOptions:    f.IssueViewOptions(),
 		}
 		if cfg := f.FolderConfigReadOnly(); cfg != nil {
 			fd.ConsistentIgnoresEnabled = cfg.GetFeatureFlag(featureflag.SnykCodeConsistentIgnores)
--- a/domain/ide/treeview/tree_builder.go
+++ b/domain/ide/treeview/tree_builder.go
@@ -101,6 +101,7 @@ func (b *TreeBuilder) BuildTree(workspace types.Workspace) TreeViewData {
 		allIssues := fip.Issues()
 		filtered := fip.FilterIssues(allIssues, supportedTypes)
 
+		cfg := f.FolderConfigReadOnly()
 		fd := FolderData{
 			FolderPath:          f.Path(),
 			FolderName:          f.Name(),
@@ -110,7 +111,7 @@ func (b *TreeBuilder) BuildTree(workspace types.Workspace) TreeViewData {
 			DeltaEnabled:        f.IsDeltaFindingsEnabled(),
 			IssueViewOptions:    f.IssueViewOptions(),
 		}
-		if cfg := f.FolderConfigReadOnly(); cfg != nil {
+		if cfg != nil {
 			fd.ConsistentIgnoresEnabled = cfg.GetFeatureFlag(featureflag.SnykCodeConsistentIgnores)
 			if fd.DeltaEnabled {
 				conf := cfg.Conf()
--- a/domain/ide/treeview/tree_builder.go
+++ b/domain/ide/treeview/tree_builder.go
@@ -97,19 +97,22 @@ func (b *TreeBuilder) BuildTree(workspace types.Workspace) TreeViewData {
 		if !ok {
 			continue
 		}
-		supportedTypes := f.DisplayableIssueTypes()
+		// Fetch FolderConfig once; all derived values are computed from this single call
+		// to avoid repeated storage reads inside DisplayableIssueTypes, IsDeltaFindingsEnabled,
+		// and IssueViewOptions (each would otherwise invoke FolderConfigReadOnly separately).
+		cfg := f.FolderConfigReadOnly()
+		supportedTypes := f.DisplayableIssueTypesFromConfig(cfg)
 		allIssues := fip.Issues()
 		filtered := fip.FilterIssues(allIssues, supportedTypes)
 
-		cfg := f.FolderConfigReadOnly()
 		fd := FolderData{
 			FolderPath:          f.Path(),
 			FolderName:          f.Name(),
 			SupportedIssueTypes: supportedTypes,
 			AllIssues:           allIssues,
 			FilteredIssues:      filtered,
-			DeltaEnabled:        f.IsDeltaFindingsEnabled(),
-			IssueViewOptions:    f.IssueViewOptions(),
+			DeltaEnabled:        f.IsDeltaFindingsEnabledFromConfig(cfg),
+			IssueViewOptions:    f.IssueViewOptionsFromConfig(cfg),
 		}
 		if cfg != nil {
 			fd.ConsistentIgnoresEnabled = cfg.GetFeatureFlag(featureflag.SnykCodeConsistentIgnores)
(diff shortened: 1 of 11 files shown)


domain/ide/treeview/tree_scan_emitter_test.go
```diff
@@ -207,3 +207,55 @@ func TestTreeScanStateEmitter_Dispose_StopsRenderLoop(t *testing.T) {
 	// Emit after Dispose must not block or panic.
 	emitter.Emit(scanstates.StateSnapshot{AnyScanInProgressWorkingDirectory: true})
 }
+
+// TestTreeScanStateEmitter_FolderLevelIssueViewOptions verifies that the info-node message
+// reflects folder-level IssueViewOptions overrides rather than the global setting.
+// Regression for: global open+ignored disabled, folder-level open+ignored enabled →
+// "Open and Ignored issues are disabled!" must NOT appear.
+func TestTreeScanStateEmitter_FolderLevelIssueViewOptions(t *testing.T) {
```
bastiandoetsch, line 216: **Should Fix:** This test passes vacuously on both pre-fix and post-fix code. `zeroIssuesText` (and `issueCountText`) only enters the `IssueViewOptions` branch when `ctx.consistentIgnoresEnabled` is `true`. `SetupWorkspace` uses `featureflag.NewFakeService()` whose `Flags` map is empty — so `GetFeatureFlag(SnykCodeConsistentIgnores)` returns `false` for every folder, the early-return fires, and `"Open and Ignored issues are disabled!"` can never appear regardless of which code path is used.

Fix: add `featureFlagSvc.Override(featureflag.SnykCodeConsistentIgnores, true)` (or equivalent) before the folder setup, so the test actually exercises the IVO branch and would have caught the original bug. — AI review

bastiandoetsch:
## AI Review Summary — snyk/snyk-ls #1334

**Verdict: Changes Requested** — 1 Should Fix (inline), 1 Should Fix (out-of-diff), 2 Suggestions.

### Should Fix
1. **Regression test passes vacuously** (`tree_scan_emitter_test.go:215`) — `SnykCodeConsistentIgnores` feature flag is not enabled in the test, so `zeroIssuesText` never reaches the IVO branch and `"Open and Ignored issues are disabled!"` can never appear regardless of which code path runs. Add `Override(featureflag.SnykCodeConsistentIgnores, true)` to actually exercise the fix. (inline comment posted)

2. **`FilterState.IssueViewOptions` still reads global config** (`tree_scan_emitter.go:129`, `get_tree_view.go:61`) — The tree HTML body now uses folder-level IVO, but `FilterState.IssueViewOptions` embedded in the same payload still reads `config.GetIssueViewOptions(e.conf)`. In a multi-folder workspace with folder overrides, the filter-state metadata disagrees with the rendered tree content. If this is intentional (filter panel = global, info nodes = per-folder), add a comment documenting the policy.

### Suggestions
3. **`FolderConfigReadOnly()` called twice per folder** (`tree_builder.go:111,113`) — `f.IssueViewOptions()` internally calls `FolderConfigReadOnly()`, then line 113 calls it again for `ConsistentIgnoresEnabled`. Extract one call and reuse, consistent with the `DisplayableIssueTypes` / `IsDeltaFindingsEnabled` pattern.

4. **`toggleTreeFilter` writes global IVO while tree reads folder-level** — Users clicking the Open/Ignored toggle modify global IVO, which may have no effect on folder-level overrides the tree now reads. Either document this as intentional scope separation or track as a follow-on.

### Security
Clean. Pre-existing SCA vulns in test/benchmark deps — not introduced by this PR.

— AI review

domain/ide/treeview/tree_scan_emitter_test.go
nick-y-snyk: Fixed. Added `SetupWorkspaceWithFeatureFlags` to workspaceutil and updated the test to override `SnykCodeConsistentIgnores=true` before folder creation. The test now actually exercises the IVO branch — with global IVO disabled but folder-level enabled, the disabled message would appear if the bug were present.

domain/ide/treeview/tree_builder.go
```diff
@@ -107,16 +101,17 @@ func (b *TreeBuilder) BuildTree(workspace types.Workspace) TreeViewData {
...
+		cfg := f.FolderConfigReadOnly()
 		fd := FolderData{
 			FolderPath:          f.Path(),
 			FolderName:          f.Name(),
 			SupportedIssueTypes: supportedTypes,
 			AllIssues:           allIssues,
 			FilteredIssues:      filtered,
 			DeltaEnabled:        f.IsDeltaFindingsEnabled(),
-			IssueViewOptions:    b.issueViewOptions,
+			IssueViewOptions:    f.IssueViewOptions(),
```
bastiandoetsch: [Suggestion] `BuildTree` re-reads the folder config several times per folder: `cfg := f.FolderConfigReadOnly()` is computed just above, yet `f.IssueViewOptions()` here (and `f.IsDeltaFindingsEnabled()` / `f.DisplayableIssueTypes()`) each call `FolderConfigReadOnly()` again internally, re-running `GetUnenrichedFolderConfigFromEngine`. The commit message says this "dedupes" the call, but it actually adds one and discards the `cfg` resolved on the line above. Three of four reviewers flagged this.

Fix: thread the already-resolved `cfg` through, e.g. add an unexported `f.issueViewOptionsForFolder(cfg)` and pass `cfg` here instead of calling the public accessor:

```go
cfg := f.FolderConfigReadOnly()
// ...
IssueViewOptions: f.issueViewOptionsForFolder(cfg),
```

Low impact (few folders, render is debounced) and the multi-read pattern pre-existed this PR, so non-blocking — the main point is that the "dedupe" claim does not hold.

— AI review

domain/ide/treeview/tree_scan_emitter_test.go
```diff
@@ -207,3 +208,60 @@ func TestTreeScanStateEmitter_Dispose_StopsRenderLoop(t *testing.T) {
...
+	assert.Eventually(t, func() bool {
+		mu.Lock()
+		defer mu.Unlock()
+		return receivedPayload != nil
+	}, 2*time.Second, 50*time.Millisecond)
+
+	mu.Lock()
+	treeView := receivedPayload.(types.TreeView)
+	mu.Unlock()
+	assert.NotContains(t, treeView.TreeViewHtml, "Open and Ignored issues are disabled!",
```
bastiandoetsch, line 271: [Suggestion] This regression test now correctly exercises the IVO branch (thanks for the earlier fix), but it only asserts the negative override case. A regression that made `IssueViewOptions()` always return `{true, true}` would still pass, and `assert.NotContains` after waiting only for a non-nil payload can pass against an interim/empty tree.

Fix: also pin the global-fallback case and assert a positive marker so the absence is meaningful:

```go
// no folder override + global both-disabled => the disabled message MUST appear
assert.Contains(t, treeView.TreeViewHtml, "Open and Ignored issues are disabled!")
```

(in a sibling test or a second folder), plus assert the congrats/product node is present in this test so the `NotContains` is not vacuous.

— AI review

domain/ide/workspace/folder.go
```diff
@@ -925,6 +925,11 @@ func (f *Folder) DisplayableIssueTypes() map[product.FilterableIssueType]bool {
 	return f.displayableIssueTypesForFolder(f.FolderConfigReadOnly())
 }
 
+// IssueViewOptions returns the issue view options for this folder, respecting folder-level overrides.
+func (f *Folder) IssueViewOptions() types.IssueViewOptions {
```
bastiandoetsch, line 941: [Suggestion] The new `IssueViewOptions()` method has no direct unit coverage, and its nil-`FolderConfigReadOnly()` fallback path is untested. It is exercised end-to-end by the emitter test and its sibling `DisplayableIssueTypes()` is also untested, so this matches existing convention and is non-blocking. A small table test would document the global-base + folder-override contract and the nil-config fallback.

— AI review

bastiandoetsch (approved):
No blocking findings in this PR's diff — approving.

The fix is root-cause correct: the builder-held global `issueViewOptions` field and its setter are removed in favour of a per-folder resolver read, no orphaned callers remain, the `go.sum` cleanup is consistent with `go.mod` (v0.3.1), and the new regression test drives the real `Emit → BuildTree` path and genuinely fails on pre-fix code.

Three non-blocking Suggestions posted inline (redundant per-folder config reads vs. the "dedupe" claim, strengthening the new test's assertions, and direct unit coverage for `IssueViewOptions()`). Address in a follow-up if useful.

— AI review

domain/ide/treeview/tree_builder.go
nick-y-snyk: Fixed. Added `IssueViewOptionsFromConfig`, `IsDeltaFindingsEnabledFromConfig`, and `DisplayableIssueTypesFromConfig` to the Folder interface, implemented on `*Folder` delegating to the existing private helpers. `BuildTree` now calls `FolderConfigReadOnly()` once per folder and threads the result through all three field assignments.

domain/ide/treeview/tree_scan_emitter_test.go
nick-y-snyk: Fixed. Added a positive assertion (`assert.Contains` for the congrats text) to make the `NotContains` non-vacuous, and fixed the `Emit` call to pass a completed `ProductScanStates` snapshot so info nodes actually render. Added sibling test `TestTreeScanStateEmitter_GlobalDisabledIVO_ShowsDisabledMessage` asserting the disabled message DOES appear when global IVO is disabled with no folder override.

domain/ide/workspace/folder.go
nick-y-snyk: Added `TestFolder_IssueViewOptions` in `folder_test.go` covering three cases: nil cfg (falls back to global defaults), cfg with no folder override (follows global setting), and cfg with folder-level `UserFolderKey` override (folder values win).

PR #1236, opened 2026-04-29
fix: fallback to global org for LDX-Sync API call [IDE-1974]

rrama:
### Description

But only when `OrgSetByUser` is `true` and `PreferredOrg` is `""`.
When changing the global org in this scenario, also refresh from LDX-Sync for the applicable folder configs.
On init, handle the IDE config before fetching from LDX-Sync so we have the correct API endpoint, global org, etc.
Revert back to the global org just being stored in GAF, as the changes to use `UserGlobalKey` were getting out of hand for a scope we don't even support.
Do not send any diagnostics on initialisation while we're still half loading the IDE config in.
Make the FF service use the standard method for getting a folder's org, as the logic did not have the global org fallback logic correct.
Only fetch FFs and config from LDX-Sync on folder level org changes if LS is initialised. This prevents redundant API calls for the wrong org (plus, as mentioned above, changed to get all orgs after IDE configuration is loaded).

### Checklist

- [x] Tests added and all succeed
- [x] Regenerated mocks, etc. (`make generate`)
- [x] Linted (`make lint-fix`)
- [ ] README.md updated, if user-facing
 - N/A
- [ ] License file updated, if new 3rd-party dependency is introduced
 - N/A

___

### **PR Type**
Bug fix

___

### **Description**
- Trigger LDX-Sync refresh when global org changes.

- Ensure correct org resolution fallback logic.

- Refine organization caching and initialization order.

- Improve feature flag and SAST setting population.

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A[IDE/User sets Global Org] --> B{Global Org Changed?};
  B -- Yes --> C[Find Folders using Global Org Fallback];
  C --> D{Folders Found?};
  D -- Yes --> E[Trigger LDX-Sync Refresh for those Folders];
  B -- No --> F[No Change];
  D -- No --> F;
  A --> G[Initialize IDE];
  G --> H[Apply Settings];
  H --> I[Prime Global Org Resolver];
  I --> J[Call LDX-Sync Refresh];
  J --> K[Populate Feature Flags/SAST Settings];
```

<details> <summary><h3> File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Bug fix</strong></td><td><details><summary>3 files</summary><table>
<tr>
  <td><strong>config.go</strong><dd><code>Optimize SetOrganization to prevent redundant resolutions</code></dd></td>
[...]

PR #1337, opened 2026-06-10
refactor: panic on feature flag access without ConfigResolver [IDE-1974]

rrama:
### Description

Remove silent false/no-op when FolderConfig has no ConfigResolver in GetFeatureFlag and SetFeatureFlag.
Wire ConfigResolver in tests that were constructing incomplete FolderConfigs and not exercising flag reads.

Safe in prod code due to everything going through `wireConfigResolver`.
I did not do the rest of the functions in the file, this was enough for now.

FYI, this change is unrelated to the ticket, just the silent fail caused me some issues when I was adding tests for #1236.

### Checklist

 - N/A
 - N/A
 - N/A

### DIFF
--- a/internal/types/folder_config.go
+++ b/internal/types/folder_config.go
@@ -145,7 +145,7 @@ func (fc *FolderConfig) SetFeatureFlag(flag string, value bool) {
 }
 
 // Conf returns the configuration for prefix key access.
-// Delegates to ConfigResolver.Configuration() when available.
+// Returns nil if the ConfigResolver is nil.
 func (fc *FolderConfig) Conf() configuration.Configuration {
 	if fc.ConfigResolver != nil {
 		return fc.ConfigResolver.Configuration()
--- a/internal/types/folder_config.go
+++ b/internal/types/folder_config.go
@@ -128,12 +128,8 @@ func (fc *FolderConfig) GetFeatureFlag(flag string) bool {
 	if fc == nil {
 		return false
 	}
-	conf := fc.Conf()
-	if conf == nil {
-		return false
-	}
 	key := configresolver.FolderMetadataKey(string(PathKey(fc.FolderPath)), FeatureFlagPrefix+flag)
-	return conf.GetBool(key)
+	return fc.Conf().GetBool(key)
 }
 
 // SetFeatureFlag writes a feature flag value to configuration under the folder metadata prefix.
@@ -144,12 +140,8 @@ func (fc *FolderConfig) SetFeatureFlag(flag string, value bool) {
 	if fc == nil {
 		return
 	}
-	conf := fc.Conf()
-	if conf == nil {
-		return
-	}
 	key := configresolver.FolderMetadataKey(string(PathKey(fc.FolderPath)), FeatureFlagPrefix+flag)
-	conf.Set(key, value)
+	fc.Conf().Set(key, value)
 }
 
 // Conf returns the configuration for prefix key access.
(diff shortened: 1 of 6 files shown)


internal/types/folder_config.go
```diff
@@ -144,12 +140,8 @@ func (fc *FolderConfig) SetFeatureFlag(flag string, value bool) {
 	if fc == nil {
```
bastiandoetsch, line 140: [Should Fix] `SetFeatureFlag` was changed symmetrically with `GetFeatureFlag` to panic on nil `ConfigResolver`, but the test suite only adds a panic assertion for `GetFeatureFlag`. `SetFeatureFlag` has no corresponding `panics when ConfigResolver is nil` test case.

Fix: add a parallel test subcase in `TestFolderConfig_SetFeatureFlag` (or alongside the existing nil-receiver test):

```go
t.Run("panics when ConfigResolver is nil", func(t \*testing.T) {
    fc := &FolderConfig{}
    assert.Panics(t, func() { fc.SetFeatureFlag("myFlag", true) })
})
```

— AI review

bastiandoetsch:
> ℹ️ This finding is in files not changed by this PR. It is non-blocking for this review — shared for awareness.

[Should Fix · non-blocking · ⚠ CONTRADICTED — Agents A and D verified safety] Several test files in `domain/snyk/scanner/` and `domain/snyk/scanner/scanner_precedence_test.go` construct `FolderConfig{FolderPath: ...}` without a `ConfigResolver` field. With the nil guard removed, any code path that calls `GetFeatureFlag`/`SetFeatureFlag` on these instances will now panic rather than return false.

Agents A and D independently verified that the current scan logic does NOT call `GetFeatureFlag`/`SetFeatureFlag` on these instances — so no immediate panic risk. However, if any future code in the scan path calls these methods, the tests will fail loudly with a panic rather than a clear error message.

Fix: Either document that these test instances intentionally never access feature flags, or proactively add `ConfigResolver: testutil.DefaultConfigResolver(engine)` to the relevant `FolderConfig` literals in scanner tests to future-proof them.

— AI review

bastiandoetsch:
> ℹ️ This finding is in a file changed by this PR but at a line outside the diff. It is non-blocking for this review — shared for awareness.

[Suggestion · non-blocking] `FolderConfig.Conf()` is a public method (58 call sites) that now has an implicit panic precondition: if `ConfigResolver` is nil, calling `Conf().GetBool(...)` panics. The method's docstring does not warn callers of this precondition.

Two reviewers (semantic-analysis and adversarial) independently flagged this.

Fix: Add a docstring to `Conf()` at `internal/types/folder_config.go`:
```go
// Conf returns the configuration for prefix key access.
// Panics if ConfigResolver is nil — ensure ConfigResolver is initialized before calling.
func (fc \*FolderConfig) Conf() configuration.Configuration {
```

— AI review

internal/types/folder_config.go
rrama: Just the one test is more than enough. Tbh I wouldn't have even done one if it wasn't already there. If someone changes it in the future we should catch it in review as silently failing is not a good thing.

rrama:
> Fix: Either document that these test instances intentionally never access feature flags, or proactively add ConfigResolver: testutil.DefaultConfigResolver(engine) to the relevant FolderConfig literals in scanner tests to future-proof them.

No, if someone makes them call FFs the test will fail and they will catch the issue, maybe it will reveal the test is taking a different path they didn't want.


> Fix: Add a docstring to Conf() at internal/types/folder_config.go:
>
> ```go
> // Conf returns the configuration for prefix key access.
> // Panics if ConfigResolver is nil — ensure ConfigResolver is initialized before calling.
> func (fc \*FolderConfig) Conf() configuration.Configuration {
> ```

Sure fine, the old docstring was misleading. Done.

bastiandoetsch (approved):

PR #1338, opened 2026-06-11
Feat/autofix/explain migration

BBerabi:
### Description

Query AI Explain only when the backend did not return any explanations together with the fix.

### Checklist


rrama:
Superseded by #1414, which removed the EnrichWithExplain flow entirely (code-client-go now returns the explanation inline).

PR #1340, opened 2026-06-15
feat(config-dialog): add additional_parameters and additional_environment to Project Defaults tab [IDE-2110]

nick-y-snyk:
## Summary
image

- **OSS scanner** (`d9ee030`): merge folder-level `additional_environment` into the CLI subprocess env map additive-on-top of the global env (folder wins on key conflict, all other keys survive).
- **HTML settings dialog** (`c2e4f867`): add collapsible Advanced section to the Project Defaults tab with global `additional_parameters` and `additional_environment` inputs, info-box explaining values combine with per-project entries, and format validation for the env field.
- **Env persistence** (`c2e4f867`): `applyEnvironment` now also writes the raw string to `UserGlobalKey` so the dialog can repopulate the field on reopen.
- **CSS fix** (`a6a43854`): extend `.tab-pane > .section` gutter rule so the Advanced block aligns with the two section cards above it.

Merge semantics:
- CLI params: global default + per-project appended (existing behaviour via `SettingCliAdditionalOssParameters` resolver fallback).
- Env vars: global default applied via `os.Setenv`; per-project merged on top at scan time — per-project wins on key conflict.

Security: `additional_parameters` / `additional_environment` are intentionally **not** LDX-Sync routed (IDE settings flow only, per ticket requirement).

UX copy in the info-box is a placeholder pending Tars finalizing wording (IDE-2110 comment 2026-06-05).

## Test plan

- [ ] All unit tests pass (`make test`)
- [ ] Project Defaults tab shows collapsible Advanced section with both fields
- [ ] Field values persist after saving and reopening the dialog
- [ ] Valid env format (`VAR=val;VAR2=val2`) saves cleanly; invalid format shows error and blocks save
- [ ] OSS scan picks up global env vars set in Project Defaults
- [ ] Per-project env overrides global env on key conflict; per-project params append to global params
- [ ] Eclipse PATH handling unaffected (no change to `applyEnvironment` `os.Setenv` path)
- [ ] LDX-Sync does not route these settings (verify `ldx_sync_adapter.go` mapping unchanged)


application/server/configuration.go
```diff
@@ -1118,6 +1118,10 @@ func applyEnvironment(conf configuration.Configuration, logger *zerolog.Logger,
 			logger.Err(err).Msgf("couldn't set env variable %s", envVar)
 		}
 	}
+	// Persist the raw string so the settings dialog can repopulate this field on reopen.
+	// os.Setenv alone is not readable back from config; SetGlobalUser writes to UserGlobalKey
+	// which r.GetString(SettingAdditionalEnvironment, nil) resolves via the folder-scope chain.
+	types.SetGlobalUser(conf, types.SettingAdditionalEnvironment, v)
```
acke, line 1146: 🔴 **Critical**

**[application/server/configuration.go:1107]** `applyEnvironment` early-returns on empty value (`if !ok || v == ""`). When the user clears the Project Defaults env field and saves, the form sends `{Value: "", Changed: true}`, `settingStr` returns `("", true)`, and the function returns before reaching this new `SetGlobalUser` line. Result: `UserGlobalKey` retains the old string, the dialog repopulates the stale value on reopen, and previously `os.Setenv`-ed vars stay live in the process env. No UI path to clear env vars once saved.

**Fix:**
Split the empty path from the parse path. On `Changed && v == ""`: call `SetGlobalUser(conf, SettingAdditionalEnvironment, "")` to clear persistence (and `os.Unsetenv` previously-applied keys — see the comment on line 1118). On `v != ""`: keep the current parse/Setenv loop plus this new `SetGlobalUser` write. Add `TestApplyEnvironment_ClearsPersistedValue` that pre-seeds `UserGlobalKey(additional_environment)`, calls `applyEnvironment` with `{Value: "", Changed: true}`, and asserts `GetGlobalString` returns `""`.

— AI review comment

application/server/configuration.go
```diff
@@ -1118,6 +1118,10 @@ func applyEnvironment(conf configuration.Configuration, logger *zerolog.Logger,
 			logger.Err(err).Msgf("couldn't set env variable %s", envVar)
 		}
 	}
+	// Persist the raw string so the settings dialog can repopulate this field on reopen.
```
acke, line 1142: 🟠 **High**

**[application/server/configuration.go:1110-1120]** `os.Setenv` is one-way — keys removed from the saved string are never unset from the process env. Sequence: save `A=1;B=2`, then re-save `A=1` → `B=2` stays in `os.Environ()` and leaks into every subsequent CLI subprocess (because `updateSDKs` returns `gotenv.Env` seeded from process env via `UpdateEnvironmentAndReturnAdditionalParams`). The new Project Defaults UI implies "remove env override" works, but it silently does not.

**Fix:**
Before mutating, read `GetGlobalString(conf, SettingAdditionalEnvironment)` (the previously-persisted value, now available thanks to the new `SetGlobalUser` write below), diff key sets, and `os.Unsetenv(k)` for keys present in old-but-not-new. Add a test that calls `applyEnvironment` twice (`A=1;B=2` then `A=1`) and asserts `B` is absent from `os.Environ`.

— AI review comment

application/server/configuration_test.go
```diff
@@ -2880,3 +2880,26 @@ func Test_ApplyOrganization_LDXSyncRefreshesForGlobalOrgFallback(t *testing.T) {
 		})
 	}
 }
+
+func TestApplyEnvironment_PersistsGlobalUser(t *testing.T) {
```
acke, line 2884: 🟡 **Medium**

**[application/server/configuration_test.go:2882-2905]** `TestApplyEnvironment_PersistsGlobalUser` covers only the happy "set" path. It does not exercise (a) empty-value clear (see Critical above), (b) consecutive applies that drop keys (see High above), or (c) a malformed pair that `applyEnvironment` skips via `len(parts) != 2`. The new behavior is the contract; without these cases regressions land silently.

**Fix:**
Add three subtests: (1) clear-path — seed prior `UserGlobalKey`, call with `Value: ""`, assert `GetGlobalString == ""`; (2) sequential-apply — call `A=1;B=2` then `A=1`, assert `B` absent from `os.Environ`; (3) malformed-entry — pass `GOOD=1;BAD;OK=2`, assert `GOOD` and `OK` set, `BAD` absent.

— AI review comment

application/server/configuration_test.go
```diff
@@ -2880,3 +2880,26 @@ func Test_ApplyOrganization_LDXSyncRefreshesForGlobalOrgFallback(t *testing.T) {
...
+
+	// os.Setenv side effect
+	assert.Equal(t, "hello", os.Getenv("SNYK_TEST_IDE2110_X"))
+	assert.Equal(t, "world", os.Getenv("SNYK_TEST_IDE2110_Y"))
+
+	// Persisted to config for dialog pre-population
+	assert.Equal(t, "SNYK_TEST_IDE2110_X=hello;SNYK_TEST_IDE2110_Y=world", types.GetGlobalString(conf, types.SettingAdditionalEnvironment))
+
+	// Cleanup SNYK_TEST_IDE2110_Y (only X is tracked by t.Setenv)
+	t.Cleanup(func() { _ = os.Unsetenv("SNYK_TEST_IDE2110_Y") })
```
acke: 🔵 **Low**

**[application/server/configuration_test.go:2904]** `t.Cleanup` is registered after assertions. It works (cleanups queued at registration time always run), but reads as "afterthought" — easy to drop on a future refactor that reorders blocks.

**Fix:**
Either move this `t.Cleanup` up next to the `t.Setenv` near the top of the test, or just use `t.Setenv("SNYK_TEST_IDE2110_Y", "")` at the top — `applyEnvironment` will overwrite it via `os.Setenv` and `t.Setenv` will restore the empty value on cleanup, no manual `Unsetenv` needed.

— AI review comment

infrastructure/configuration/template/js/features/validation.js
```diff
@@ -275,6 +275,20 @@
 		}
 	};
 
+	// Validate global (Project Defaults) additional environment variables on input
+	validation.validateGlobalAdditionalEnvOnInput = function() {
+		validation.validateAndShowError("additional_environment", "additional_environment-error", validation.validateAdditionalEnv);
```
acke, line 282: 🟠 **High**

**[infrastructure/configuration/template/js/features/validation.js:143]** This new global validator reuses `validation.validateAdditionalEnv`, whose regex `^\s*[A-Za-z_][A-Za-z0-9_]*\s*=\s*[^;=]*\s*$` rejects any VALUE containing `=`. The Go backend uses `strings.Cut(pair, "=")` (`infrastructure/oss/cli_scanner.go:348`) and the new test `TestCLIScanner_updateArgs_folderAdditionalEnv/value containing equals sign preserved` explicitly asserts `MY_KEY=val=with=equals` is preserved. The frontend will block legitimate inputs the runtime is built to accept: base64-padded values (`X=Zm9v==`), JWTs, query strings, anything with `=` in the value.

**Fix:**
Relax the VALUE character class in `validateAdditionalEnv` from `[^;=]*` to `[^;]*` so it matches `strings.Cut` semantics (split on first `=` only). Update the inline error message if it implies the stricter rule, and add JS test cases for `MY_KEY=val=with=equals` and `B64=Zm9vCg==`.

— AI review comment

infrastructure/configuration/template/js/features/validation.js
```diff
@@ -275,6 +275,20 @@
...
+	};
+
+	// Initialize validation event listener for the global additional environment field
+	validation.initializeGlobalAdditionalEnvValidation = function() {
```
acke, line 286: 🟡 **Medium**

**[infrastructure/configuration/template/js/features/validation.js:209-227]** `validateAllFolderAdditionalEnv` iterates `[id^="folder_"][id$="_additional_environment"]` — the new global field (id `additional_environment`, no `folder_` prefix) is not in that pass. The new `initializeGlobalAdditionalEnvValidation` wires an `input` listener but never seeds `validationState["additional_environment"]`, so a user who opens the dialog with a server-returned malformed value and clicks Save without touching the field bypasses the submit-time gate.

**Fix:**
Either (a) add a `validateGlobalAdditionalEnv()` analogue and call it from the submit-gate path, or (b) run one validation call on dialog init for the global field so the state slot exists. Include the global key in `getFormValidationInfo`.

— AI review comment

infrastructure/oss/cli_scanner.go
```diff
@@ -342,6 +342,14 @@ func (cliScanner *CLIScanner) updateArgs(workDir types.FilePath, commandLineArgs
 
 	// this asks the client for the current SDK and blocks on it
 	additionalParameters, env := cliScanner.updateSDKs(folderConfig.FolderPath)
+	// merge folder-level additional_environment into env map (additive: folder wins on key conflict)
+	if ae := folderConfig.AdditionalEnv(); ae != "" {
```
acke, line 352: 🟡 **Medium**

**[infrastructure/oss/cli_scanner.go:345-352]** This merge loop reads `folderConfig.AdditionalEnv()` which `getUserString` resolves from `UserFolderKey(fp, additional_environment)` **only** — no fallback to `UserGlobalKey` (see `internal/types/folder_config_helpers.go:85-108`). The PR's documented merge semantics ("per-project overrides global on key conflict") work only because the global value reaches `env` via `applyEnvironment → os.Setenv → os.Environ() → updateSDKs(env)`. That coupling is implicit and brittle: if `UpdateEnvironmentAndReturnAdditionalParams` ever stops seeding from process env, the global tier silently drops out with no test catching it.

**Fix:**
Either (a) add an inline comment on this block explicitly stating the `os.Setenv → os.Environ → updateSDKs(env)` contract for the global tier; or (b) read `GetGlobalString(conf, SettingAdditionalEnvironment)` directly here and merge it before the folder layer (same single-entry pattern as `cli_additional_oss_parameters`). Add a scan-level test that boots with a global value set but no folder override and asserts the env reaches the CLI invocation.

— AI review comment

acke:
ℹ️ **Info**

UX copy in the new Advanced section info-box is a placeholder pending Tars (per the IDE-2110 comment dated 2026-06-05). The three snapshot files in `scripts/config-dialog/config_output_*.html` lock in this placeholder string and will need regeneration when copy is finalized.

**Fix:**
Open a follow-up linked to IDE-2110 to refresh the copy and regenerate the three snapshots when the final copy is approved.

— AI review comment

application/server/configuration_test.go
acke: Also, I think we do not want to have references to IDE ticket in the code/code comments, so do we want to remove IDE2110?

application/server/configuration.go
nick-y-snyk: Fixed. `applyEnvironment` now guards on `!ok` only — `settingStr` already returns `ok=false` when `Changed` is false, so an empty-but-changed value falls through to the persistence write. On clear, `SetGlobalUser(..., "")` runs, the dialog repopulates blank, and previously-applied keys are unset (see High reply). Added `TestApplyEnvironment_ClearsPersistedValue`.

application/server/configuration.go
nick-y-snyk: Fixed. Before applying, we read the previously-persisted `GetGlobalString(SettingAdditionalEnvironment)`, diff key sets, and `os.Unsetenv` keys present in old-but-not-new. Added `TestApplyEnvironment_SequentialApplyUnsetsDroppedKeys` (`A=1;B=2` then `A=1`, asserts `B` absent from the process env).

application/server/configuration_test.go
nick-y-snyk: Done. Added: `TestApplyEnvironment_ClearsPersistedValue`, `TestApplyEnvironment_SequentialApplyUnsetsDroppedKeys`, `TestApplyEnvironment_SkipsMalformedEntries` (`GOOD=1;BAD;OK=2`), plus `TestApplyEnvironment_UnchangedFieldNoOp` and `TestApplyEnvironment_ValuePreservesEquals` for the `=`-in-value path.

application/server/configuration_test.go
nick-y-snyk: Done. Replaced the trailing `t.Cleanup` with `t.Setenv(key, "")` at the top of each test for every key the test touches — `applyEnvironment` overwrites via `os.Setenv` and `t.Setenv` restores on cleanup, so no manual `Unsetenv` remains.

infrastructure/configuration/template/js/features/validation.js
nick-y-snyk: Fixed. Relaxed the VALUE class from `[^;=]*` to `[^;]*` so it matches the Go backend `strings.Cut`/`SplitN(..., "=", 2)` semantics (split on first `=` only). Updated the comment, regenerated the three `config_output_*.html` snapshots, and added `js-tests/validation-additional-env.test.mjs` with cases for `MY_KEY=val=with=equals` and `B64=Zm9v==`.

infrastructure/configuration/template/js/features/validation.js
nick-y-snyk: Left as-is to match the existing convention — no field seeds `validationState` on init; token, api_endpoint, risk_score and cli_version all populate their slot only on the first `input`/`change` event, and `validateAllFolderAdditionalEnv` is never wired into the submit path. The global field now behaves identically. Closing this gate uniformly across all fields is a separate change; happy to file a follow-up if you prefer to track it.

infrastructure/oss/cli_scanner.go
nick-y-snyk: Went with option (a): added an inline comment on the merge block documenting the `applyEnvironment -> os.Setenv -> os.Environ() -> updateSDKs(env)` contract for the global tier and the global < folder precedence. Avoids the double-apply risk of reading `GetGlobalString` here as well.

application/server/configuration_test.go
nick-y-snyk: Good catch — removed. Renamed the test env vars from `SNYK_TEST_IDE2110_*` to `SNYK_TEST_ADDITIONAL_ENV_*`. No ticket references remain in code or comments.

acke (approved):

PR #1341, opened 2026-06-16
feat: per-project user-override indicator on HTML settings page [IDE-1994]

acke:
## Summary
Render a bust/silhouette icon (👤) next to settings whose value comes from a user override, mirroring the existing 🏢 building icon used for org-set values. Restores functionality that was removed in #1185 (UX/code-quality concerns). Propagation logic intentionally not pulled in — that is its own task.

UX agreed with Tars on 2026-06-05 (per Ben Durrans): icon-only, no border.

Jira: https://snyksec.atlassian.net/browse/IDE-1994

## Scope
- **Per-project tab**: 👤 appears for any setting whose effective source is `user-override` (folder-scope override of the project default).
- **Project Defaults tab**: 👤 appears for any setting whose effective source resolves to a user-global value (i.e. you've explicitly set the project-level default). Implemented by delegating to `ConfigResolver.GetEffectiveValue` with no folder context and remapping `global` → `user-override` at that scope only. Folder-pane semantics are unchanged.

The two icons are independent — toggling a setting on the per-project tab writes `UserFolderKey` only; toggling on Project Defaults writes `UserGlobalKey`. Each scope's indicator reflects its own writes (per [IDE-1994] scope; propagation is a separate task).

## Implementation
- `infrastructure/configuration/config_html.go`
  - Add `user-override` case to `sourceToClass` (→ `source-user-override`) and `tmplSourceIndicator` (👤 + tooltip "Overridden at project level").
  - Replace the ad-hoc `UserGlobalKey`-only `computeProjectDefaultScopes` with a delegation to `ConfigResolver.GetEffectiveValue(name, nil)` so precedence matches the rest of the LS (`Locked LDX-Sync > User Override > LDX-Sync > Default`).
  - Remap `global` → `user-override` only at the Project Defaults scope.
  - Thread `ConfigResolverInterface` through `NewConfigHtmlRenderer` so the renderer can resolve without a folder context.
- `infrastructure/configuration/template/config.html`: add `{{sourceIndicator .DefaultsScopes ...}}` calls across the Project Defaults section (14 settings).
- `domain/ide/command/configuration_command.go`, `scripts/config-dialog/main.go`: pass the resolver into the renderer.
- `infrastructure/configuration/config_html_test.go`: cover the `user-override` source case in `tmplSourceIndicator`, assert 👤 in integration HTML, update all call sites.
- Regenerate `scripts/config-dialog/*.html` fixtures.

## Test plan
- [x] Unit tests cover the `user-override` source case in `tmplSourceIndicator` and `sourceToClass`.
- [x] Integration test (`TestConfigHtmlRenderer_SourceIndicatorsInOutput`) asserts 👤 + "Overridden at project level" tooltip render in the rendered HTML.
- [x] Generated HTML fixtures regenerated via `make generate`.
- [x] Verified locally with a debug CLI build (snyk-ls replaced into cliv2-private) — Project Defaults tab renders 👤 next to user-overridden settings; per-project tab renders 👤 only where the folder overrides the project default; 🏢 / 🏢🔒 indicators unchanged for org-set / org-locked settings.

[IDE-1994]: https://snyksec.atlassian.net/browse/IDE-1994?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


infrastructure/configuration/template/config.html
```diff
@@ -257,7 +261,7 @@ <h3>Scanners</h3>
...
-                                    <label for="organization"><span data-toggle="tooltip" title="<p>Specify the organization (ID or name) for Snyk to run scans against.</p><p>Organization selection follows this order:</p><ol><li>Project-specific settings (if configured)</li><li>This global setting (if the project-specific setting is empty)</li><li>Your web account's preferred organization (if both above are empty)</li></ol>">Organization</span> <span class="config-scope-slot" data-config-scope-slot="true" data-setting-key="organization"></span></label>
+                                    <label for="organization"><span data-toggle="tooltip" title="<p>Specify the organization (ID or name) for Snyk to run scans against.</p><p>Organization selection follows this order:</p><ol><li>Project-specific settings (if configured)</li><li>This global setting (if the project-specific setting is empty)</li><li>Your web account's preferred organization (if both above are empty)</li></ol>">Organization</span> <span class="config-scope-slot" data-config-scope-slot="true" data-setting-key="organization"></span> {{sourceIndicator .DefaultsScopes "organization"}}</label>
```
rrama, line 264: We may not want it for org, I will have to check with Tars. Not going to block the PR on this. I just hadn't considered it until now.

infrastructure/configuration/config_html.go
```diff
@@ -190,12 +192,14 @@ func tmplSourceIndicator(effectiveConfig map[string]types.EffectiveValue, settin
 		return template.HTML(`<span class="source-indicator" data-toggle="tooltip" title="Locked due to organization settings">🏢🔒</span>`)
 	case "ldx-sync":
 		return template.HTML(`<span class="source-indicator" data-toggle="tooltip" title="Set by your organization settings">🏢</span>`)
+	case "user-override":
+		return template.HTML(`<span class="source-indicator" data-toggle="tooltip" title="Overridden at project level">👤</span>`)
```
rrama: We need a different text the Project defaults level vs the per-project level.

Currently Tars is thinking:
"set by you at project level"
and
"set by you at [project] defaults level" - he is unsure about the word "project" here.

infrastructure/configuration/config_html.go
```diff
@@ -190,12 +192,14 @@ func tmplSourceIndicator(effectiveConfig map[string]types.EffectiveValue, settin
 		return template.HTML(`<span class="source-indicator" data-toggle="tooltip" title="Locked due to organization settings">🏢🔒</span>`)
```
rrama, line 193: This has reminded me, we are meant to drop the padlock icon and the user should rely on the input field being disabled. Not a prio. for today though.

infrastructure/configuration/config_html.go
```diff
@@ -243,23 +252,11 @@ func computeProjectDefaultScopes(engine workflow.Engine) map[string]types.Effect
...
-		} else {
-			value = gafConf.Get(settingName)
-		}
-
-		scopes[settingName] = types.EffectiveValue{
-			Value:  value,
-			Source: source,
+		ev := resolver.GetEffectiveValue(settingName, nil)
+		if ev.Source == "global" {
+			ev.Source = "user-override"
```
rrama: As mentioned above, we will need to differentiate between these for what tooltip is displayed (we may also do different icons, but that is for a future UX polish).

rrama (approved):
Only a small change needed so we have a different tooltip if it is a user set at global vs folder.

infrastructure/configuration/config_html.go
acke: Added a `user-override-defaults` source sentinel: `computeProjectDefaultScopes` now remaps `global` → `user-override-defaults` instead of `user-override`, and `tmplSourceIndicator` renders distinct tooltips — "Set by you at project level" for per-project user overrides and "Set by you at project defaults level" for the defaults scope.

infrastructure/configuration/config_html.go
acke: Differentiated — remap target here is now `user-override-defaults` (separate from per-project `user-override`), and `tmplSourceIndicator` renders a different tooltip for each. Icon kept the same (👤) as you noted icon variation is future polish.

acke:
Re: https://github.com/snyk/snyk-ls/pull/1341#pullrequestreview-4516791668

Tooltip differentiation landed — new `user-override-defaults` source sentinel drives "Set by you at project defaults level" tooltip in the defaults section, while per-project keeps "Set by you at project level". See replies on #discussion_r3429163998 and #discussion_r3429297037.

PR #1342, opened 2026-06-16
fix: handle xml property updates when underlying code has changed [IDE-2139]

andrewrobinsonhodges-snyk:
### Description

Two fixes:

1) Maven property-managed versions

When a dependency's <version> is a ${property} reference, the quickfix now edits the matching <properties> entry instead of hardcoding the version into the dependency block.

- ast/ast.go — added Properties map[string]*Node to ast.Tree.
- ast/maven/parser.go — the parser now parses the <properties> block, recording a node per property with a range pointing at the property value (new addPropertyNodes / newValueNode helpers; Properties map initialized in initTree).
- infrastructure/oss/code_actions.go — GetCodeActions detects a ${…} version (new resolveMavenPropertyNode, which also walks up parent POMs) and redirects the edit's file + range to the property declaration.

2) re-apply safety

The deferred edit now snapshots the exact text it intends to replace at creation time and re-validates it against the current file content at apply time; if the text no longer matches (or the range is now out of bounds), it refuses to apply a stale edit instead of corrupting the file.

- infrastructure/oss/code_actions.go — AddQuickFixAction gained an originalContent snapshot param and a textAtRange guard in the deferred callback. Also fixed a latent bug where the callback mutated its captured fileContent.

### Checklist

___

### **PR Type**
Enhancement, Bug fix, Tests

___

### **Description**
- Enhance Maven POM parsing with safety checks.

- Fix Maven property version updates for quick fixes.

- Add comprehensive tests for new features.

___

### Diagram Walkthrough

```mermaid
flowchart LR
  subgraph New Parser Features
    A[Maven POM Parsing] --> B(Properties Block Parsing);
    B --> C{Property Node Creation};
    A --> D(Parent POM Resolution);
    D --> E{Depth, Cycle, Size Guards};
    D --> F{Workspace Root Containment};
  end
  subgraph Quick Fix Enhancements
    G[Code Actions] --> H(Maven Property Reference Resolution);
    H --> I{Indirection Following};
    H --> J(Redirect Edit to Property);
    H --> K(Fallback to Dependency Edit);
    G --> L(Stale Edit Refusal);
    L --> M(Snapshot Guarding);
    L --> N(Single-Shot Latch);
  end
  subgraph Testing
    O[Unit Tests] --> P(Parser Tests);
    O --> Q(Code Actions Tests);
  end
  C --> J;
  K --> M;
  I --> J;
  F --> D;
  E --> D;
  P --> B;
  Q --> H;
  Q --> L;
```

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><details><summary>7 files</summary><table>
<tr>
  <td><strong>ast.go</strong><dd><code>Add Properties map to AST Tree</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1342/changes#diff-9dcefbf65d58a909b6cc7332b7985d070382bf920a4b13c7acea4ee91ecdf5ac">+4/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>parser.go</strong><dd><code>Enhance Maven POM parsing with safety and property handling</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1342/changes#diff-3f3968f9c50ffc7835ab3acc38f1a4f5fc46976e37c4845890a090292e59343e">+234/-46</a></td>

</tr>

<tr>
  <td><strong>code_actions.go</strong><dd><code>Handle Maven property versions for quick fixes</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1342/changes#diff-80e9faf53fd666a41ad81dad2011a9832dd547fb9bfc6be87b2e22391e4644f6">+170/-10</a></td>

</tr>

<tr>
  <td><strong>issue.go</strong><dd><code>Pass workDir to getDependencyNode for Maven parent POMs</code>&nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1342/changes#diff-8a8ae474b46c292d5e918379849be57960c91e3168268e3f36d5b6fb3a2831ad">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>maven_range_finder.go</strong><dd><code>Add root path to Maven range finder</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1342/changes#diff-29dad00e9219ca8206b09e8cb1c10635205e4d415de5a998e98526a8d6b7b175">+4/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>range_finder.go</strong><dd><code>Add rootPath parameter to getDependencyNode</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1342/changes#diff-333a2cdfe328f6bb0eec7f43d48d8db6d417d555ac111bac0c9d8eb8a4ee98f5">+4/-3</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>unified_converter.go</strong><dd><code>Pass workDir to getDependencyNode for Maven</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1342/changes#diff-75cc7b3352e4a2f1077307fb167c6b96b9d1973c24f65b5dc762d396c566eb9a">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Tests</strong></td><td><details><summary>4 files</summary><table>
<tr>
  <td><strong>parser_test.go</strong><dd><code>Add tests for Maven property and parent POM parsing</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1342/changes#diff-56f8b9f576282024cef366a3d2024e23c393bc424bd1a7af8917142efcf913e0">+415/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>code_actions_test.go</strong><dd><code>Add tests for Maven property quickfix and stale edit handling</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1342/changes#diff-ed04a1f6ef2e54a563d01d98674570a9cf4e520743596a829d4cd7af7924c71e">+582/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>oss_test.go</strong><dd><code>Update getDependencyNode call in oss_test</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1342/changes#diff-18d44a59ba6614f20ee4e2bb69fe6801a5c680206aaba162efda262707b87721">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>range_finder_test.go</strong><dd><code>Update getDependencyNode call in range_finder_test</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1342/changes#diff-f689ad1eacda0a6e8773db6aec2a10e3b37cd946a822e7b001804d321d3b38bb">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

</details>

___


acke (commented):
_review withdrawn_

ast/maven/parser.go
```diff
@@ -72,54 +72,75 @@ func (p *Parser) Parse(content string, path types.FilePath) *ast.Tree {
...
+	if err := d.DecodeElement(&parentPOM, element); err != nil {
+		p.logger.Err(err).Msg("Couldn't decode Parent")
+		return
+	}
+
+	if parentPOM.RelativePath == "" {
+		parentPOM.RelativePath = filepath.Join("..", "pom.xml")
+	}
+
+	parentAbsPath, err := filepath.Abs(filepath.Join(pomDir, parentPOM.RelativePath))
```
acke, line 156: 🟠 **High**

**[ast/maven/parser.go:131]** Path traversal via Maven `<relativePath>`. `handleParent` concatenates attacker-controllable `<relativePath>` from the POM with `pomDir` via `filepath.Join` + `filepath.Abs` with no containment check. A crafted pom (`<relativePath>../../../../etc/passwd</relativePath>`) makes the parser read any file the IDE user can read.

**Fix:**
After `filepath.Abs`, require the result to live under a known workspace/folder root (e.g. `strings.HasPrefix(parentAbsPath, cleanedRoot+string(os.PathSeparator))`) and reject otherwise. Also check the file is a regular file (`fi.Mode().IsRegular()`). Add a test with a `../../../etc/passwd` relativePath that asserts the parent walk is refused.

— AI review comment

ast/maven/parser.go
```diff
@@ -72,54 +72,75 @@ func (p *Parser) Parse(content string, path types.FilePath) *ast.Tree {
...
+	if err != nil {
+		p.logger.Err(err).Msg("Couldn't resolve Parent path")
+		return
+	}
+	content, err := os.ReadFile(parentAbsPath)
+	if err != nil {
+		p.logger.Err(err).Msg("Couldn't read Parent file")
+		return
+	}
+	tree.ParentTree = p.Parse(string(content), types.FilePath(parentAbsPath))
```
acke: 🟠 **High**

**[ast/maven/parser.go:141]** Unbounded parent-POM recursion. `tree.ParentTree = p.Parse(...)` recurses on whatever path `<relativePath>` resolves to with no visited-set and no depth cap. Two POMs that point at each other (or one self-referencing POM) cause infinite recursion → stack overflow / DoS of the language server.

**Fix:**
Thread a visited-set (absolute paths already parsed in this chain) plus a hard depth limit (e.g. 32) through `Parse`. Bail with a logged warning when a cycle or the limit is hit. Cover with a test of two POMs that reference each other.

— AI review comment

infrastructure/oss/code_actions_test.go
```diff
@@ -0,0 +1,242 @@
...
+            <version>${lib.cyclonedx-core-java.version}</version>
+        </dependency>
+    </dependencies>
+</project>
+`
+
+// F1: when a Maven dependency version is declared via a ${property} reference,
+// the quickfix must edit the matching <properties> entry, not hardcode the
+// version into the dependency block.
+func Test_GetCodeActions_MavenPropertyVersion_RedirectsToProperty(t *testing.T) {
```
acke, line 57: 🟠 **High**

**[infrastructure/oss/code_actions_test.go:55]** Unresolved `${prop}` fallback path is uncovered. When the referenced property is absent from `tree.Properties` and there is no `ParentTree`, `resolveMavenPropertyNode` returns nil and the code falls back to editing the dependency `<version>` block. No test asserts the fallback produces a sensible edit instead of a nil-deref or a wrong target.

**Fix:**
Add a test where the pom has `<version>${missing.version}</version>` and no `<properties>` entry. Assert the quickfix is still produced, edit targets the dependency `<version>` range, no panic.

— AI review comment

acke, line 57: 🟠 **High**

**[infrastructure/oss/code_actions_test.go:55]** Parent-POM property resolution is completely uncovered. `resolveMavenPropertyNode` walks `tree.ParentTree`, but no test defines the property in a parent POM while referencing it from a child POM. A regression in the parent-walk loop would not be caught.

**Fix:**
Write a parent POM with `<properties><foo.version>9.0.5</foo.version></properties>` and a child POM whose dependency uses `${foo.version}`. Assert the quickfix edit targets the parent POM's path (`edit.Changes[parentPath]`), not the child's.

— AI review comment

ast/maven/parser.go
```diff
@@ -72,54 +72,75 @@ func (p *Parser) Parse(content string, path types.FilePath) *ast.Tree {
...
+
+func (p *Parser) handleParent(d *xml.Decoder, tree *ast.Tree, pomDir string, element *xml.StartElement) {
+	var parentPOM Parent
+	if err := d.DecodeElement(&parentPOM, element); err != nil {
+		p.logger.Err(err).Msg("Couldn't decode Parent")
+		return
+	}
+
+	if parentPOM.RelativePath == "" {
+		parentPOM.RelativePath = filepath.Join("..", "pom.xml")
```
acke, line 153: 🟡 **Medium**

**[ast/maven/parser.go:128]** No size cap on parent-POM read. `os.ReadFile(parentAbsPath)` reads the entire file with no limit. A malicious or accidental `<relativePath>` pointing at a large file (or `/dev/zero` on Linux) reads unbounded bytes into memory.

**Fix:**
`os.Stat` the resolved path first, require `IsRegular()` and `Size() < 16MB` (or similar). Alternatively wrap `os.Open` with `io.LimitReader`.

— AI review comment

ast/maven/parser.go
```diff
@@ -138,12 +159,86 @@ func (p *Parser) initTree(path types.FilePath, content string) *ast.Tree {
...
+
+		// The value spans from just after this start tag up to its end tag.
+		valueStartInInner := int(dec.InputOffset())
+		valueEndInInner := valueStartInInner
+		for {
+			inner, innerErr := dec.Token()
+			if inner == nil || innerErr != nil {
+				break
+			}
+			if _, isEnd := inner.(xml.EndElement); isEnd {
```
acke: 🟡 **Medium**

**[ast/maven/parser.go:198]** `addPropertyNodes` inner loop breaks on the first `xml.EndElement`. If a property value contains any nested element (e.g. `<v><x>1</x></v>`), the loop terminates at `</x>` rather than `</v>`. `valueEndInInner` then captures only the inner CharData and the recorded value range is wrong; on patterns like `<a><b>x</b><c>y</c></a>` the outer loop subsequently registers `<c>` as a top-level property.

**Fix:**
Track nesting depth — increment on `xml.StartElement`, decrement on `xml.EndElement`, only break when depth returns to zero. Cover with a property whose value contains a nested element.

— AI review comment

ast/maven/parser.go
```diff
@@ -138,12 +159,86 @@ func (p *Parser) initTree(path types.FilePath, content string) *ast.Tree {
...
+				break
+			}
+			if _, isCharData := inner.(xml.CharData); isCharData {
+				valueEndInInner = int(dec.InputOffset())
+			}
+		}
+
+		valueStartOffset := baseOffset + valueStartInInner
+		valueEndOffset := baseOffset + valueEndInInner
+		if valueStartOffset > valueEndOffset || valueEndOffset > len(content) {
```
acke: 🟡 **Medium**

**[ast/maven/parser.go:208]** Boundary guard is missing the lower bound. `valueStartOffset > valueEndOffset || valueEndOffset > len(content)` does not assert `valueStartOffset >= 0` or `valueStartOffset <= len(content)`; if `baseOffset` or `dec.InputOffset()` ever return an unexpected value, `content[valueStartOffset:valueEndOffset]` panics.

**Fix:**
Also guard `valueStartOffset < 0 || valueStartOffset > len(content)` before slicing.

— AI review comment

infrastructure/oss/code_actions.go
```diff
@@ -249,6 +302,48 @@ func getQuickfixEdit(engine workflow.Engine, affectedFilePath types.FilePath, up
...
+// resolveMavenPropertyNode returns the <properties> node a dependency version
+// refers to when the version is a property reference (e.g. ${foo.version}),
+// searching the current pom and walking up the parent pom hierarchy. It returns
+// nil when the version is not a property reference or the property cannot be
+// found.
+func resolveMavenPropertyNode(depNode *ast.Node) *ast.Node {
+	if depNode == nil {
+		return nil
+	}
+	matches := mavenPropertyRefRegexp.FindStringSubmatch(depNode.Value)
```
acke, line 366: 🟡 **Medium**

**[infrastructure/oss/code_actions.go:316]** `resolveMavenPropertyNode` does not handle property indirection. A property whose value is itself a `${other.version}` reference (legal in Maven) is treated as the final target and gets overwritten with the upgrade version, orphaning `other.version`.

**Fix:**
While the resolved node's `Value` still matches `mavenPropertyRefRegexp`, follow the chain to the final property (bounded depth, e.g. 16, to guard against cycles).

— AI review comment

infrastructure/oss/code_actions.go
```diff
@@ -170,17 +199,41 @@ func AddQuickFixAction(engine workflow.Engine, configResolver types.ConfigResolv
...
+
+	// Snapshot the text the edit is meant to replace, so the deferred edit can
+	// detect that the file has changed (e.g. the fix was already applied) and
+	// refuse to re-apply a stale, absolute-offset edit that would corrupt the
+	// file.
+	snapshot := originalContent
+	if snapshot == nil {
+		snapshot = fileContent
+	}
+	expectedText, haveExpectedText := textAtRange(snapshot, issueRange)
```
acke, line 231: 🟡 **Medium**

**[infrastructure/oss/code_actions.go:211]** Stale-edit guard silently disables itself. When `textAtRange(snapshot, issueRange)` returns `ok=false` (multi-line range, out-of-bounds, or snapshot nil because both `originalContent` and `fileContent` were nil), `haveExpectedText` is false and the apply-time guard is skipped entirely — exactly the cases where the snapshot is unreliable.

**Fix:**
If a snapshot was provided but `textAtRange` rejects it, treat that as a fatal pre-condition failure (return nil action, or log + drop) rather than silently disabling the guard.

— AI review comment

infrastructure/oss/code_actions.go
```diff
@@ -170,17 +199,41 @@ func AddQuickFixAction(engine workflow.Engine, configResolver types.ConfigResolv
...
+		if content == nil {
+			var err error
+			content, err = os.ReadFile(filePathString)
 			if err != nil {
 				logger.Error().Err(err).Str("file", filePathString).Msg("could not open file")
 				return edit
 			}
 		}
 
+		if haveExpectedText {
```
acke, line 267: 🟡 **Medium**

**[infrastructure/oss/code_actions.go:225]** Confidence-vs-correctness mismatch in the stale-edit guard. Comment claims it detects "the file has changed since the quickfix was created", but it only compares disk content, not the live editor buffer. When the LSP client applies the edit in-memory without flushing to disk (the actual IDE-2139 scenario), a second apply reads unchanged disk content, the guard passes, and a 5→6 char edit overwrites the trailing character of the prior edit (e.g. `11.0.1` → `11.0.11`). The property-redirect keeps XML well-formed, but the guard does not prevent the re-apply corruption it advertises.

**Fix:**
Dedupe via the edit itself — record `applied` in the closure, or compare `currentText` to the quickfix's replacement and short-circuit when the file is already at the upgrade target. Alternatively rename/document the guard as disk-only so callers do not rely on it for in-memory drift.

— AI review comment

infrastructure/oss/code_actions.go
```diff
@@ -249,6 +302,48 @@ func getQuickfixEdit(engine workflow.Engine, affectedFilePath types.FilePath, up
...
+func textAtRange(content []byte, r types.Range) (text string, ok bool) {
+	if content == nil {
+		return "", false
+	}
+	lines := strings.Split(string(content), "\n")
+	if r.Start.Line != r.End.Line || r.Start.Line < 0 || r.Start.Line >= len(lines) {
+		return "", false
+	}
+	line := lines[r.Start.Line]
+	if r.Start.Character < 0 || r.Start.Character > r.End.Character || r.End.Character > len(line) {
```
acke, line 426: 🟡 **Medium**

**[infrastructure/oss/code_actions.go:341]** `textAtRange` rejects multi-line ranges outright (`Start.Line != End.Line`). Since this silently disables the staleness guard for any caller passing a multi-line range, a future caller could regress safety without test failure.

**Fix:**
Either support multi-line ranges (slice flat content using line-start offsets) or surface the rejection at action-creation time so callers learn at build time instead of at apply time.

— AI review comment

infrastructure/oss/code_actions.go
```diff
@@ -155,7 +179,12 @@ func AddSnykLearnAction(
 	return action
 }
 
-func AddQuickFixAction(engine workflow.Engine, configResolver types.ConfigResolverInterface, affectedFilePath types.FilePath, issueRange types.Range, fileContent []byte, addFileNameToFixTitle bool, packageManager string, dependencyPath []string, upgradePath []any, folderConfig *types.FolderConfig) types.CodeAction {
+// AddQuickFixAction builds a deferred "upgrade" code action. fileContent, when
+// non-nil, is the document content used at apply time (otherwise the file is
+// read from disk on apply). originalContent is a snapshot of the document at
+// action-creation time; it is used to capture the exact text the edit expects
+// to replace so a stale edit can be refused at apply time.
+func AddQuickFixAction(engine workflow.Engine, configResolver types.ConfigResolverInterface, affectedFilePath types.FilePath, issueRange types.Range, fileContent []byte, originalContent []byte, addFileNameToFixTitle bool, packageManager string, dependencyPath []string, upgradePath []any, folderConfig *types.FolderConfig) types.CodeAction {
```
acke: 🟡 **Medium**

**[infrastructure/oss/code_actions.go:187]** Breaking change to exported `AddQuickFixAction` signature — added required `originalContent []byte` parameter. All in-repo callers are updated, but the package sits under a public Go module path (`github.com/snyk/snyk-ls/infrastructure/oss`); external importers will fail to compile.

**Fix:**
Either unexport `addQuickFixAction` (no external consumers found in this repo), or keep the old signature and introduce `AddQuickFixActionWithSnapshot` that wraps it. If the breakage is intentional, call it out in the PR description.

— AI review comment

infrastructure/oss/code_actions_test.go
```diff
@@ -0,0 +1,242 @@
...
+	propLine, propStart := locate(t, propertyManagedPom, "<lib.cyclonedx-core-java.version>", "9.0.5")
+	assert.Equal(t, propLine, edits[0].Range.Start.Line, "edit should be on the <properties> line")
+	assert.Equal(t, propStart, edits[0].Range.Start.Character)
+	assert.Equal(t, propStart+len("9.0.5"), edits[0].Range.End.Character)
+}
+
+// F2: a deferred quickfix edit must refuse to apply if the file content at the
+// cached range no longer matches what the action was created for. This prevents
+// re-applying a stale edit (which corrupts the file).
+func Test_AddQuickFixAction_RefusesStaleEdit(t *testing.T) {
```
acke: 🟡 **Medium**

**[infrastructure/oss/code_actions_test.go:92]** Stale-edit test only exercises the "text changed" branch. `textAtRange` has three other refusal branches (`Start.Line >= len(lines)`, `End.Character > len(line)`, `Start.Character > End.Character`) that fire when the file got shorter after an apply — none are tested.

**Fix:**
Add cases where the post-change file has fewer lines than the cached range, and where the cached `End.Character` exceeds the now-shorter line length. Assert `edit.Changes` is empty in both.

— AI review comment

infrastructure/oss/code_actions_test.go
```diff
@@ -0,0 +1,242 @@
...
+	require.Len(t, edits, 1)
+	assert.Equal(t, "11.0.1", edits[0].NewText)
+	assert.Equal(t, versionRange, edits[0].Range)
+}
+
+// This is the customer-reported corruption (IDE-2139): the same cached quickfix
+// applied twice without a rescan. With the fix the edit targets the
+// <properties> value, so the dependency's <version>...</version> block — and
+// crucially its closing tag — is never touched, and the file stays well-formed.
+func Test_MavenPropertyQuickfix_AppliedTwice_DoesNotCorruptPom(t *testing.T) {
```
acke, line 385: 🟡 **Medium**

**[infrastructure/oss/code_actions_test.go:152]** `Test_MavenPropertyQuickfix_AppliedTwice_DoesNotCorruptPom` does not actually exercise the stale-edit guard — the disk file is never updated between iterations, so `currentText == expectedText` on both passes. The test only proves "edit pointed at `<properties>` line ⇒ no `</version>` deletion"; it does not prove the guard kicks in on a real second apply, nor does it assert the property value itself is `11.0.1` after two applies (under the in-memory-only re-apply path it would silently become `11.0.11`).

**Fix:**
Between the two `GetDeferredEdit()` calls, write the post-edit content back to disk (`os.WriteFile(pomPath, []byte(result), 0600)`); assert iteration 2 returns `edit.Changes` empty (guard refuses stale). Also assert the property value after N applies equals `11.0.1`, so the regression actually fails if in-memory drift returns.

— AI review comment

infrastructure/oss/code_actions_test.go
acke, line 57: 🟡 **Medium**

**[infrastructure/oss/code_actions_test.go:55]** Only one property name (`lib.cyclonedx-core-java.version`) is verified. No coverage for property names that contain characters the regex `^\s*\$\{([^}]+)\}\s*$` might mishandle, or values with surrounding whitespace (`<v>  9.0.5  </v>`).

**Fix:**
Parameterise as table-driven subtests covering bare name, dotted name, leading/trailing whitespace inside `${ }`, and whitespace-padded property value.

— AI review comment

ast/maven/parser_test.go
```diff
@@ -56,6 +57,47 @@ func TestCreateDependencyTree(t *testing.T) {
 	assert.Equal(t, index+len(version), node.EndChar)
 }
 
+func TestParse_PopulatesProperties(t *testing.T) {
```
acke, line 63: 🟡 **Medium**

**[ast/maven/parser_test.go:60]** Only the cyclonedx property asserts `Line/StartChar/EndChar`; the second property (`maven.compiler.source`) only checks `Value`. Off-by-one regressions in `newValueNode` position math on the simple (no-comment) case would not be caught.

**Fix:**
Assert `Line/StartChar/EndChar` for `maven.compiler.source` too.

— AI review comment

acke, line 63: 🟡 **Medium**

**[ast/maven/parser_test.go:60]** No negative / edge cases for properties — empty `<properties/>`, whitespace-only value, no `<properties>` block, duplicate property names, nested-element property values. `addPropertyNodes`' explicit `trimmedValue == ""` and `valueStartOffset > valueEndOffset` skip branches are unreachable from current tests.

**Fix:**
Add subtests: (a) `<properties></properties>` → `tree.Properties` non-nil and empty; (b) pom with no `<properties>` block → `tree.Properties` non-nil and empty (not nil — `resolveMavenPropertyNode` does `tree.Properties[name]` lookup); (c) `<foo>   </foo>` → key absent.

— AI review comment

ast/maven/parser.go
andrewrobinsonhodges-snyk: Added a regular-file check (os.Stat + Mode().IsRegular()) before reading the parent POM, so non-regular paths (devices, pipes, dirs) are skipped with a warning. We deliberately did not add a workspace-root containment check: parent POMs legitimately live outside the project directory (e.g. ../pom.xml or a shared parent), and the relativePath originates from a POM already inside the user's trusted workspace, so strict containment would break valid multi-module layouts.

ast/maven/parser.go
andrewrobinsonhodges-snyk: Fixed: Parse now delegates to an internal recursive worker that threads a visited-set (absolute paths already parsed in this chain) plus a depth cap (maxParentDepth=32) as a backstop. handleParent refuses a parent path already in the visited-set (cycle) or beyond the depth limit, logging a warning in both cases. Parse's exported signature is unchanged. Added TestParse_CyclicParentReference_DoesNotRecurseInfinitely covering two mutually-referencing POMs.

infrastructure/oss/code_actions_test.go
andrewrobinsonhodges-snyk: Added Test_GetCodeActions_MavenUnresolvedProperty_FallsBackToDependencyVersion: a pom with <version>${missing.version}</version> and no <properties> entry. resolveMavenPropertyNode returns nil, so the quickfix falls back to editing the dependency <version> block. The test asserts the action is still produced, the edit targets the dependency <version> range (start/end of ${missing.version}), NewText is 11.0.1, and there's no panic.

andrewrobinsonhodges-snyk: Added Test_GetCodeActions_MavenPropertyInParentPom_RedirectsToParent: a child pom whose dependency uses ${foo.version} with a <parent> relativePath pointing at a parent pom that defines <foo.version>9.0.5</foo.version>. This exercises the tree.ParentTree walk in resolveMavenPropertyNode. Asserts the quickfix edit lands in edit.Changes[parentPath] (the parent's <properties> value range), the child pom is untouched, and NewText is 11.0.1.

ast/maven/parser.go
andrewrobinsonhodges-snyk: Added a size cap (maxParentPOMSize = 16MiB) checked against the fi from os.Stat we already obtain, before reading. Files larger than the cap are skipped with a warning. Note the IsRegular() check added above already rejects /dev/zero and other non-regular files.

ast/maven/parser.go
andrewrobinsonhodges-snyk: Fixed: addPropertyNodes now tracks nesting depth — increment on StartElement, decrement on EndElement, only ending the value loop when depth returns to 0 (the property's own closing tag). valueEnd is only advanced on depth-0 CharData. This keeps the decoder positioned past the whole subtree so a nested child (e.g. <c>) is no longer mis-registered as a top-level property. Added TestParse_PropertyWithNestedElement_DoesNotMisregisterSiblings, which asserts the nested child isn't registered and the sibling property after it is still parsed with correct position.

ast/maven/parser.go
andrewrobinsonhodges-snyk: Added the lower-bound guard: the slice condition now also checks valueStartOffset < 0. Combined with the existing valueStartOffset > valueEndOffset and valueEndOffset > len(content) checks, valueStartOffset <= len(content) is transitively guaranteed, so content[valueStartOffset:valueEndOffset] can no longer panic on unexpected offsets.

infrastructure/oss/code_actions.go
andrewrobinsonhodges-snyk: Fixed: resolveMavenPropertyNode now follows property-to-property indirection. It loops up to maxPropertyIndirectionDepth (16) with a seen-set for cycle detection, and only returns a node whose value is a concrete (non-${...}) string. If the chain hits a cycle, a missing link, or the depth limit, it returns nil and the quickfix falls back to the dependency <version> edit. Lookup is factored into lookupMavenProperty (current pom + parent hierarchy). Added Test_GetCodeActions_MavenPropertyIndirection_FollowsChainToConcreteValue (${a.version} -> ${b.version} -> 9.0.5) asserting the edit targets b.version, not the intermediate reference.

infrastructure/oss/code_actions.go
andrewrobinsonhodges-snyk: Fixed: AddQuickFixAction now drops the action (returns nil + warning) when a snapshot was provided but textAtRange can't read the fix range — rather than silently creating an unguarded edit. When no snapshot is provided at all there's nothing to guard, so legacy callers are unaffected. Added Test_AddQuickFixAction_DropsActionWhenSnapshotProvidedButRangeUnreadable using a multi-line range.

infrastructure/oss/code_actions.go
andrewrobinsonhodges-snyk: Added a single-shot latch in the deferred callback: once the action has produced its edit, subsequent invocations are refused. The disk-based guard only sees on-disk content, so it can't detect the in-memory re-apply case (disk unchanged, buffer already at 11.0.1) where a second apply would shift to 11.0.11 — the latch closes that gap and keeps the disk guard as a secondary defense. Updated Test_MavenPropertyQuickfix_AppliedTwice_DoesNotCorruptPom to persist the result to disk between applies, assert the second apply is refused (empty), and assert the property value upgrades exactly once (11.0.1, not 11.0.11).

infrastructure/oss/code_actions.go
andrewrobinsonhodges-snyk: Addressed via the action-creation-time rejection added for the related comment: when a snapshot is provided but textAtRange rejects the range (multi-line included), AddQuickFixAction now drops the action with a warning rather than silently disabling the guard at apply time. This is the 'surface the rejection at action-creation time' option you suggested. Full multi-line slicing isn't needed here since Maven version/property values are single-line. Test_AddQuickFixAction_DropsActionWhenSnapshotProvidedButRangeUnreadable covers a multi-line range, so a future caller relying on a multi-line range now fails loudly (nil action) instead of getting an unguarded edit.

infrastructure/oss/code_actions_test.go
andrewrobinsonhodges-snyk: Added Test_textAtRange_RefusesUnreadableRanges — a table-driven unit test directly on textAtRange covering all refusal branches: nil content, multi-line range, negative start line, start line past EOF (file got shorter), negative start char, start char > end char, and end char past line length (line got shorter), plus the valid single-line happy path. Testing textAtRange directly is more precise than driving the full action flow, and locks in that every unreadable range is refused so the stale-edit guard can't silently pass.

infrastructure/oss/code_actions_test.go
andrewrobinsonhodges-snyk: Done as part of the related stale-edit-guard fix. Test_MavenPropertyQuickfix_AppliedTwice_DoesNotCorruptPom now writes the post-edit content back to disk between applies (os.WriteFile), asserts the second apply returns edit.Changes empty (refused), and asserts the property value is upgraded exactly once — <lib.cyclonedx-core-java.version>11.0.1</...>, not 11.0.11 — so the regression fails loudly if in-memory drift returns.

infrastructure/oss/code_actions_test.go
andrewrobinsonhodges-snyk: Added Test_resolveMavenPropertyNode_NameAndWhitespaceVariants (table-driven) covering a bare name, a dotted name, a reference written with padding (${ foo.version }), and a whitespace-padded property value (<v>  9.0.5  </v>). To make the padded-reference case work I also trimmed the regex capture in resolveMavenPropertyNode (the ${...} capture group includes inner whitespace), so ${ foo.version } now resolves to the foo.version key. The padded value was already trimmed by the parser. Each case asserts resolution to the concrete 9.0.5 node. Tested resolveMavenPropertyNode directly rather than the full action flow for precision.

ast/maven/parser_test.go
andrewrobinsonhodges-snyk: Added Line/StartChar/EndChar assertions for maven.compiler.source in TestParse_PopulatesProperties, mirroring the cyclonedx checks, so off-by-one regressions in newValueNode position math on the simple (no-comment) case are now caught.

andrewrobinsonhodges-snyk: Added TestParse_PropertiesEdgeCases (table-driven): (a) empty <properties></properties> → Properties non-nil and empty; (b) no <properties> block → Properties non-nil and empty (the non-nil invariant resolveMavenPropertyNode's map lookup relies on); (c) <foo>   </foo> whitespace-only value → key absent (exercises the trimmedValue == "" skip branch). Nested-element property values are already covered by TestParse_PropertyWithNestedElement_DoesNotMisregisterSiblings.

infrastructure/oss/code_actions.go
```diff
@@ -64,13 +66,35 @@ func GetCodeActions(engine workflow.Engine, configResolver types.ConfigResolverI
 			)
 		}
 	} else {
```
rrama, line 88: AI found improvement (unchecked by me):

> Parent-POM dependency versions aren't checked for ${property} references.
> 
> When the version lives in a parent POM (the issueDepNode.Value == "" branch, code_actions.go:51-67), fixNode is the <dependency> node from the parent POM and fixNode.Value can be a ${prop} reference. But resolveMavenPropertyNode is only called in the else-branch (code_actions.go:82-88), so here the quickfix hardcodes the literal version into the parent's <version>, orphaning the property — the same corruption this PR fixes for the child case.
> 
> Fix: mirror the else-branch before the addQuickFixAction call:
> 
> ```go
> fixFilePath := types.FilePath(fixNode.Tree.Document)
> fixRange := getRangeFromNode(fixNode)
> originalContent := []byte(fixNode.Tree.Root.Value)
> if ossIssueData.PackageManager == "maven" {
>     if propNode := resolveMavenPropertyNode(fixNode); propNode != nil {
>         fixFilePath = types.FilePath(propNode.Tree.Document)
>         fixRange = getRangeFromNode(propNode)
>         originalContent = []byte(propNode.Tree.Root.Value)
>     }
> }
> ```
> then pass fixFilePath/fixRange/originalContent into addQuickFixAction. Verified resolveMavenPropertyNode works on fixNode: getDependencyNode sets LinkedParentDependencyNode from the parent POM's content (infrastructure/oss/range_finder.go:80-83), so fixNode.Tree carries the parent's .Properties and .ParentTree chain. Worth a test too — no parent-POM-property case exists today.

ast/maven/parser.go
```diff
@@ -50,15 +63,29 @@ type Dependency struct {
...
+	return p.parse(content, path, map[string]bool{}, 0)
+}
+
+// parse is the recursive worker behind Parse. visited holds the absolute paths of
+// POMs already parsed in this parent chain (cycle detection) and depth bounds the
+// chain length (see maxParentDepth).
+func (p *Parser) parse(content string, path types.FilePath, visited map[string]bool, depth int) *ast.Tree {
 	content = strings.ReplaceAll(content, "\r", "")
 	tree := p.initTree(path, content)
+	if absPath, err := filepath.Abs(string(path)); err == nil {
```
rrama: AI comment (unchecked by me):

> **Cycle detection uses unresolved paths while isWithinRoot resolves symlinks**
> 
> The `visited` set in `ast/maven/parser.go:86-88` keys on `filepath.Abs(path)` — which does NOT resolve symlinks. Meanwhile `isWithinRoot` at `parser.go:209-214` resolves both the root and parent paths via `filepath.EvalSymlinks`. This means two different symlinks pointing at the same real file would be treated as distinct POMs by the cycle detector but as the same path by the containment check.
> 
> In practice the `maxParentDepth` backstop (line 167) bounds the chain regardless, so this cannot cause infinite recursion. But if symlinks are common in a workspace, the depth budget could be consumed by redundant re-parses of the same real file. Aligning the visited-set with `EvalSymlinks` would be a minor improvement.

infrastructure/oss/code_actions.go
rrama: Also, AI is telling me about a disk guard that is missing in this branch, but it seems like something we don't even need in the other branch, so make with this information what you will.

rrama (approved):

infrastructure/oss/code_actions.go
andrewrobinsonhodges-snyk: Good catch — fixed. `GetCodeActions` now mirrors the property-resolution logic into the parent-POM branch (`issueDepNode.Value == ""`): when the dependency version is inherited from a parent POM and that parent's `<version>` is itself a `${property}` reference, `resolveMavenPropertyNode(fixNode)` redirects the edit (and its guard snapshot) to the matching `<properties>` entry instead of hardcoding the version into `<version>` and orphaning the property. Added `Test_GetCodeActions_MavenParentPomPropertyVersion_RedirectsToParentProperty` covering a versionless child dep whose version lives in the parent POM as `${foo.version}`, asserting the edit targets the parent's `<foo.version>` property (not `<version>`) and the child POM is untouched. Re the disk guard: this branch already passes the resolved tree's content as the apply-time snapshot, so the stale-edit guard still applies without extra plumbing.

ast/maven/parser.go
andrewrobinsonhodges-snyk: Fixed. Added a `canonicalPath` helper (`filepath.Abs` → `filepath.EvalSymlinks`, falling back to the abs path / input when symlinks can't be resolved) and keyed the `visited` set through it on **both** the populate side (`parse`) and the lookup side (`handleParent`), so the cycle detector now agrees with `isWithinRoot`'s symlink semantics — two symlinks to the same real POM are deduped rather than re-parsed. Added `TestParse_CyclicParentViaSymlink_DetectedByCanonicalPath`, which reaches a parent via a symlink back to the same real file: it's now caught as a cycle on the first hop (previously it re-parsed up to `maxParentDepth`). Left `isWithinRoot` unchanged. Note this resolves symlink aliasing specifically — hard links to the same inode remain distinct keys, but that's still bounded by `maxParentDepth`.

PR #1343, opened 2026-06-17
fix: hide autofix messages when autofix is not available [IDE-1880]

andrewrobinsonhodges-snyk:
### Description

Only show the Autofix summaries if they are applicable:

Code: If Autofix is enabled
OSS: If packages can be automatically updgraded
IAC: Never
Secrets: Never

This prevents the "There are no issues automatically fixable" text from showing in scenarios where there would **never** be automatically fixable issues.

### Checklist


rrama (approved):
Other than your branches being a bit out of sync and some failing tests that may need some tweaks, LGTM.

PR #1344, opened 2026-06-17
feat: per-folder reset overrides + fix webview no-op [IDE-1945]

nick-y-snyk:
### Description

Implements per-folder **Reset overrides** in the settings dialog and fixes two defects that stopped it from working in VSCode.

**Feature** ([IDE-1945](https://snyksec.atlassian.net/browse/IDE-1945)): a per-folder "Reset overrides" button that clears all of that folder's user overrides, reverting each setting to its org / LDX-sync / default value. Reset rides the existing settings-save path — no new `executeCommand`. A reset emits flat `null` for each override field; the IDE maps it to `{value:null, changed:true}` and the LS `Unset`s the `user:folder:` override.

**Fixes** found during verification:

1. **No-op in sandboxed VSCode webview.** The dialog renders in a sandboxed iframe without `allow-modals`, so `window.confirm()` is silently ignored and returns `false`. Both reset handlers gated their work behind `confirm()`, so clicking the button bailed and nothing happened. Removed the `confirm()` prompts (reset is reversible and the dialog shows a disclaimer banner).

2. **Incomplete reset.** `additional_parameters`, `additional_environment` and `scan_command_config` are folder overrides editable in the dialog but were never cleared: the JS omitted them from `FOLDER_RESET_FIELDS`, and on the LS side they are processed by `applyBasicFolderFields` (marked `handled`), so the generic null-reset loop skipped them while their own handlers ignored a `nil` value — making the reset unreachable. Added the three keys and taught `applyStringField` / `applyStringSliceField` / `applyScanCommandConfig` to `Unset` on `{Changed:true, Value:nil}`.

### Checklist

[IDE-1945]: https://snyksec.atlassian.net/browse/IDE-1945?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


infrastructure/configuration/template/js/ui/reset-handler.js
```diff
@@ -87,21 +83,36 @@
...
+		// Keep dirty state in sync (a reset changes no DOM input, so no change/blur event fires).
 		if (window.dirtyTracker) {
 			window.dirtyTracker.runChangeListeners();
 			window.dirtyTracker.checkDirty();
 		}
+
+		// Persist the reset immediately on every IDE. We call getAndSaveIdeConfig() directly rather
+		// than formState.triggerChangeHandlers(), which only saves on auto-save IDEs — a reset must
+		// be a commit point everywhere (incl. OK/Cancel IDEs), since no input event would carry it.
+		if (window.ConfigApp.autoSave && window.ConfigApp.autoSave.getAndSaveIdeConfig) {
```
rrama, line 109: AI bug (unverified by me):

> **Validation errors block folder reset save silently**
> 
> The immediate save triggered by clicking 'Reset overrides' (`reset-handler.js:109-111`) calls `getAndSaveIdeConfig()` which checks form validation at `auto-save.js:31-37`. If ANY form field has a validation error (e.g. invalid API endpoint or risk score), the entire save is blocked — including the folder reset. The folderResets map IS preserved (since `applyFolderResets` wasn't reached), so the reset would be saved on the next successful save. But from the user's perspective, clicking 'Reset overrides' appears to do nothing when there's an unrelated validation error elsewhere in the form. Not a bug per se, but a UX consideration.

I think this would be a bug if true, but not one large enough to block the PR, would just need a follow-on Jira bug created.

docs/configuration-dialog.md
```diff
@@ -151,6 +151,32 @@ On save (or auto-save), `features/auto-save.js` calls `form-handler.collectData(
 
 See [Saving Configuration Flow](#saving-configuration-flow) for the sequence diagram.
 
+### 4a. Resetting folder overrides
+
+The per-folder **"Reset overrides"** button (one per folder tab, top-right in the disclaimer banner) clears all of that folder's user overrides so the effective values fall back to org / LDX-sync / default.
+
+**JS side (snyk-ls, already implemented):** clicking the button marks the folder *by its `folderPath`* (read from the hidden `folder_<index>_folderPath` input — not the index, which is compacted away during diffing). On save, `form-handler.applyFolderResets()` emits **flat `null`** for each of these 14 folder fields on that folder's entry in `folderConfigs`:
```
rrama: AI found discrepancy:

> **Documentation lists 14 reset fields but production code emits 17**
> 
> The documentation says `applyFolderResets()` emits flat null for "14 folder fields" and lists them explicitly, but the production `FOLDER_RESET_FIELDS` array in `infrastructure/configuration/template/js/ui/form-handler.js:216-234` has 17 entries. The 3 missing fields are `additional_parameters`, `additional_environment`, and `scan_command_config`. This understates what the JS actually emits and could mislead IDE plugin developers implementing the reset contract. The repository rules require documentation to be kept up-to-date.

docs/configuration.md
```diff
@@ -456,6 +456,23 @@ The `Changed` field on `ConfigSetting` controls whether the LS processes a setti
...
+A `ConfigSetting` value of `nil` combined with `Changed: true` is a **reset**: it tells the LS to `Unset` the `user:folder:<path>:<name>` override so the effective value falls back through the precedence chain (org / LDX-sync / GAF default). This is how the HTML settings dialog's per-folder "Reset overrides" button clears a folder's user edits.
+
+| `ConfigSetting` on the wire | LS behaviour |
+|---|---|
+| absent from the map (`cs == nil`) | no-op |
+| `{Changed: false}` (or omitted) | no-op (skipped, as above) |
+| `{Changed: true, Value: <x>}` | set `user:folder:` override to `<x>` |
+| `{Changed: true, Value: nil}` | **Unset** `user:folder:` override → fall back |
+
+The null-reset applies to org-scope folder fields (those resolved through `applyGenericFolderOverrides`), which have a lower-precedence layer to fall back to. **Basic folder-native fields** (`base_branch`, `scan_command_config`, etc.) have no fallback layer, so a null value on them is treated as a no-op rather than a reset.
```
rrama: AI found discrepancy:

> **Documentation incorrectly claims scan_command_config null-reset is a no-op**
> 
> The newly-added documentation states that basic folder-native fields like `scan_command_config` treat a null value as a no-op. But the code added in this same PR at `internal/types/folder_config.go:503-504` explicitly handles null-reset for `scan_command_config` (via `isFolderReset` + `unsetFolderOverride`), and the Go E2E test at `application/server/configuration_test.go:779,791` confirms it works. The same applies to `additional_parameters` (`folder_config.go:488-489`) and `additional_environment` (`folder_config.go:473-474`). An IDE plugin developer reading this documentation would incorrectly believe these fields don't support null-reset. The repository rules require documentation to be kept up-to-date (`always keep the documentation up-to-date in (./docs)`).

js-tests/folder-reset.test.mjs
```diff
@@ -0,0 +1,205 @@
...
+	"severity_filter_low",
+	"snyk_oss_enabled",
+	"snyk_code_enabled",
+	"snyk_iac_enabled",
+	"snyk_secrets_enabled",
+	"issue_view_open_issues",
+	"issue_view_ignored_issues",
+	"risk_score_threshold",
+	"preferred_org",
+];
```
rrama, line 27: AI found test gap:

> **Test RESET_FIELDS list missing 3 fields that FOLDER_RESET_FIELDS emits**
> 
> The test's `RESET_FIELDS` constant has 14 entries but the production `FOLDER_RESET_FIELDS` in `infrastructure/configuration/template/js/ui/form-handler.js:216-234` has 17 entries. The missing fields are `additional_parameters`, `additional_environment`, and `scan_command_config`. The comment on line 8 explicitly claims it "mirrors FOLDER_RESET_FIELDS in form-handler.js" but it does not. All `assertAllNull` calls in the test only check 14 fields, so if any of the 3 missing fields were accidentally removed from the production list, the tests would not catch the regression.

infrastructure/configuration/template/js/ui/reset-handler.js
```diff
@@ -87,21 +83,36 @@
 			return;
 		}
 
-		if (!confirm("Reset all overrides for this folder to defaults? Your custom overrides will be removed.")) {
```
rrama, line 90: Why were these removed? Can we add them back?

Should also warn for non-auto save that it will save all other unsaved changes (we could display that conditionally if there are unsaved changes).

infrastructure/configuration/template/js/ui/reset-handler.js
```diff
@@ -87,21 +83,36 @@
...
 		if (window.dirtyTracker) {
 			window.dirtyTracker.runChangeListeners();
 			window.dirtyTracker.checkDirty();
 		}
+
+		// Persist the reset immediately on every IDE. We call getAndSaveIdeConfig() directly rather
+		// than formState.triggerChangeHandlers(), which only saves on auto-save IDEs — a reset must
+		// be a commit point everywhere (incl. OK/Cancel IDEs), since no input event would carry it.
+		if (window.ConfigApp.autoSave && window.ConfigApp.autoSave.getAndSaveIdeConfig) {
+			window.ConfigApp.autoSave.getAndSaveIdeConfig();
```
rrama, line 110: Nitpick: Calling a function in `autoSave` for non-**auto** saving is a bit of a code smell, may want to move the function out to somewhere else that makes more sense.

internal/types/config_resolver_test.go
```diff
@@ -1175,6 +1175,60 @@ func TestFolderConfig_ApplyLspUpdate(t *testing.T) {
 		assert.True(t, types.HasUserOverride(conf, fc.FolderPath, types.SettingScanAutomatic),
 			"ApplyLspUpdate should apply setting; lock enforcement is the caller's responsibility (validateLockedFields)")
 	})
+
+	t.Run("clears preferred_org override via explicit null and reverts to global org", func(t *testing.T) {
```
rrama: Reset goes back to auto-org not global. The test is correct, the name is not.

internal/types/folder_config.go
```diff
@@ -571,6 +598,24 @@ func (fc *FolderConfig) applyPreferredOrg(update *LspFolderConfig, handled map[s
...
+		changed := false
+		if HasUserOverride(conf, fc.FolderPath, SettingPreferredOrg) {
+			conf.Unset(configresolver.UserFolderKey(fp, SettingPreferredOrg))
+			changed = true
+		}
+		if HasUserOverride(conf, fc.FolderPath, SettingOrgSetByUser) {
+			conf.Unset(configresolver.UserFolderKey(fp, SettingOrgSetByUser))
+			changed = true
+		}
+		return changed
```
rrama: I think this does the same but simpler:
```go
if isFolderReset(update.Settings, SettingPreferredOrg) {
    return fc.unsetFolderOverride(conf, fp, SettingPreferredOrg) || 
        fc.unsetFolderOverride(conf, fp, SettingOrgSetByUser)
}
```

js-tests/folder-reset.test.mjs
```diff
@@ -0,0 +1,205 @@
...
+	const data = {};
+	fh.markFolderForReset(PATH_A);
+	fh.applyFolderResets(data);
+
+	assert.ok(Array.isArray(data.folderConfigs), "folderConfigs initialized");
+	assert.equal(data.folderConfigs.length, 1, "reset-only entry pushed");
+	assert.equal(data.folderConfigs[0].folderPath, PATH_A);
+});
+
+test("applyFolderResets nulls only the marked folder; others untouched (bug #1 — compaction/index drift)", async () => {
```
rrama: AI comment left in name.

js-tests/folder-reset.test.mjs
```diff
@@ -0,0 +1,205 @@
...
+	fh.markFolderForReset(PATH_A);
+	fh.applyFolderResets({ folderConfigs: [{ folderPath: PATH_A }] });
+
+	// Prototype-agnostic: the reset map is created in the JSDOM realm, so compare by key count
+	// rather than deepStrictEqual against this realm's {} (different Object.prototype).
+	assert.equal(Object.keys(win.ConfigApp.folderResets).length, 0, "folderResets cleared");
+	assert.equal(fh.isFolderMarkedForReset(PATH_A), false, "no longer marked");
+});
+
+test("DOM-driven: clicking .reset-overrides-btn produces 14 nulls for that folderPath in the save payload", async () => {
```
rrama: Again, the test talks about 14 nulls, but there should be more according to AI.

js-tests/folder-reset.test.mjs
```diff
@@ -0,0 +1,205 @@
...
+	btn.click();
+
+	assert.ok(calls.length > 0, "reset-only click must still save");
+	const saved = JSON.parse(calls[calls.length - 1]);
+	const entry = (saved.folderConfigs || []).find((f) => f.folderPath === PATH_A);
+	assert.ok(entry, "reset-only folder still emitted in payload");
+	assertAllNull(entry, RESET_FIELDS);
+});
+
+test("clicking reset triggers a save even with no other edits", async () => {
```
rrama: This test seems to be a dupe of the one above, so can we remove it?

js-tests/folder-reset.test.mjs
rrama, line 1: One testing gap would be the dirty tracker, we should assert it goes back to clean after resets.

rrama (approved):
Just a few bugs and my opinion on the confirmation dialog (Tars did want it kept), but rest looks good.

Also, please add pics to the PR description.

infrastructure/configuration/template/js/ui/reset-handler.js
rrama: I will do this in a follow up PR: https://github.com/snyk/snyk-ls/pull/1350 and https://github.com/snyk/vscode-extension/pull/771

infrastructure/configuration/template/js/ui/reset-handler.js
rrama: AI reply:

Confirmed — validated against the code. getAndSaveIdeConfig() hits the validation check at auto-save.js:32 and returns early before applyFolderResets() at line 50 ever runs. The folderResets map is preserved in memory so the reset lands on the next successful save, but the user sees nothing.

Created [IDE-2162](https://snyksec.atlassian.net/browse/IDE-2162) to track.

docs/configuration-dialog.md
rrama: Fixed.

docs/configuration.md
rrama: Fixed.

infrastructure/configuration/template/js/ui/reset-handler.js
rrama: Not going to do in this PR given the time crunch.

internal/types/config_resolver_test.go
rrama: Fixed.

internal/types/folder_config.go
rrama: It did not, but I have a correct fix in.

js-tests/folder-reset.test.mjs
rrama: Removed.

js-tests/folder-reset.test.mjs
rrama: Fixed.

js-tests/folder-reset.test.mjs
rrama: Removed.

js-tests/folder-reset.test.mjs
rrama: Added.

js-tests/folder-reset.test.mjs
rrama: Done.

PR #1346, opened 2026-06-18
feat(settings): propagate Project Default changes to folder panes [IDE-2111]

acke:
## Summary

- After saving a Project Default value, walk folder pane inputs (matched via `data-setting=<key>`) and overwrite those that were inheriting from PD.
- Skip folder inputs whose wrapper carries `source-org` / `source-org-locked`, or whose current value differs from the pre-save PD value (user override).
- No `location.reload()` — works in webviews without IDE plugin changes.

Closes [IDE-2111](https://snyksec.atlassian.net/browse/IDE-2111).
Parent epic: [IDE-2029](https://snyksec.atlassian.net/browse/IDE-2029).

## Why prior PR #1185 was reverted

It looked for `folder_N_override_<setting>` — but the template names folder inputs `folder_N_<setting>` with `data-setting="<setting>"`. Severity-filter propagation silently no-op'd. This PR keys off `data-setting`, making that bug structurally impossible.

## Field-name audit

| Scope | Element naming | Container | Wrapper source classes |
|---|---|---|---|
| Project Defaults | `name="<setting>"` | `#fallbacks-pane` | `source-org-locked` / `source-org` / "" |
| Folder pane | `name="folder_N_<setting>"` + `data-setting="<setting>"` | `#folder-pane-N`, `.folder-pane` | `source-org-locked` / `source-org` / "" |

`getSourceClass` (in `config_html.go`) emits only three values; **no `source-override` class exists server-side**. User-override is detected client-side by comparing folder DOM value to `dirtyTracker.originalData[<setting>]` (the pre-save PD baseline).

## Algorithm

For each key in `changedData` (skipping `folderConfigs` / `trusted_folders`):
1. Find PD input in `#fallbacks-pane` by `name=<key>`. Absent → skip (not PD-scope).
2. Read post-save PD value from DOM.
3. For each folder input with `data-setting=<key>`:
   - Skip if wrapper has `source-org` / `source-org-locked`.
   - Skip if folder's current value ≠ pre-save PD value (user override).
   - Otherwise write the new PD value + dispatch `change` event.

Runs inside `auto-save.js` save-success branch, **before** `dirtyTracker.reset()` so propagated DOM values land in the fresh baseline (form stays clean — propagation is invisible to the dirty tracker).

## Files

| File | Change |
|---|---|
| `infrastructure/configuration/template/js/features/project-default-propagator.js` | New, ~140 LOC |
| `infrastructure/configuration/template/js/features/auto-save.js` | 6-line wiring inside `saveSuccess` |
| `js-tests/propagation.test.mjs` | 11 unit tests |
| `js-tests/auto-save-propagation.test.mjs` | 2 integration tests |
| `docs/requirements/requirements.md` | 7 IDE-2111 requirements |

## Test plan

- [x] Unit: walks folder panes, only updates non-overridden fields
- [x] Unit: setting discovery via `data-setting`, not name-parsing (PR #1185 regression guard)
- [x] Unit: checkbox / select / text / number branches each handled
- [x] Unit: `source-org` / `source-org-locked` wrappers skipped
- [x] Unit: keys with no PD-scope input ignored (no error, no propagation)
- [x] Integration: full auto-save → propagation, dirty state remains clean
- [x] Integration: severity_filter_high rename regression guard
- [x] Full `js-tests` suite: 127/127 pass
- [ ] Manual smoke: VS Code
- [ ] Manual smoke: IntelliJ
- [ ] Manual smoke: Eclipse

[IDE-2111]: https://snyksec.atlassian.net/browse/IDE-2111?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[IDE-2029]: https://snyksec.atlassian.net/browse/IDE-2029?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/infrastructure/configuration/template/js/features/auto-save.js
+++ b/infrastructure/configuration/template/js/features/auto-save.js
@@ -66,6 +66,12 @@
 				if (window.ConfigApp.authFieldMonitor && window.ConfigApp.authFieldMonitor.resetSavedState) {
 					window.ConfigApp.authFieldMonitor.resetSavedState();
 				}
+				// Propagate changed Project Default values into folder panes that were
+				// inheriting from PD (i.e. not user-overridden). Must run BEFORE reset()
+				// so the propagated folder values become the new baseline (not dirty).
+				if (window.ConfigApp.projectDefaultPropagator && window.dirtyTracker) {
+					window.ConfigApp.projectDefaultPropagator.propagate(data, window.dirtyTracker.originalData);
+				}
 				// Reset dirty state after successful save (baseline must be the full form state)
 				if (window.dirtyTracker) {
 					window.dirtyTracker.reset();
--- a/infrastructure/configuration/template/js/features/project-default-propagator.js
+++ b/infrastructure/configuration/template/js/features/project-default-propagator.js
@@ -1,0 +1,143 @@
+// ABOUTME: Project Default Propagation — after a successful save, walks all
+// ABOUTME: folder panes and overwrites inputs that were inheriting from
+// ABOUTME: Project Defaults (PD) with the new PD value.
+//
+// Exposed as: window.ConfigApp.projectDefaultPropagator.propagate(changedData, originalPdValues)
+//
+// Algorithm:
+//   For each key in changedData (skipping folderConfigs / trusted_folders):
+//     1. Find the PD input in #fallbacks-pane by name=<key>.
+//        If absent, skip — not a PD-scope field.
+//     2. Read the post-save PD value from the DOM (authoritative after save).
+//     3. For each folder input matched via data-setting=<key>:
+//        a. Skip if wrapper carries source-org or source-org-locked.
+//        b. Skip if folder's current value != old PD value (user override).
+//        c. Otherwise write the new PD value and dispatch a change event.
+
+(function () {
+	'use strict';
+
+	window.ConfigApp = window.ConfigApp || {};
+
+	// ---------------------------------------------------------------------------
+	// Module-private helpers
+	// ---------------------------------------------------------------------------
+
+	/**
+	 * Read the effective value from a form element.
+	 * Returns a boolean for checkboxes, a string for everything else.
+	 * @param {HTMLInputElement|HTMLSelectElement} el
+	 * @returns {boolean|string}
+	 */
+	function readValue(el) {
+		if (el.type === 'checkbox') {
+			return el.checked;
+		}
+		return el.value;
+	}
+
+	/**
+	 * Write a value to a form element.
+	 * @param {HTMLInputElement|HTMLSelectElement} el
+	 * @param {boolean|string} value
+	 */
+	function writeValue(el, value) {
+		if (el.type === 'checkbox') {
+			el.checked = !!value;
+		} else {
+			el.value = String(value);
+		}
+	}
+
+	/**
+	 * Compare two values for equality, normalising types consistently.
+	 * Checkboxes produce booleans; text/number/select produce strings.
+	 * We compare with String() on both sides when neither is a pure boolean.
+	 * @param {boolean|string} a
+	 * @param {boolean|string} b
+	 * @returns {boolean}
+	 */
+	function valuesEqual(a, b) {
+		if (typeof a === 'boolean' || typeof b === 'boolean') {
+			return !!a === !!b;
+		}
+		return String(a) === String(b);
+	}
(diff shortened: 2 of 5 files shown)


infrastructure/configuration/template/js/features/auto-save.js
```diff
@@ -66,6 +66,12 @@
 				if (window.ConfigApp.authFieldMonitor && window.ConfigApp.authFieldMonitor.resetSavedState) {
 					window.ConfigApp.authFieldMonitor.resetSavedState();
 				}
+				// Propagate changed Project Default values into folder panes that were
+				// inheriting from PD (i.e. not user-overridden). Must run BEFORE reset()
+				// so the propagated folder values become the new baseline (not dirty).
+				if (window.ConfigApp.projectDefaultPropagator && window.dirtyTracker) {
```
basti-snyk, line 72: **Should Fix (blocking the feature) — the propagator is never wired into the rendered page, so this guard is always falsy in production.** Every other feature script is embedded via a `//go:embed template/js/features/<name>.js` directive + a template field + a `{{...}}` slot in `template/config.html` (e.g. `auto-save.js` → `{{.AutoSave}}`). `project-default-propagator.js` has none of these: no `//go:embed`, no template-data field, no `{{.ProjectDefaultPropagator}}` script slot. Regenerating `config_output_*.html` confirms the propagator script does not appear in the output. So `window.ConfigApp.projectDefaultPropagator` is undefined for every IDE and propagation silently no-ops — the entire IDE-2111 behaviour is dead in production. The `.mjs` tests inject the script manually, which is why they pass while the feature never runs.

Wiring an embedded asset requires editing `config_html.go` (Go) and `config.html` — the ticket's "JS only, no Go" constraint makes the feature unshippable as written; worth raising. A Go assertion in `config_html_test.go` that the rendered HTML contains `projectDefaultPropagator` would guard this.

— AI review

infrastructure/configuration/template/js/features/project-default-propagator.js
```diff
@@ -0,0 +1,143 @@
...
+	/**
+	 * Compare two values for equality, normalising types consistently.
+	 * Checkboxes produce booleans; text/number/select produce strings.
+	 * We compare with String() on both sides when neither is a pure boolean.
+	 * @param {boolean|string} a
+	 * @param {boolean|string} b
+	 * @returns {boolean}
+	 */
+	function valuesEqual(a, b) {
+		if (typeof a === 'boolean' || typeof b === 'boolean') {
```
basti-snyk, line 61: **Should Fix — `data-bool` selects (`scan_automatic`, `scan_net_new`) are propagated incorrectly.** This boolean branch does `!!a === !!b`, but for a `data-bool="1"` select, `form-handler.setFieldValue` stores the PD baseline as a real boolean (`el.value === "true"`), while `readValue` returns the raw string (`"true"`/`"false"`). Since `!!"false"` is `true` (non-empty string):

- folder inheriting `"false"` vs old PD `true` → `!!"false" === !!true` → `true === true` → treated as inheriting and **overwritten** (clobbers a user override the user explicitly set to off); and
- folder inheriting `"false"` vs old PD `false` → `!!"false" === !!false` → `true === false` → treated as an override and **skipped** (a real PD change never reaches the folder).

So toggling "Scan automatically" / "Scan net-new" at the PD level is unreliable for inheriting folders. Root cause: this file re-implements value coercion (`readValue`/`writeValue`/`valuesEqual`) that diverges from the canonical `form-handler.setFieldValue` / `FormUtils.normalizeValue` / `dirtyTracker.deepEquals`. Comparing through the existing normalizer (and reading folder values with the same type rules) fixes this and the empty-number case (`valuesEqual("", null)` → `"" === "null"` → false → empty `risk_score_threshold` inheritance wrongly skipped) at the root.

The unit tests miss this because they feed `originalPdValues` as strings (e.g. `scan_automatic: "true"`) where production produces booleans — derive test baselines from `collectData()`/`setFieldValue` semantics and add `data-bool` `false` + empty-number cases.

— AI review

rrama:
HTML refresh code covers this nicely, so closing.

PR #847, opened 2025-04-25
wip: tests mock refactor

ShawkyZ:
### Description

_Provide description of this PR and changes, if linked Jira ticket doesn't cover it in full._

### Checklist

- [ ] Tests added and all succeed
- [ ] Linted
- [ ] README.md updated, if user-facing
- [ ] License file updated, if new 3rd-party dependency is introduced

PR #1348, opened 2026-06-19
feat: implement global (Project Defaults) reset overrides [IDE-2149]

nick-y-snyk:
## What

Implements the language-server core of **IDE-2149** — global ("Project Defaults") reset overrides in the HTML settings page (sibling of the per-folder reset, IDE-1945). This is the shared wire contract the IDE PRs depend on.

On `{changed:true, value:null}` for an org-scope global setting, the LS now `Unset`s the `user:global` override so the effective value falls back through the resolver chain (LDX-Sync → org → flagset default). `organization` is reset via `config.ResetOrganization` (it is not stored at `UserGlobalKey`).

## Changes
- `config_writers.go` — `UnsetGlobalUser`, `HasGlobalUserOverride`, `GlobalResettableSettings` (14 org-scope keys; includes `organization`, excludes folder-only `preferred_org`).
- `config_readers.go` — `userGlobalValue` now treats GAF's `keyDeleted` sentinel as absent (a real bug: `conf.Unset` writes a `struct{}{}` tombstone, not nil, so a reset key would otherwise read as a present override).
- `config.go` — `ResetOrganization` (Unset `ORGANIZATION` + `last_set_organization`).
- `configuration.go` — `applyGlobalResets` pre-pass in `processConfigSettings` (runs before the typed appliers, deletes handled keys to avoid double-apply/analytics; filter resets refresh diagnostics; org reset folds into the single summary-panel path); extracted `refreshFoldersForGlobalOrgChange` shared with `applyOrganization`.
- HTML/JS — `reset-global-overrides-btn` on the Project Defaults tab, `GLOBAL_RESET_FIELDS` + `markGlobalForReset` + `applyGlobalResets` emitting top-level nulls, wired through `auto-save.js` and `reset-handler.js`. Regenerated config-dialog fixtures.
- Docs — `configuration.md`, `configuration-dialog.md`.

## Tests
- `config_writers_test.go` (4), `Test_ProcessConfigSettings_GlobalReset` (4 subtests), `js-tests/global-reset.test.mjs` (4). Targeted Go + `make test-js` + `make lint` green locally.
- ⚠️ Pushed with `--no-verify` at the author's request; the full `make test` gate runs in CI.

## Merge order
Merge **first** — the IDE PRs (vscode #768, intellij #847, eclipse #429) consume this contract. Until merged, an IDE null-reset is a no-op on the LS side, so the IDE PRs are independently safe.

___

### **PR Type**
Enhancement, Bug fix

___

### **Description**
- Implement global "Project Defaults" reset feature.

- Fix infinite settings refresh loop (IDE-2149).

- Add UI and backend logic for global reset.

- Refine LSP configuration update notification logic.

___

### Diagram Walkthrough

```mermaid
flowchart LR
  UI(Settings UI) -- Reset overrides button click --> JS(Form Handler JS)
  JS -- markGlobalForReset --> JS
  JS -- applyGlobalResets --> Payload(Save Payload {key: null})
  Payload -- Save Config --> Backend(Backend Services)
  Backend -- processConfigSettings --> applyGlobalResets{Apply Global Resets}
  applyGlobalResets -- UnsetGlobalUser/ResetOrganization --> Config(Configuration Store)

  Backend -- UpdateSettings --> beforeLspConfigSnapshot(Snapshot effective config)
  Backend -- UpdateSettings --> afterLspConfigSnapshot(Build new effective config)
  afterLspConfigSnapshot -- DeepEqual check --> configChanged{Effective config changed?}
  configChanged -- Yes --> Notify(Send $/snyk.configuration)
  configChanged -- No --> SkipNotify(Skip notification)

```

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><details><summary>10 files</summary><table>
<tr>
  <td><strong>config.go</strong><dd><code>Add ResetOrganization function for global org reset</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1348/changes#diff-d9a33e7e08d7d8fb8292de9d94d5bab14156125922662d7448de0951894ec11d">+16/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>configuration.go</strong><dd><code>Implement global reset logic and fix infinite refresh loop</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1348/changes#diff-eb3fe6fd1956591838595b08979653ec23c0b0c16f5340e9e89a05c30e921607">+187/-31</a></td>

</tr>

<tr>
  <td><strong>config_writers.go</strong><dd><code>Add GlobalResettableSettings, UnsetGlobalUser, HasGlobalUserOverride</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1348/changes#diff-83bf9ef0975d24f841c7e76f6b6f0f758b2f54d8fd63695ef1e9eb594068b7c6">+50/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>auto-save.js</strong><dd><code>Integrate global reset into save process and state clearing</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1348/changes#diff-cfc68dff85df34081fbe16a30f5f7658da5f71a0186710e89472a8ab2cbdf120">+10/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>form-handler.js</strong><dd><code>Define global reset fields and logic</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1348/changes#diff-631e55de1a3395dd284c5a8bb15cc71fa9de087f03b3442858eae129883538ee">+48/-3</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>reset-handler.js</strong><dd><code>Add handler for global reset button</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1348/changes#diff-4b26124d7fbe25fbe7b9c22f55d72d52dc3457ef6dc56b70e967cec69bed7987">+40/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>config.html</strong><dd><code>Add global reset button to Project Defaults tab</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1348/changes#diff-700a5eb160e325cc14adc0153f9fb936b41e4667abaf10337d6e2696d7644762">+1/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>config_output_multi_project.html</strong><dd><code>Add global reset button and styles for multi-project dialog</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1348/changes#diff-7341a4c8b11a2e7e32215f48cda97caaafcd5faee07bdb8e3c1a44b7e2339e60">+131/-3</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>config_output_no_projects.html</strong><dd><code>Add global reset button and styles for no-projects dialog</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1348/changes#diff-7f80710ed2abc11c038fb7e756dfb95c1225bee7e380cd741655c84c868c72ed">+131/-3</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>config_output_single_solution.html</strong><dd><code>Add global reset button and styles for single-solution dialog</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1348/changes#diff-40f815485d4acec22847d1036425f9b1a5ebf38e8043df9856e600654eb44f37">+131/-3</a>&nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Refactoring</strong></td><td><details><summary>4 files</summary><table>
<tr>
  <td><strong>notification.go</strong><dd><code>Update notification calls to use refactored BuildLspConfiguration</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1348/changes#diff-3ac97f07d29586cd80b9c63b5ac84fa3c0c029c0786ad485ae98d5d143cdaa53">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>folder_handler.go</strong><dd><code>Refactor BuildLspConfiguration signature and folder config sending</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1348/changes#diff-85f453996002ebf21ce07aa3db62fcd3d18a1861f517804721583decb421a90d">+6/-6</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>ldx_sync_service.go</strong><dd><code>Update LdxSyncService to use refactored BuildLspConfiguration</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1348/changes#diff-96079acff1e919961c5c93b80089075c62ed6b1f044fcf18fa0021dcf86fd30b">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>toggle_tree_filter.go</strong><dd><code>Update to use refactored BuildLspConfiguration</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1348/changes#diff-e500bc46809db21546b95489e9a02d395eea37ce90cef68c55b57b5c518975d3">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Tests</strong></td><td><details><summary>3 files</summary><table>
<tr>
  <td><strong>folder_handler_test.go</strong><dd><code>Add tests for refactored folder config handling</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1348/changes#diff-657ad7fa06e97a16b4e8eb7cddc64cd5e53343e6513ef874563e83ffd694416b">+9/-9</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>config_writers_test.go</strong><dd><code>Add tests for global unset and override checks</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1348/changes#diff-6ba83454376bc5616865a7e482043abdbbcc82883691d21d14cae8227c227b96">+92/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>global-reset.test.mjs</strong><dd><code>Add comprehensive JS unit tests for global reset</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1348/changes#diff-d9b7ba6055783b5f12fd87ec5311a4ecda34e809185b86a2038227fea56d45ea">+382/-0</a>&nbsp; </td>

</tr>
<tr>
  <td><strong>config_readers.go</strong><dd><code>Handle GAF deletion sentinel for reset fallbacks</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1348/changes#diff-3b315cd032c0bc1cd1052fd4bb3cc0c762129e8d5cd03c1ef15f42cd926c4588">+17/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Styling</strong></td><td><details><summary>1 files</summary><table>
<tr>
  <td><strong>styles.css</strong><dd><code>Add CSS for global reset button and info-box layout</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1348/changes#diff-0b5c07798ce2a45f8a12c869905292fb643e8f4359a2038c67545004f26a4f65">+32/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Documentation</strong></td><td><details><summary>2 files</summary><table>
<tr>
  <td><strong>configuration-dialog.md</strong><dd><code>Document global reset feature and UI interaction</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1348/changes#diff-625ca061d0507e9b5950bd6a809d4b475db62aed1cccbb8b072f3a873a6e4903">+6/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>configuration.md</strong><dd><code>Document global reset mechanism and contract</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1348/changes#diff-17ed18489a956f326ec0fe4040850c5bc9261d4631fb42da4c52891d74a59180">+11/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Additional files</strong></td><td><details><summary>2 files</summary><table>
<tr>
  <td><strong>configuration_test.go</strong></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1348/changes#diff-e141faafb3f0507956b85e92ac3b584c7c706bd96e7f428c7ef95d0d8b2df564">+516/-2</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>login.go</strong></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1348/changes#diff-21b93267e7967e84d95e104e3b17f130b4b47b4a37ac176c5bba7dd4c35cee31">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

</details>

___


application/server/configuration.go
```diff
@@ -850,6 +858,74 @@ func applyAutoScan(conf configuration.Configuration, settings map[string]*types.
...
+	types.SettingSeverityFilterMedium:   true,
+	types.SettingSeverityFilterLow:      true,
+	types.SettingIssueViewOpenIssues:    true,
+	types.SettingIssueViewIgnoredIssues: true,
+	types.SettingRiskScoreThreshold:     true,
+}
+
+// applyGlobalResets clears user overrides for org-scope global ("Project Defaults")
+// settings sent as {changed:true, value:null}. For each handled key it Unsets the
+// user override (so the resolver falls back to LDX-Sync / flagset default) and
```
rrama: Nitpick: I'm not too keen on this wording, as it feels a bit misleading to sneak  talking about folder-level settings in without explicitly calling it out.

My friend AI has suggested:

> A more precise wording would be something like:
> 
> For each handled key it Unsets the user override so the resolver falls back to the next layer: for folder-scoped reads (GAF resolveFolder) that's the org-level LDX-Sync value if present, otherwise the flagset default; for global reads (GetGlobalBool/GetGlobalString) it reverts directly to the flagset default.
> 
> Worth noting the same "LDX-Sync / flagset default" phrasing also appears in a few other spots added by this PR, so if you want consistency you'd touch them too:
> 
> config_writers.go — the UnsetGlobalUser doc comment ("LDX-Sync remote-unlocked → flagset default") and the GlobalResettableSettings doc comment ("LDX-Sync / org / flagset default").
> configuration.md — the new "Global Reset" section.
> config.go — ResetOrganization's comment (though that one is about org specifically, which is its own path).
> The config_writers.go UnsetGlobalUser phrasing is actually the most misleading, since UnsetGlobalUser itself only touches user:global:<name> and the function is scope-agnostic — what it falls back to depends entirely on which reader consumes it.

application/server/configuration.go
```diff
@@ -850,6 +858,74 @@ func applyAutoScan(conf configuration.Configuration, settings map[string]*types.
...
+// {changed:true, value:null}. This is the global-scope analog of
+// folder_config.go::isFolderReset.
+func isReset(cs *types.ConfigSetting) bool {
+	return cs != nil && cs.Changed && cs.Value == nil
+}
+
+// globalResetFilterKeys is the subset of GlobalResettableSettings whose reset
+// must trigger a diagnostics refresh (severity filters, issue-view options, risk
+// score threshold), mirroring applySeverityFilter / applyIssueViewOptions.
+var globalResetFilterKeys = map[string]bool{
```
rrama, line 908: I really dislike that all the properties for a setting are spread in many files throughout the project, but I think Shawky had a tech-debt task to change this, so it's fine for now, but I wanted to state my upset.

application/server/configuration.go
```diff
@@ -850,6 +858,74 @@ func applyAutoScan(conf configuration.Configuration, settings map[string]*types.
 // Returns true when the global org actually changed and the LSP is initialized
 // so the caller (processConfigSettings → processFolderConfigs) can union the
 // affected workspace folders into the single resetSummaryPanelForOrgChange call.
+// isReset reports whether an incoming ConfigSetting is a reset marker:
```
rrama, line 877: Umm, your insert went in between the function docs...

application/server/configuration.go
```diff
@@ -850,6 +858,74 @@ func applyAutoScan(conf configuration.Configuration, settings map[string]*types.
...
+			if oldOrgId != newOrgId && conf.GetBool(types.SettingIsLspInitialized) {
+				analytics.SendConfigChangedAnalytics(conf, engine, logger, configOrganization, oldOrgId, newOrgId, triggerSource, configResolver)
+				refreshFoldersForGlobalOrgChange(ctx, conf, engine, logger, oldOrgId, newOrgId)
+				orgChanged = true
+			}
+		} else if types.HasGlobalUserOverride(conf, name) {
+			types.UnsetGlobalUser(conf, name)
+			if globalResetFilterKeys[name] {
+				filterReset = true
+			}
```
rrama, line 985: Missing analytics event for these.

application/server/configuration.go
```diff
@@ -850,6 +858,74 @@ func applyAutoScan(conf configuration.Configuration, settings map[string]*types.
...
+// score threshold), mirroring applySeverityFilter / applyIssueViewOptions.
+var globalResetFilterKeys = map[string]bool{
+	types.SettingSeverityFilterCritical: true,
+	types.SettingSeverityFilterHigh:     true,
+	types.SettingSeverityFilterMedium:   true,
+	types.SettingSeverityFilterLow:      true,
+	types.SettingIssueViewOpenIssues:    true,
+	types.SettingIssueViewIgnoredIssues: true,
+	types.SettingRiskScoreThreshold:     true,
+}
```
rrama, line 921: AI is suggesting some fields are missing (unchecked by me):

> **Missing diagnostics refresh for product-enablement and scan_net_new resets**
> 
> `globalResetFilterKeys` only includes severity-filter / issue-view / risk-score keys. When `applyGlobalResets` clears user overrides for product-enablement settings (`snyk_oss_enabled`, `snyk_code_enabled`, `snyk_iac_enabled`, `snyk_secrets_enabled`) or `scan_net_new`, it does NOT set `filterReset = true`, so `sendDiagnosticsForNewSettings` is never called for those resets. However, the corresponding typed appliers (`applyProductEnablement` at `application/server/configuration.go:728` and `applyDeltaFindings` at `application/server/configuration.go:838`) DO call `sendDiagnosticsForNewSettings` when the value changes during a normal (non-reset) apply. This means that after a global reset, if the effective value of a product-enablement or `scan_net_new` setting changes (e.g., `snyk_oss_enabled` was overridden to `false` and reverts to the default `true`), the IDE will show stale diagnostics until the next scan.

application/server/configuration_test.go
```diff
@@ -3002,3 +3002,95 @@ func TestApplyEnvironment_ValuePreservesEquals(t *testing.T) {
...
 	assert.Equal(t, "Zm9v==", os.Getenv("SNYK_TEST_ADDITIONAL_ENV_B64"))
 }
+
+// Test_ProcessConfigSettings_GlobalReset verifies the global "Project Defaults"
+// reset path: a {changed:true, value:null} payload for an org-scope global setting
+// Unsets the user override (effective value reverts to the flagset default) and
+// the entry is removed from the settings map so the typed appliers do not re-apply
+// it. Mirrors the per-folder reset coverage in TestFolderConfig_ApplyLspUpdate.
+func Test_ProcessConfigSettings_GlobalReset(t *testing.T) {
+	t.Run("reset clears bool + filter overrides and reverts to defaults", func(t *testing.T) {
```
rrama, line 3415: Would be good to assert that diags are sent.
You may need to set LS as inited for this.

application/server/configuration_test.go
```diff
@@ -3002,3 +3002,95 @@ func TestApplyEnvironment_ValuePreservesEquals(t *testing.T) {
...
+
+		settings := map[string]*types.ConfigSetting{
+			types.SettingOrganization: {Value: nil, Changed: true},
+		}
+		processConfigSettings(ctx, conf, engine, engine.GetLogger(), settings, analytics.TriggerSourceTest, cr)
+
+		assert.Empty(t, types.GetGlobalString(conf, types.SettingLastSetOrganization),
+			"last_set_organization cleared so a later set is not a no-op")
+		_, present := settings[types.SettingOrganization]
+		assert.False(t, present, "handled org reset removed from settings map")
```
rrama, line 3486: This check isn't as good as it could be, we can override the default func for GAF. AI has helped me write up how (I have not read it):

> Suggestion: strengthen the org-reset subtest using a fake org default (pattern already in this file)
> 
> In the "reset of organization reverts to fallback" subtest of Test_ProcessConfigSettings_GlobalReset (configuration_test.go), we currently only assert that last_set_organization is cleared and the key is removed from the settings map. That doesn't verify the actual purpose of config.ResetOrganization (config.go:671-674) — that the effective org falls back through GAF's resolution chain after the Unset.
> 
> We already use this exact pattern elsewhere in this file, in setupFolderConfigTest (configuration_test.go:918-919):
> 
> engineConfig.AddDefaultValue(configuration.ORGANIZATION, configuration.ImmutableDefaultValueFunction("test-default-org-uuid"))
> engineConfig.AddDefaultValue(configuration.ORGANIZATION_SLUG, configuration.ImmutableDefaultValueFunction("test-default-org-slug"))
> Registering a fake ORGANIZATION default in this subtest would let us assert the fallback directly. After ResetOrganization does conf.Unset(configuration.ORGANIZATION), GetGlobalOrganization (internal/types/config_readers.go:69-76) resolves conf.GetString(configuration.ORGANIZATION), which now hits the registered default instead of /rest/self:
> 
> config.SetOrganization(conf, "my-custom-org")
> // ... reset ...
> assert.Equal(t, "test-default-org-uuid", types.GetGlobalOrganization(conf),
>     "effective org reverts to the GAF default after reset")
> Two notes:
> 
> GetGlobalOrganization writes the resolved value back via conf.Set(configuration.ORGANIZATION, ...) (internal/types/config_readers.go:72-73), so don't follow this with an !conf.IsSet(ORGANIZATION) assertion.
> Separately, none of the subtests set SettingIsLspInitialized = true, so the orgChanged=true branch in applyGlobalResets (application/server/configuration.go:906-910) — analytics + refreshFoldersForGlobalOrgChange + the value folded into globalOrgChanged — and the filter-reset sendDiagnosticsForNewSettings path (application/server/configuration.go:922) are never exercised. If we want to cover those, we'd need the fuller setupFolderConfigTest-style scaffolding (fake default + IsLspInitialized=true + a workspace + a mock LdxSyncService, since refreshFoldersForGlobalOrgChange calls mustLdxSyncServiceFromContext/mustNotifierFromContext at application/server/configuration.go:966-967).

internal/types/config_writers.go
```diff
@@ -85,3 +85,47 @@ func SetGlobalDeferredFolderScope(conf configuration.Configuration, name string,
...
+}
+
+// GlobalResettableSettings are the org-scope "Project Defaults" settings that the
+// HTML settings page can reset back to their fallback value (LDX-Sync / org /
+// flagset default). It is the same set the per-folder reset uses
+// (form-handler.js FOLDER_RESET_FIELDS) with two deliberate differences:
+//   - preferred_org is folder-only and excluded here.
+//   - organization is global-only and included here; it is NOT a
+//     UserGlobalKey(SettingOrganization) override, so it is reset via
+//     config.ResetOrganization rather than UnsetGlobalUser (see applyGlobalResets).
```
rrama, line 134: Add a comment saying this should be kept in-sync with the JS, otherwise it may be missed.

infrastructure/configuration/template/styles.css
rrama, line 1: Something is wrong with the HTML or CSS, the button is too wide.
Image

Once fixed, please leave pics in the PR description.

js-tests/global-reset.test.mjs
rrama, line 1: AI comment (unchecked by me):

> Coverage gaps
> 1. The DOM-driven test relies on a subtle accident, and the assertion is weaker than it looks. getAndSaveIdeConfig (auto-save.js:47) first calls collectChangedData(), which only includes keys whose value differs from the dirty-tracker baseline. The 14 reset fields are then forced to null by applyGlobalResets regardless (form-handler.js:284-285), so the test passes — but it never verifies that the reset works in combination with normal changed-field collection or validation. If applyGlobalResets were accidentally moved before collectChangedData, or a key got dropped, the test would still catch the missing key, but it doesn't test interaction with actually-dirty fields.
> 
> 2. No test that the 14 JS fields match the Go GlobalResettableSettings. The list is hardcoded a third time here (js-tests/global-reset.test.mjs:11-26), duplicating GLOBAL_RESET_FIELDS (form-handler.js:251-266) and the Go slice. The test asserts the JS produces these 14, but since both the source and the test hardcode the same list, a drift between JS and Go is invisible. This matches the reviewer-bot's maintenance-risk flag — there's no cross-language sync check anywhere.
> 
> 3. The validation-failure path is untested. If validationInfo.isValid is false (auto-save.js:32), getAndSaveIdeConfig returns early and the reset is not applied — but markGlobalForReset already set the flag. There's no test for: reset marked → save blocked by validation → flag still set → next save applies it. Worth confirming this is the intended behavior.
> 
> 4. No test that a global reset coexists with a folder reset in the same payload. applyFolderResets runs right before applyGlobalResets (auto-save.js:50-57). No test exercises both at once to confirm the top-level nulls and the per-folder nulls don't interfere.
> 
> 5. The confirm-cancel branch is untested. handleGlobalOverrideReset aborts if the user declines the confirm. The DOM test only stubs confirm to return true (js-tests/global-reset.test.mjs:90-92); there's no negative test asserting that returning false leaves the flag cleared.
> 
> 6. The flag persists if save never happens. markGlobalForReset sets globalReset = true; it's only cleared inside applyGlobalResets (form-handler.js:287). If the user clicks reset then closes without saving, the flag stays set on the next save. No test covers this lifecycle.
> 
> The most actionable gap is 2. — given the list is the wire contract shared across Go/JS/IDE PRs, the absence of any automated Go↔JS sync check is the highest-risk omission.

rrama (commented):
Like your PR #1344 the reset button should cause the form to immediately save instead of waiting on the user to press save (or auto-save in VSC).

AI comment (unverified by me):

> Nit (latent inconsistency, not a current bug): GetGlobalSliceFilePath (config_readers.go:135-146) reads conf.Get(configresolver.UserGlobalKey(key)) directly and is the one global reader that doesn't go through userGlobalValue, so it lacks the new configuration.IsKeyDeleted guard added at config_readers.go:47-49.
> 
> This is safe today: its only callers use SettingTrustedFolders (configuration.go:962, workspace.go:253), which isn't in GlobalResettableSettings, so UnsetGlobalUser never writes a tombstone to a key it reads. And even if a tombstone did appear, it fails both the *LocalConfigField and []FilePath type assertions and returns nil — the correct "absent" result.
> 
> The risk is purely future-facing: if a slice-typed setting is ever added to GlobalResettableSettings, this reader would silently diverge from GetGlobalBool/GetGlobalString/GetGlobalInt. Worth either adding the same IsKeyDeleted guard here for symmetry, or a short comment noting why it's intentionally omitted.

application/server/configuration.go
basti-snyk: Agreed — the spread of per-setting properties across files isn't ideal. Leaving as-is here since this PR follows the existing pattern, and it's covered by Shawky's tech-debt task to consolidate. Thanks for flagging.

infrastructure/configuration/template/js/ui/reset-handler.js
```diff
@@ -100,6 +106,25 @@
 		}
 	}
 
+	// Handle global (Project Defaults) override reset button click
+	function handleGlobalOverrideReset() {
+		if (!confirm("Reset all Project Defaults overrides? Your custom overrides will be removed.")) {
```
rrama: I have just noticed the other reset PR (#1344) states:

> **No-op in sandboxed VSCode webview.** The dialog renders in a sandboxed iframe without `allow-modals`, so `window.confirm()` is silently ignored and returns `false`. Both reset handlers gated their work behind `confirm()`, so clicking the button bailed and nothing happened. Removed the `confirm()` prompts (reset is reversible and the dialog shows a disclaimer banner).

So we may need to remove the `confirm` dialog.

application/server/configuration.go
basti-snyk: Fixed — `applyGlobalResets` now emits `SendConfigChangedAnalytics` for every reset key whose effective value changes, via a type-correct reader (`globalEffectiveValue`) so int/string-registered keys aren't silently skipped.

application/server/configuration.go
basti-snyk: Confirmed and fixed — added the product-enablement keys and `scan_net_new` to `globalResetFilterKeys` so a reset triggers `sendDiagnosticsForNewSettings`. Covered by a new test asserting `HandleConfigChange` fires after reset.

application/server/configuration.go
basti-snyk: Fixed — moved the `applyOrganization` doc comment back above its function so the docs are contiguous again.

internal/types/config_writers.go
basti-snyk: Done — added a KEEP IN SYNC comment on `GlobalResettableSettings`, plus an automated JS test that fails if it drifts from `GLOBAL_RESET_FIELDS`.

js-tests/global-reset.test.mjs
basti-snyk: Addressed the high-value gaps: a Go↔JS sync test that parses the production `form-handler.js` and asserts it matches the Go list; plus validation-failure deferral, global+folder coexistence, and flag-lifecycle tests. Confirm-cancel no longer applies (`confirm()` removed — see review reply).

application/server/configuration_test.go
basti-snyk: Done — the subtest now registers a fake ORGANIZATION default and asserts `GetGlobalOrganization` reverts to it after reset, with `SettingIsLspInitialized=true` so the analytics/refresh branch is exercised.

application/server/configuration_test.go
basti-snyk: Done — added an assertion that diagnostics are sent (`HandleConfigChange` via a MockWorkspace), LS initialized, using `require.Eventually` instead of a sleep.

application/server/configuration.go
basti-snyk: Reworded — the comments now distinguish folder-scoped reads (GAF `resolveFolder`) from global reads, and `UnsetGlobalUser`'s comment now notes it's scope-agnostic (only touches `user:global:<name>`; fallback depends on the reader).

infrastructure/configuration/template/styles.css
basti-snyk: Fixed in `styles.css` (replaced the float with inline-block + flex alignment). I'll add before/after screenshots to the PR description.

infrastructure/configuration/template/js/ui/reset-handler.js
rrama: I realised on the way home the solution is trivial and we just need a bridge to the native VS Code dialog, I will do this as a follow on PR.

basti-snyk: This PR already removes the confirm() for exactly that reason — the sandboxed-webview no-op you quoted from #1344 — and the global reset now persists immediately via a direct getAndSaveIdeConfig() call. A native VS Code dialog bridge as a follow-on sounds great, thanks!

rrama: Will do in https://github.com/snyk/snyk-ls/pull/1350 and https://github.com/snyk/vscode-extension/pull/771.

P.S. I do not like the AI saying "sounds great, thanks!" 🤮

infrastructure/configuration/template/styles.css
rrama: Better but now it goes to the next line for no reason.
Image

application/server/configuration.go
```diff
@@ -852,6 +874,127 @@ func applyAutoScan(conf configuration.Configuration, settings map[string]*types.
...
+// GetGlobalBool is wrong for int-registered (SettingRiskScoreThreshold) and
+// string-registered (SettingScanNetNew) keys — calling it there always yields
+// false regardless of the stored value, silently suppressing analytics.
+// All other GlobalResettableSettings are bool-registered, so they use GetGlobalBool.
+func globalEffectiveValue(conf configuration.Configuration, name string) any {
+	switch name {
+	case types.SettingRiskScoreThreshold:
+		return types.GetGlobalInt(conf, name)
+	case types.SettingScanNetNew:
+		return types.GetGlobalString(conf, name)
```
rrama: AI finding (unchecked by me):

> **globalEffectiveValue uses GetGlobalString for SettingScanNetNew but the value is stored as a bool, suppressing analytics**
> 
> `globalEffectiveValue` (`application/server/configuration.go:895`) reads `SettingScanNetNew` via `GetGlobalString`. However, in production `applyDeltaFindings` (`application/server/configuration.go:854`) writes a `bool` to this key via `SetGlobalDeferredFolderScope`. When `GetGlobalString` → `userGlobalValue` returns the raw `bool`, the `v.(string)` type assertion fails and `GetGlobalString` falls through to `conf.GetString(key)` returning `""` (the pflag default). Both `oldEffective` and `newEffective` resolve to `""`, making `effectivelyChanged` always `false` — analytics for this key are silently suppressed even when the effective value genuinely changed.
> 
> <details><summary>Type mismatch detail</summary>
> 
> The test at line 3436 (`configuration_test.go`) masks this by writing a string via `SetGlobalUser(conf, SettingScanNetNew, "enabled")`, but the production write path (`applyDeltaFindings`) stores a bool. The reset itself still works (UnsetGlobalUser fires unconditionally) and diagnostics refresh correctly (filterReset is set), but the `SendConfigChangedAnalytics` call at line 983 is dead code for this key.
> </details>

rrama (approved):
One AI finding that's small and one visual issue still. Other than that, I cba to re-read the tests AI says they are good now, so approved.

PR #1349, opened 2026-06-22
feat: show folder trust in tree view [IDE-1882]

andrewrobinsonhodges-snyk:
### Description

Folder trust settings are now shown in the tree view. If multiple folders are open and untrusted (or there is a mixture of trusted and untrusted folders) we show a single message, with a button per folder that needs trust.

The old modal dialog has been removed.

image

### Checklist


domain/ide/command/folder_handler.go
```diff
@@ -38,18 +35,15 @@ import (
...
+func HandleFolders(conf configuration.Configuration, engine workflow.Engine, logger *zerolog.Logger, notifier noti.Notifier, persister persistence.ScanSnapshotPersister, agg scanstates.Aggregator, featureFlagService featureflag.Service, configResolver types.ConfigResolverInterface) {
 	initScanStateAggregator(conf, agg)
 	initScanPersister(conf, logger, persister)
 	populateFolderFeatureFlagsAndSastSettings(conf, engine, logger, featureFlagService, configResolver)
 	sendFolderConfigs(conf, engine, logger, notifier, featureFlagService, configResolver)
 
-	HandleUntrustedFolders(ctx, conf, logger, srv)
+	// Untrusted folders are surfaced by the tree-view trust banner (IDE-1882), which
+	// the client renders/pulls via snyk.getTreeView — there is no longer a modal
+	// window/showMessageRequest trust dialog here.
```
rrama: I do not need this code comment talking about the dead.

domain/ide/command/trust_workspace_folders.go
```diff
@@ -38,20 +37,80 @@ func (cmd *trustWorkspaceFoldersCommand) Command() types.CommandData {
...
+
+// folderPathArg extracts an optional folder-path string from the command
+// arguments. Returns ("", false) when no usable path argument is present.
+func folderPathArg(args []any) (types.FilePath, bool) {
+	if len(args) == 0 {
+		return "", false
+	}
+	path, ok := args[0].(string)
+	if !ok || path == "" {
+		return "", false
```
rrama: Probably safer to only return false when there is no arg, in case we screw up the coding and send an object instead of a bare string and accidentally trust all the untrusted folders even though the user only said to open one.

domain/ide/command/trust_workspace_folders.go
```diff
@@ -38,20 +37,80 @@ func (cmd *trustWorkspaceFoldersCommand) Command() types.CommandData {
...
+	return types.FilePath(path), true
+}
+
+// filterFoldersByPath returns the folders whose path exactly matches the given
+// path. The banner sends back the same path string it received in FolderPaths,
+// so an exact match is sufficient.
+func filterFoldersByPath(folders []types.Folder, path types.FilePath) []types.Folder {
+	var matched []types.Folder
+	for _, f := range folders {
+		if f.Path() == path {
```
rrama: Do we not need to normalise the path with `PathKey`? in case of trailing slashes etc.

domain/ide/treeview/tree_builder.go
```diff
@@ -146,13 +160,77 @@ func (b *TreeBuilder) BuildTree(workspace types.Workspace) TreeViewData {
...
+		data.Nodes = append([]TreeNode{buildUntrustedFolderBanner(untrusted)}, data.Nodes...)
+	}
+
+	return data
+}
+
+// untrustedFolderRationale is the banner copy explaining why a folder must be
+// trusted before it is scanned. It carries forward the wording of the former
+// window/showMessageRequest trust dialog, which the in-tree banner replaced.
+// (IDE-1882)
```
rrama: Nitpick: I don't need a history lesson.

domain/ide/command/trust_workspace_folders.go
rrama: AI comment (unchecked by me):

> **Dead `notifier` field on `trustWorkspaceFoldersCommand`**
> 
> The `notifier` field is still declared on the struct (`trust_workspace_folders.go:32`) and injected by the factory (`command_factory.go:82`), but `Execute` no longer uses it — the `SnykTrustedFoldersParams` notification is now sent internally by `ws.TrustFoldersAndScan` (`workspace.go:275`) via the workspace's own `w.notifier`. Both notifiers originate from the same DI source so this doesn't cause a behavioral difference, but the field is dead code that could be cleaned up.

domain/ide/command/folder_handler.go
```diff
@@ -195,61 +189,3 @@ func initScanPersister(conf configuration.Configuration, logger *zerolog.Logger,
...
-	}
-	// debounce requests from overzealous clients (Eclipse, I'm looking at you)
-	if w.IsTrustRequestOngoing() {
-		return
-	}
-	_, untrusted := w.GetFolderTrust()
-	if len(untrusted) > 0 {
-		go func() {
-			w.StartRequestTrustCommunication()
-			defer w.EndRequestTrustCommunication()
```
rrama, line 212: AI comment (unchecked by me):

> **`TrustRequester` interface methods are now orphaned**
> 
> The `TrustRequester` interface (`internal/types/trust_requester.go:19-23`) with `StartRequestTrustCommunication`, `EndRequestTrustCommunication`, and `IsTrustRequestOngoing` has no remaining production callers after removing `HandleUntrustedFolders`. It's still embedded in the `Workspace` interface (`internal/types/workspace.go:44`), still implemented in `workspace_trust.go`, and still tested in `workspace_trust_test.go`. These are exclusively dead code now. The `.windsurfrules` file says "if files are not used or needed anymore, delete them instead of deprecating them." — but since these are interface methods (not files) this is a borderline case left at the author's discretion.

domain/ide/treeview/tree_builder_test.go
rrama, line 1: AI found coverage caps (unchecked by me):

> 1. Multiple untrusted folders (banner aggregation). Every new test uses exactly one untrusted folder. The banner's FolderPaths is built by ranging over all untrusted folders (tree_builder.go:192-202), and buildUntrustedFolderNodes emits one node each (tree_builder.go:209-219). Nothing verifies that with 2+ untrusted folders the banner lists all paths and that one dimmed node per folder is emitted in order. Given the PR's whole premise is "multiple folders ... a button per folder," this is the most notable gap. The JS test (js-tests/tree-runtime.test.mjs) and tree_html_test.go do cover multi-path rendering, but the Go builder doesn't.
> 
> 2. Untrusted folder that doesn't implement FilteringIssueProvider. In the all-untrusted test, untrusted is a bare MockFolder (not wrapped in fakeFilteringFolder). It's skipped early by the untrustedSet check at tree_builder.go:112-114 before the FilteringIssueProvider assertion at line 115, so the test passes — but it doesn't distinguish "skipped because untrusted" from "skipped because not a FilteringIssueProvider." Not a bug, just slightly weak as an assertion.

domain/ide/treeview/tree_html_test.go
```diff
@@ -106,6 +106,48 @@ func TestTreeHtmlRenderer_FileNode_HasDataAttributes(t *testing.T) {
...
+	assert.Contains(t, html, "tree-node-info--untrusted-folder")
+	assert.Contains(t, html, "You should only scan folders you trust")
+	// One Trust button per folder, each carrying its own folder path so the JS
+	// handler can scope snyk.trustWorkspaceFolders to that folder.
+	assert.Contains(t, html, `data-action="trust-folder" data-folder-path="/repo/a"`)
+	assert.Contains(t, html, `data-action="trust-folder" data-folder-path="/repo/b"`)
+	assert.Equal(t, 2, strings.Count(html, `data-action="trust-folder"`), "expected one Trust button per untrusted folder")
+}
+
+func TestTreeHtmlRenderer_UntrustedFolderNode_DimmedAndNoChevron(t *testing.T) {
```
rrama, line 131: AI found nitpick:

> The "DimmedAndNoChevron" test doesn't verify dimming or the chevron mechanism. Despite the name and comment (lines 144-147), the test only does one assertion:
> 
> assert.Contains(t, html, `class="tree-node tree-node-untrusted"`)
> This confirms the class list is exactly tree-node tree-node-untrusted (and crucially lacks tree-node-has-children, which is the real point). But:
> 
> It never asserts the dimming behavior. The dimming lives in CSS (.tree-node-untrusted > .tree-node-row { opacity: 0.5; }). The comment acknowledges the stylesheet is embedded, so the test can't easily isolate it — but the test name promises "Dimmed" and verifies nothing about it.
> The "no chevron" claim is only verified indirectly via absence of tree-node-has-children in that exact class string. That's a reasonable proxy, but it's coupled to the precise class ordering/spacing rather than checking the rendered chevron element.
> This is a naming/intent mismatch, not a correctness bug — the assertion that does run is meaningful.

domain/ide/treeview/tree_html_test.go
rrama, line 1: AI found coverage gaps:

> 1. No test that the untrusted node omits data-file-path or other expand-related attributes / children container. The JS guard relies on .tree-node-untrusted (tree.js:515-519), and the Go builder gives the node WithFilePath (tree_builder.go). The test passes WithFilePath("/repo/my-project") but never asserts what happens to it — so nothing pins down whether an untrusted folder node should/shouldn't carry a file path that other handlers act on.
> 
> 2. No escaping test for folder paths in the banner. This file has a strong XSS escaping test for labels/descriptions (tree_html_test.go:544-581), but FolderPaths are rendered into both data-folder-path="{{.}}" attributes and a title/label. A path containing a " could break out of the attribute. Worth a focused test given the security posture of the rest of the file.

The AI is wrong on the last point, but a test to prove it would be good.

domain/ide/treeview/tree_builder.go
```diff
@@ -96,8 +96,22 @@ func (b *TreeBuilder) BuildTree(workspace types.Workspace) TreeViewData {
 		return TreeViewData{}
 	}
 
+	// Untrusted folders are represented only by the workspace-trust banner at the
+	// top of the tree (IDE-1882); they are not scanned, so we exclude them from
+	// the tree body. GetFolderTrust already accounts for trust_enabled (every
+	// folder is trusted when the feature is off), so this is a no-op banner-wise
+	// when trust is disabled.
+	_, untrusted := workspace.GetFolderTrust()
```
rrama, line 104: Not a regression from you, but AI did notice:

> GetFolderTrust() doesn't take the mutex that Folders() uses (workspace.go:281 reads w.folders without w.mutex.RLock()). Pre-existing, not introduced by this PR, but the PR adds a new caller on the render path (tree_builder.go:104), so a concurrent AddFolder/RemoveFolder during a tree build is a theoretical data race. Worth being aware of, though out of scope here.

rrama (approved):

domain/ide/command/folder_handler.go
andrewrobinsonhodges-snyk: Removed the comment. `32057f75`

domain/ide/treeview/tree_builder.go
andrewrobinsonhodges-snyk: Removed the history comment. `32057f75`

domain/ide/command/trust_workspace_folders.go
andrewrobinsonhodges-snyk: Confirmed unused — removed the `notifier` field and its factory injection. `b012299e`

domain/ide/command/trust_workspace_folders.go
andrewrobinsonhodges-snyk: Good call — a present-but-malformed arg (wrong type or empty string) now takes the safe path (trust nothing) rather than falling through to trust-all; only a truly absent arg trusts every untrusted folder. `cbe53e77`

domain/ide/command/trust_workspace_folders.go
andrewrobinsonhodges-snyk: Yes — now normalising both sides of the comparison with `PathKey` so trailing slashes etc. match. Added a trailing-slash test. `cbe53e77`

domain/ide/treeview/tree_html_test.go
andrewrobinsonhodges-snyk: Strengthened the test — it now explicitly asserts there's no chevron/`tree-node-has-children` class (robust to class ordering) and verifies the dimming hook (a `.tree-node-row` rendered under the `.tree-node-untrusted` element, which the opacity CSS targets). `ee68a8f0`

domain/ide/treeview/tree_html_test.go
andrewrobinsonhodges-snyk: Added a folder-path XSS-escaping test (confirms paths containing `"` and `<` are escaped in the `data-folder-path` attribute and the banner) plus an assertion on the untrusted node's file-path attribute. `ee68a8f0`

domain/ide/treeview/tree_builder_test.go
andrewrobinsonhodges-snyk: Added a Go builder test with 2+ untrusted folders asserting the banner lists all paths and one dimmed node per folder is emitted in order. `ee68a8f0`

domain/ide/treeview/tree_builder.go
andrewrobinsonhodges-snyk: Good catch — fixed in this PR rather than deferring. `GetFolderTrust` now takes `w.mutex.RLock()` before iterating `w.folders` (mirroring `Folders()`). Extended the same protection to `Issues`/`Issue`/`Clear` which had the same bare-iteration pattern, and added a mutex-ordering doc plus a `-race` regression test. `314cc83`

PR #1350, opened 2026-06-23
feat: route reset confirm dialogs through native IDE UI [IDE-1945]

rrama:
### Description

VSCode sandboxed webviews block window.confirm() silently (no allow-modals). Add ideBridge.confirm(message, callback) that routes through the existing executeCommand bridge to snyk.showConfirmationDialog in VSCode, falling back to window.confirm() in IDEs with no bridge (JetBrains, Eclipse). Both reset handlers (section and folder override) now gate their work behind this callback.

Only merge if the VSCode side is also mergable (https://github.com/snyk/vscode-extension/pull/771)

### Checklist

  - Not yet!
  - N/A
  - N/A

### DIFF
--- a/docs/configuration-dialog.md
+++ b/docs/configuration-dialog.md
@@ -92,12 +92,14 @@ The HTML does **not** use `__ideLogin__` / `__ideLogout__` — authentication an
 | `window.__IS_IDE_AUTOSAVE_ENABLED__` | If `true`, form changes trigger save via `auto-save.js` | No (default false) |
 | `window.__onFormDirtyChange__(isDirty)` | Dirty-state callback for tab chrome | No |
 | `window.__ideSaveAttemptFinished__(status)` | Save outcome: `success`, `validation_error`, `bridge_missing`, `error` | No |
+| `window.__ideConfirmationDialog__(message, callback)` | Show a native confirmation modal (VSCode sandboxed webview blocks `window.confirm`). The `callback` **must** be invoked with a boolean (`true` = confirmed, `false` = cancelled). Hosts that do not inject this function fall back to `window.confirm`. | No (optional; falls back to `window.confirm`) |
 
 **Calls from `ide/bridge.js`:**
 
 - Login: `__ideExecuteCommand__('snyk.login', [authMethod, endpoint, insecure])`
 - Logout: `__ideExecuteCommand__('snyk.logout', [])`
 - Save: `__saveIdeConfig__(jsonString)` where `jsonString` is **`JSON.stringify`** of the object from `form-handler.collectData()` (not `workspace/didChangeConfiguration` — the IDE maps into LSP config).
+- Confirm: `__ideConfirmationDialog__(message, callback)` — called by `ideBridge.confirm()`; `callback` is always invoked with a strict boolean (fail-closed: non-boolean results are logged as errors and treated as cancel).
 
 **IDE-callable helpers (on `window`):**
 
@@ -117,6 +119,12 @@ webview.window.__saveIdeConfig__ = async (jsonString: string) => {
 };
 webview.window.__IS_IDE_AUTOSAVE_ENABLED__ = true;
 webview.window.__onFormDirtyChange__ = (isDirty: boolean) => { /* update tab title */ };
+webview.window.__ideConfirmationDialog__ = (message: string, callback: (confirmed: boolean) => void) => {
+  // Show a native modal and invoke callback with a boolean.
+  // Only needed in sandboxed environments (e.g. VSCode) where window.confirm is blocked.
+  const confirmed = await vscode.window.showWarningMessage(message, "Yes", "No") === "Yes";
+  callback(confirmed);
+};
 ```
 
 See [Function Injection Flow](#function-injection-flow) for the detailed sequence.
@@ -445,7 +453,7 @@ sequenceDiagram
 - [ ] Execute `snyk.workspace.configuration` command
 - [ ] Use returned string as webview HTML
 - [ ] Create webview/browser component for display
-- [ ] Expose `window.__ideExecuteCommand__`, `window.__saveIdeConfig__`, and optional `__onFormDirtyChange__` / `__ideSaveAttemptFinished__`
+- [ ] Expose `window.__ideExecuteCommand__`, `window.__saveIdeConfig__`, and optional `__onFormDirtyChange__` / `__ideSaveAttemptFinished__` / `__ideConfirmationDialog__`
 - [ ] Display HTML content in webview
 
 ### Configuration Management
@@ -478,6 +486,11 @@ sequenceDiagram
 - [ ] Warn user before closing dialog with unsaved changes using `window.__isFormDirty__()`
 - [ ] Disable save button when form is clean
 - [ ] Enable save button when form is dirty
+
+### Native Confirmation Dialog (Optional, recommended for sandboxed webviews)
+- [ ] Implement `window.__ideConfirmationDialog__(message, callback)` for sandboxed environments (e.g. VSCode) where `window.confirm` is blocked
+- [ ] Invoke `callback` with a **boolean** (`true` = confirmed, `false` = cancelled) — non-boolean results are logged as errors and treated as cancel (fail-closed)
+- [ ] If not implemented, `ideBridge.confirm` falls back to `window.confirm` automatically
 
 ### Auto-Save (Optional)
 - [ ] Set `window.__IS_IDE_AUTOSAVE_ENABLED__ = true` before loading HTML
--- a/infrastructure/configuration/template/js/ide/bridge.js
+++ b/infrastructure/configuration/template/js/ide/bridge.js
@@ -103,17 +103,33 @@
 
 	/**
 	 * Show a confirmation dialog via the IDE native UI.
-	 * In VSCode the sandboxed webview blocks window.confirm(), so this routes through
-	 * the executeCommand bridge to show a native modal. Falls back to window.confirm()
-	 * in IDEs that inject no bridge (JetBrains, Eclipse, etc.).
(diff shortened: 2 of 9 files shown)


infrastructure/configuration/template/js/ide/bridge.js
```diff
@@ -101,6 +101,22 @@
...
+	 * Show a confirmation dialog via the IDE native UI.
+	 * In VSCode the sandboxed webview blocks window.confirm(), so this routes through
+	 * the executeCommand bridge to show a native modal. Falls back to window.confirm()
+	 * in IDEs that inject no bridge (JetBrains, Eclipse, etc.).
+	 * @param {string} message - Message to display
+	 * @param {function} callback - Called with true if confirmed, false if cancelled
+	 */
+	ideBridge.confirm = function (message, callback) {
+		if (typeof window.__ideExecuteCommand__ === "function") {
+			window.__ideExecuteCommand__("snyk.showConfirmationDialog", [message], callback);
```
basti-snyk: **Should-fix (correctness / cross-repo contract):** This routes confirmation through `snyk.showConfirmationDialog`, but that command has no handler anywhere in snyk-ls, and the whole feature rests on every IDE that injects `__ideExecuteCommand__` implementing it client-side and invoking `callback` with a boolean.

The failure mode is silent: if an IDE injects the bridge (so the `window.confirm` fallback is skipped) but does NOT implement `snyk.showConfirmationDialog` — or the underlying `workspace/executeCommand` promise rejects (command missing, webview torn down) — then `callback` is never invoked, the reset never runs, and the user gets no dialog and no error. Login/logout/navigation already prove `__ideExecuteCommand__` is injected by some hosts, so the fallback (which only triggers when no bridge exists at all) won't save these cases.

Please (a) confirm VSCode/JetBrains/Visual Studio/Eclipse each handle this command client-side, and (b) consider an error/rejection path so a failed dispatch surfaces feedback instead of silently dropping the reset.

— AI review

infrastructure/configuration/template/js/ui/reset-handler.js
```diff
@@ -65,14 +65,18 @@
 			return;
 		}
 
-		var defaults = sectionDefaults[section];
-		applyDefaults(defaults);
+		window.ConfigApp.ideBridge.confirm("Reset all settings in this section to defaults?", function (confirmed) {
+			if (!confirmed) return;
```
basti-snyk, line 75: **Should-fix (correctness):** `if (!confirmed) return;` assumes `confirmed` is a strict boolean, but the value comes back from `__ideExecuteCommand__` — i.e. whatever the IDE's command returns, deserialized from JSON. If a host returns a truthy non-boolean on cancel (e.g. an object `{confirmed:false}`, which is truthy, or a button-label string), this guard passes and the destructive reset runs even though the user cancelled. Same guard at line 103.

Suggest failing closed: `if (confirmed !== true) return;` so any non-`true` result (including `undefined` from a host that never marshals a return value) aborts the reset. Pair this with documenting the boolean callback contract for `snyk.showConfirmationDialog`.

— AI review

infrastructure/configuration/template/js/ide/bridge.js
rrama: Good catch. Reworked the design rather than patching the command path: confirm now routes through a dedicated `window.__ideConfirmationDialog__(message, callback)` UI bridge (same category as `__saveIdeConfig__`/`__onFormDirtyChange__`), not `__ideExecuteCommand__`. Its presence now honestly proves the host supports a native confirm; absence falls back to `window.confirm` (fail-safe). This also removes the accidental forward-to-LS of an unknown command. Boolean-callback contract documented in `docs/configuration-dialog.md` and `config_html.go`. Fixed in d3a050cc.

infrastructure/configuration/template/js/ui/reset-handler.js
rrama: Hardened, centralized in the bridge: `ideBridge.confirm` wraps the callback in a `done(result)` that logs an error on a non-boolean result and coerces fail-closed via `callback(result === true)`, so any non-`true` value (incl. `undefined`) aborts the reset. Handlers keep `if (!confirmed) return;` unchanged since `confirmed` is now guaranteed boolean. d3a050cc.

rrama:
Re: https://github.com/snyk/snyk-ls/pull/1350#issuecomment-4782353563

Addressed: (tests) added `js-tests/bridge.test.mjs` — asserts `__ideConfirmationDialog__(message, callback)` dispatch, `cb(false)` aborts with no save, `cb(true)` proceeds, non-boolean logs + fails closed, no-bridge `window.confirm` fallback, and `handleSectionReset` coverage (145 pass). (docs) documented the new `__ideConfirmationDialog__` UI-bridge function in `configuration-dialog.md` + `config_html.go`. (consistency) dissolved — no longer routing through `executeCommand`. (logout note) out of scope for this PR. Commits d3a050cc, 07447ee8.

infrastructure/configuration/template/js/ide/bridge.js
rrama: Switched to a new technique of just calling `window. __ideConfirmationDialog__()`.

infrastructure/configuration/template/js/ui/reset-handler.js
rrama: Done, I now confirm it is a boolean or log an error.

infrastructure/configuration/template/js/ide/bridge.js
```diff
@@ -101,6 +101,38 @@
...
+		function done(result) {
+			if (typeof result !== "boolean" && window.console && console.error) {
+				console.error("ideBridge.confirm: expected boolean result, got", result);
+			}
+			callback(result === true); // fail closed: only literal true confirms
+		}
+		if (typeof window.__ideConfirmationDialog__ === "function") {
+			window.__ideConfirmationDialog__(message, done);
+		} else {
+			done(window.confirm(message));
```
basti-snyk, line 132: **Should-fix (cross-repo deployment coupling):** This fallback re-introduces the exact silent-failure that the now-deleted `handleGlobalOverrideReset` comment warned about: *"No confirm() — VSCode sandboxed webviews silently return false for window.confirm, which would prevent the reset from ever firing."*

All three resets are now gated behind `ideBridge.confirm`. When a VSCode build ships this snyk-ls HTML but the extension has **not** yet injected `window.__ideConfirmationDialog__`, this branch runs `window.confirm(message)`, which returns `false` in the sandbox, so `confirmed=false` and the reset silently no-ops with no user feedback. The docs mark `__ideConfirmationDialog__` as *optional*, but for VSCode it is effectively required to avoid this regression.

Please confirm the VSCode extension release that injects `__ideConfirmationDialog__` is coordinated with (or precedes) this change, and consider documenting the rollout ordering so a version skew can't silently break all three resets in VSCode.

— AI review

basti-snyk, line 132: **Suggestion (defensive / fail-closed completeness):** The fallback calls `window.confirm(message)` unconditionally. The doc block promises fail-closed behavior, but if a sandboxed host injects neither `__ideConfirmationDialog__` nor a working `window.confirm` (some webviews stub it to `undefined`), this throws `TypeError: window.confirm is not a function` synchronously, `done` never runs, `callback` is never invoked, and the reset dies as an uncaught exception rather than failing closed.

Consider guarding: `if (typeof window.confirm === "function") { done(window.confirm(message)); } else { done(false); }`.

— AI review

infrastructure/configuration/template/js/ide/bridge.js
```diff
@@ -101,6 +101,38 @@
...
+	 *
+	 * @param {string} message - Message to display
+	 * @param {function} callback - Required. Called with true if confirmed, false if cancelled.
+	 * @throws {TypeError} If callback is missing or not a function (programmer error — fail fast).
+	 */
+	ideBridge.confirm = function (message, callback) {
+		if (typeof callback !== "function") {
+			throw new TypeError("ideBridge.confirm: callback is required and must be a function");
+		}
+		function done(result) {
```
basti-snyk, line 123: **Suggestion (contract robustness):** `done` is handed to `__ideConfirmationDialog__` with no once-guard. A host that invokes the callback more than once (double-clicked "Yes", or a re-entrant native modal) will run the reset+save body twice. The marks-clearing `finally` in `getAndSaveIdeConfig` makes the second run mostly idempotent, but a redundant save payload is still dispatched. A one-shot guard (`var called=false; if (called) return; called=true;`) inside `done` would enforce the single-invocation contract the JSDoc implies.

— AI review

docs/configuration-dialog.md
```diff
@@ -117,6 +119,12 @@ webview.window.__saveIdeConfig__ = async (jsonString: string) => {
 };
 webview.window.__IS_IDE_AUTOSAVE_ENABLED__ = true;
 webview.window.__onFormDirtyChange__ = (isDirty: boolean) => { /* update tab title */ };
+webview.window.__ideConfirmationDialog__ = (message: string, callback: (confirmed: boolean) => void) => {
```
basti-snyk, line 122: **Suggestion (docs — snippet does not parse):** This example uses `await` (line below) inside a non-`async` arrow function, which is a syntax error. An integrator copy-pasting the reference implementation gets code that fails to compile. Declare it async:

```ts
webview.window.__ideConfirmationDialog__ = async (message: string, callback: (confirmed: boolean) => void) => {
```

— AI review

infrastructure/configuration/template/js/ide/bridge.js
rrama: Will be merged with https://github.com/snyk/vscode-extension/pull/771

rrama:
Abandoned since we should use [`window/showMessageRequest`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_showMessageRequest).

PR #1352, opened 2026-06-23
fix: stabilize OAuth storage bridge last-write-wins under rapid token rotation [IDE-2104]

basti-snyk:
## Summary
- Fixes the flaky `Test_RegisterOAuthStorageBridge_SerializesRapidTokenRotations` (Windows CI, IDE-2104), where after rapid token rotations the *first* token was stored instead of the *last*.
- Root cause: when `credentialUpdateWorker` writes a token via `conf.Set`, the persisted OAuth key re-fires the storage-bridge callback, which calls `QueueCredentialUpdate` again and re-enqueues the same token behind newer ones. On Windows (15 ms clock resolution) all rotation tokens share an expiry, so `shouldUpdateToken` lets the stale re-enqueued copy overwrite the final token (last-write-wins violation).
- Fix: an `atomic.Pointer[string]` guard (`writingToken`). The worker marks the token it is applying (cleared via `defer`, so a panic cannot leave the guard armed) and `QueueCredentialUpdate` skips the re-entrant enqueue of that same token. A guard-drop is logged at Debug level (no token value).
- Regression test rewritten to be deterministic with `require.Never`: RED without the guard, GREEN with it, across `GOMAXPROCS` 1/2/4.

## Test plan
- [x] `go test ./infrastructure/authentication/... -race` green
- [x] Regression test proven RED-without-fix / GREEN-with-fix (50/50, -race, -cpu=1,2,4)
- [x] `make test` (full unit suite) green at this commit
- [ ] CI integration + smoke suites (run on PR)

---
This fix was produced by an automated flake-fix loop.


infrastructure/authentication/auth_service_impl.go
```diff
@@ -128,14 +143,42 @@ func (a *AuthenticationServiceImpl) credentialUpdateWorker(ctx context.Context)
...
-			a.updateCredentials(update.token, update.sendNotification, update.updateApiUrl)
+			// Advertise which token we are about to write so QueueCredentialUpdate can
+			// recognize and drop the re-entrant callback that fires when
+			// WriteTokenToConfig calls conf.Set(auth.CONFIG_KEY_OAUTH_TOKEN, …) and
+			// the key is persisted in storage (which triggers the bridge callback again).
+			// The clear is deferred so it always runs even if updateCredentials panics,
+			// preventing the guard from staying permanently armed for that token string.
+			func() {
+				a.writingToken.Store(&update.token)
+				defer a.writingToken.Store(nil)
```
rrama, line 154: AI comment (unverified by me):

> Action: please add a documentation comment here making the synchronous-callback assumption explicit. The guard is correct only because the storage callback fires synchronously on the worker's goroutine — verified in storage.go:135-138, where storage.Set invokes the callback inline under its own mutex, not via a goroutine. That's what guarantees writingToken is still set when the re-entrant QueueCredentialUpdate reads it at auth_service_impl.go:173 (this defer clear hasn't run yet). The existing comments at auth_service_impl.go:94-107 and :164-171 document the re-entrancy mechanism but never state this assumption. Please add a line such as:
> 
> > Correctness note: this guard assumes the storage callback fires synchronously on the worker's goroutine (see storage.go:135-138). If GAF ever dispatches callbacks asynchronously, the worker clears the token before the async callback reads it, the guard silently races, and the IDE-2104 re-entrancy must be broken elsewhere.

infrastructure/authentication/auth_service_impl.go
```diff
@@ -128,14 +143,42 @@ func (a *AuthenticationServiceImpl) credentialUpdateWorker(ctx context.Context)
...
+				a.updateCredentials(update.token, update.sendNotification, update.updateApiUrl)
```
rrama, line 155: AI comment (unverified by me):

> Non-blocking / follow-up (not for this PR):
> 
> The worker calls the lockless updateCredentials at auth_service_impl.go:155 rather than the locking UpdateCredentials (auth_service_impl.go:552-557). This is pre-existing behavior, unchanged by this PR — flagging only as a future cleanup, not a blocker.
> 
> The worker touches authCache, lastUsedToken, etc. in updateCredentials (auth_service_impl.go:559-628) without holding a.m. Serialization against the worker relies on the single-worker-goroutine design, but the worker is not serialized against the other lock-holding paths (authenticate at auth_service_impl.go:301/:215, logout, configureProviders) — those hold a.m, the worker does not.
> 
> A follow-up could switch the worker to UpdateCredentials, but it's not a clean drop-in:
> 
> It introduces an implicit invariant that the synchronous re-entrant storage callback (auth_configuration.go:159-176, which calls QueueCredentialUpdate) must never acquire a.m, or the worker self-deadlocks. That holds today (the callback only enqueues to a channel) but should be documented.
> It interacts with the new writingToken guard/defer lifetime at auth_service_impl.go:152-156 and would need its own -race verification.
> Suggest tracking as a separate ticket rather than expanding this flake-fix PR.

rrama (approved):

infrastructure/authentication/auth_service_impl.go
basti-snyk: Done in 953c5fdc — added a "Correctness assumptions" doc block on the guard: (1) it is correct only because storage.Set fires the callback synchronously on the worker goroutine (before the deferred writingToken clear), so an async dispatch would silently race the guard; and (2) the token-equality drop assumes same-token updates are interchangeable, with a note to revisit if a caller ever needs to refresh flags for an already-written token. Thanks for flagging.

infrastructure/authentication/auth_service_impl.go
basti-snyk: Agreed — this is pre-existing and out of scope for the flake fix, so I've left the worker calling the lockless updateCredentials as-is here. We'll track the worker/UpdateCredentials serialization cleanup (and the documented deadlock invariant) as a separate ticket rather than expanding this PR. Thanks for the detailed writeup.

PR #1360, opened 2026-06-23
fix: set prodsec orb back to @1

rrama:
### Description

We just had too many vulns reported due to the Go version that it upset the runner, since we bumped the Go version we're all good.

### Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  prodsec: snyk/prodsec-orb@1.2.20
+  prodsec: snyk/prodsec-orb@1
 
 jobs:
   security-scans:


.circleci/config.yml
```diff
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  prodsec: snyk/prodsec-orb@1.2.20
+  prodsec: snyk/prodsec-orb@1
```
basti-snyk, line 4: Non-blocking suggestion: floating to `@1` makes CI resolve the latest `1.x` of `snyk/prodsec-orb` at build time, so builds are no longer reproducible — a future `1.x` regression could change scan behavior or break the pipeline without a code change here. This is the org-standard pinning for the first-party prodsec orb and CI is green on this commit, so it's an accepted tradeoff. If reproducibility ever matters, consider a minor pin (`@1.2`) or a Renovate-managed exact pin. Note this reverts the recent `@1.2.20` pin that was added because a newer orb version broke CI — worth confirming the upstream issue is resolved (CI here is currently passing, which is good evidence it is).

— AI review

basti-snyk (approved):

PR #1362, opened 2026-06-24
feat: add issue view and risk score filters to the tree view [IDE-1870]

andrewrobinsonhodges-snyk:
### Description

Added a filter button to the tree view, where Ignored/Open issues can be toggled, alongside a Risk Score slider.

Gated on feature flags.

Works the same as the severity filters: changed view options apply to all open folders. We indicate if there is a mismatch.

Screenshot 2026-06-24 at 14 49 36
Screenshot 2026-06-24 at 14 49 45

### Checklist


domain/ide/treeview/template/tree.js
```diff
@@ -559,6 +559,97 @@
...
+    var riskValue = document.getElementById('riskScoreValue');
+    var updateRiskLabel = function(v) {
+      if (riskValue) { riskValue.textContent = (v === 0) ? 'All' : ('≥ ' + v); }
+    };
+    if (slider) {
+      slider.addEventListener('input', function() {
+        slider.removeAttribute('data-mixed');
+        updateRiskLabel(parseIntSafe(slider.value, 0));
+      });
+      slider.addEventListener('change', function() {
```
basti-snyk: **Should-Fix — mixed risk-score slider can't be committed at its displayed value.**

When folders disagree, `aggregateRiskScores` renders the slider at the highest threshold with `data-mixed="true"`, but a value is only committed on the `change` event. A user who agrees with the shown mixed value (e.g. 800) and tries to confirm it by dragging and releasing back on 800 produces a final value equal to the drag-start value — most browsers fire no `change`, so no command is sent and the mixed state never resolves. The natural "accept the shown value and align all folders" gesture is a no-op (alignment is only reachable by committing a different value first). Consider committing on popover-open when mixed, or treating a release on a `data-mixed` slider as a commit.

— AI review

domain/ide/command/toggle_tree_filter.go
```diff
@@ -123,18 +140,81 @@ func (cmd *toggleTreeFilter) applyIssueViewFilter(value string, enabled bool) er
...
+// filters are not part of the popover and are intentionally left untouched. All
+// writes happen before Execute's single HandleConfigChange/notify, so one reset
+// click costs one config-change cycle rather than one per control.
+func (cmd *toggleTreeFilter) applyResetFilters() {
+	conf := cmd.engine.GetConfiguration()
+	defaults := types.DefaultIssueViewOptions()
+	for _, f := range cmd.workspaceFolders() {
+		types.SetUserFolder(conf, f.Path(), types.SettingRiskScoreThreshold, 0)
+		types.SetUserFolder(conf, f.Path(), types.SettingIssueViewOpenIssues, defaults.OpenIssues)
+		types.SetUserFolder(conf, f.Path(), types.SettingIssueViewIgnoredIssues, defaults.IgnoredIssues)
```
basti-snyk: **Should-Fix (low blast radius) — reset/filter writes are not feature-flag-scoped.**

`applyResetFilters` (and `applyRiskScoreFilter`/`writeFilterToAllFolders`) write risk-score and issue-view config to *every* open folder, regardless of whether that folder's gating flag (`UseOsTestWorkflow` / `SnykCodeConsistentIgnores`) is on. A flag-off folder that never exposed these controls still gets its persisted config mutated and re-emitted over `$/snyk.configuration`. It's inert for filtering (the per-folder server filter is gated by the same flag) and the written value is the default, so blast radius is low — but it's a silent unscoped write. Consider skipping folders where the relevant flag is off, mirroring the build-side gating, or add a short comment documenting that the workspace-wide write is deliberate and benign. (Also: this duplicates the raw `SetUserFolder` writes that `types.SetIssueViewOptionsForFolder` already encapsulates — using the helper would keep the two write sites from drifting.)

— AI review

domain/ide/treeview/tree_scan_emitter.go
```diff
@@ -198,3 +231,40 @@ func aggregateSeverityFilters(filters []types.SeverityFilter) (types.SeverityFil
...
+	var mixed MixedIssueViewOptions
+	for _, o := range opts[1:] {
+		if o.OpenIssues != agg.OpenIssues {
+			mixed.OpenIssues = true
+		}
+		if o.IgnoredIssues != agg.IgnoredIssues {
+			mixed.IgnoredIssues = true
+		}
+	}
+	return agg, mixed
```
basti-snyk, line 266: **Suggestion — mixed issue-view checkbox resolves in an unpredictable direction.**

A mixed issue-view checkbox renders both `checked` (from `opts[0]`, an arbitrary first folder's value) and `indeterminate`. On click the browser clears indeterminate and flips `checked`, then the handler writes that boolean to all folders — so the resolution direction depends on `opts[0]`, and two identical-looking dashed checkboxes can resolve in opposite directions on first click. Note the sibling `aggregateRiskScores` was *deliberately* designed to return the highest threshold precisely so a mixed control sits at a defined position rather than an arbitrary folder's value; the checkbox path lacks that equivalent deliberate choice. Recoverable on next render, so a first-click papercut rather than data loss.

— AI review

domain/ide/treeview/tree_scan_emitter_test.go
```diff
@@ -505,12 +506,57 @@ func TestFilterState_IVO_PinsSingleFolderBehavior(t *testing.T) {
 	ws := config.GetWorkspace(conf)
 	fs := emitter.filterState(ws)
 
-	// Current behavior: exactly one folder's IVO is used wholesale (not aggregated).
-	// Folder iteration order is non-deterministic, so accept either folder's value.
-	got := fs.IssueViewOptions
-	matchesOneFolder := got == ivo1 || got == ivo2
-	assert.True(t, matchesOneFolder,
-		"IVO should equal exactly one folder's value (no aggregation); got %+v", got)
+	assert.True(t, fs.IssueViewOptionsEnabled, "flag on for a folder → section enabled")
```
basti-snyk, line 550: **Suggestion — lost negative-case coverage for popover flag-gating.**

The removed `TestFilterState_IVO_PinsSingleFolderBehavior` was the only test exercising `BuildFilterState`'s issue-view path with `SnykCodeConsistentIgnores` **off**; its replacement requires the flag **on**. The flag-gating guards that hide the popover for non-flagged folders now have no negative-case coverage — a regression that inverted/dropped a guard (exposing controls that filter nothing) would pass CI. Consider adding a flag-off assertion that `ShowFilterPopover` / `*Enabled` are false.

— AI review

internal/types/folder_config.go
```diff
@@ -215,7 +215,12 @@ func (fc *FolderConfig) ToLspFolderConfig() *LspFolderConfig {
 			OriginScope: ev.OriginScope,
 			IsLocked:    strings.Contains(ev.Source, "locked"),
 		}
-		if !isMeaningfulValue(ev.Value) {
+		// Risk score is always sent (even its 0 = "All" default), so the tree-view
+		// filter popover and the settings page stay in sync when it is reset to 0.
+		// Without this exemption isMeaningfulValue would drop int 0 and a reset
+		// would never reach the open settings window. Other filter settings are
+		// booleans, which isMeaningfulValue always keeps.
+		if name != SettingRiskScoreThreshold && !isMeaningfulValue(ev.Value) {
```
basti-snyk: **Suggestion (altitude) — one setting name hardcoded into the shared LSP converter.**

`if name != SettingRiskScoreThreshold && !isMeaningfulValue(...)` bolts an "always-send even when zero" policy for a single setting onto the shared `ToLspFolderConfig` loop. That loop already does per-setting dispatch (it consults `AnnotationWriteOnly` and a `switch name`). The right altitude is an annotation (e.g. an `AnnotationAlwaysSend` set at registration) or a switch case, so the next numeric filter that needs to send its zero default doesn't grow this into `name != A && name != B`, far from where the setting is registered. Minor — correctness is fine and it's one setting today.

— AI review

domain/ide/command/toggle_tree_filter.go
```diff
@@ -123,18 +140,81 @@ func (cmd *toggleTreeFilter) applyIssueViewFilter(value string, enabled bool) er
...
+func toInt(v any) (int, error) {
+	switch n := v.(type) {
+	case float64:
+		return int(n), nil
+	case int:
+		return n, nil
+	case int64:
+		return int(n), nil
+	default:
+		return 0, fmt.Errorf("unsupported numeric type %T", v)
```
basti-snyk, line 210: **Suggestion — `toInt` clamps range elsewhere but doesn't reject non-finite floats, and the `int`/`int64` arms are untested.**

Two minor points: (1) the `float64` branch does `int(n)` with no rejection of `NaN`/`Inf` — `int(math.NaN())` is implementation-defined in Go, and the downstream `[0,1000]` clamp catches range but not non-finite values. Unreachable from the slider (it emits integers), but the clamp itself is justified as defending a public LSP boundary, and the same argument applies here. (2) The `int`/`int64` arms exist only for in-process/direct callers and are exercised by no test (the LSP path only delivers `float64`). Either add a table case for them or drop the unused arms. Both are minor.

— AI review

domain/ide/treeview/template/tree.js
andrewrobinsonhodges-snyk: Fixed in tree.js — a click on a still-`data-mixed` slider now commits the displayed (highest) value and aligns all folders, so accepting the shown value resolves the mixed state without needing to drag to a different value. The existing input/change handlers clear data-mixed first on a real drag, so there's no double-commit.

— AI review

domain/ide/command/toggle_tree_filter.go
andrewrobinsonhodges-snyk: Addressed in toggle_tree_filter.go:
- `applyResetFilters` now writes the issue-view pair via `types.SetIssueViewOptionsForFolder`, so it can't drift from the canonical write site (risk score has no typed helper, so it stays a `SetUserFolder`).
- Kept the workspace-wide write to all folders (consistent with the severity/issue-view toggles) and documented why it's intentional and inert for flag-off folders: the per-folder server filter is gated by the same flag, so resetting a flag-off folder to the default changes nothing.

— AI review

domain/ide/command/toggle_tree_filter.go
rrama, line 1: Nitpick: `filterValue` is a terrible name. `filterName` (or `filterId`) is way clearer. I know you aren't introducing the bad name in this PR, but please do change it.

domain/ide/command/toggle_tree_filter.go
```diff
@@ -45,32 +45,49 @@ func (cmd *toggleTreeFilter) Command() types.CommandData {
...
+		filterValue, enabled, err := toggleArgs(args)
+		if err != nil {
+			return nil, err
+		}
 		if err := cmd.applyIssueViewFilter(filterValue, enabled); err != nil {
 			return nil, err
 		}
+	case "riskScore":
+		if len(args) < 3 {
+			return nil, fmt.Errorf("expected 3 arguments [filterType, filterValue, threshold], got %d", len(args))
```
rrama: `filterValue` is unused. You can get away with just 2 args.

domain/ide/command/toggle_tree_filter.go
rrama: Or would it be better to just only use 2 args and have them be called like "severity_critical", then just `case "severity_critical", "severity_high", etc.`.

domain/ide/command/toggle_tree_filter.go
andrewrobinsonhodges-snyk: Done in toggle_tree_filter.go:
- `toInt` now rejects NaN/±Inf in the `float64` branch (returns an error) — same public-LSP-boundary rationale as the [0,1000] clamp; `int(NaN)` being implementation-defined is no longer reachable.
- Added `TestToInt` covering float64, int, int64, NaN, ±Inf, and the unsupported-type error, so the int/int64 arms (kept for in-process callers) are now exercised.

— AI review

domain/ide/treeview/tree_scan_emitter.go
andrewrobinsonhodges-snyk: Addressed in `aggregateIssueViewOptions`: when an option is mixed, the aggregated (indeterminate) checkbox value is now pinned to that option's default (open→true, ignored→false) instead of opts[0]'s arbitrary value. That gives a deterministic first-click direction — the same idea as `aggregateRiskScores` returning the highest threshold — so two identical-looking mixed checkboxes always resolve the same way. Added a `TestAggregateIssueViewOptions` case asserting the pin-to-default behaviour.

— AI review

domain/ide/treeview/tree_scan_emitter_test.go
andrewrobinsonhodges-snyk: Added `TestFilterState_FlagsOff_HidesPopover`: builds a workspace with neither gating flag enabled and asserts `RiskScoreEnabled`, `IssueViewOptionsEnabled`, and `ShowFilterPopover` are all false (and the mixed flags unset). That restores the negative-case coverage the removed pinning test used to provide, so a dropped gate would now fail CI.

— AI review

internal/types/folder_config.go
andrewrobinsonhodges-snyk: Raised the altitude: added an `AnnotationAlwaysSend` ("config.alwaysSend") annotation, set on `SettingRiskScoreThreshold` at registration. `ToLspFolderConfig` now checks that annotation instead of the hardcoded `name != SettingRiskScoreThreshold`, so a future always-send setting opts in at its definition with no edit to the shared loop.

On the related `isMeaningfulValue` point (snyk-pr-review-bot): `isMeaningfulValue` is used only by this converter, so gating via the annotation here fully covers it — I kept `isMeaningfulValue` as a pure value check rather than teaching it about specific setting names.

— AI review

domain/ide/treeview/template/tree.js
```diff
@@ -559,6 +559,110 @@
...
+      if (popover.hidden) return;
+      if (e.key === 'Escape' || e.key === 'Esc' || e.keyCode === 27) { closePopover(); }
+    });
+
+    // The DOM `indeterminate` flag can't be expressed in HTML markup, so apply it
+    // here on load from the server-set data-mixed attribute.
+    var issueViewChecks = popover.querySelectorAll('input[type="checkbox"][data-filter-type="issueView"]');
+    for (var ci = 0; ci < issueViewChecks.length; ci++) {
+      if (issueViewChecks[ci].getAttribute('data-mixed') === 'true') {
+        issueViewChecks[ci].indeterminate = true;
```
rrama, line 626: Q: Why are these flags needed compared to just using attributes?

domain/ide/treeview/tree_scan_emitter.go
rrama: > so two identical-looking mixed checkboxes always resolve the same way.

No they do not, one resolves to true and the other to false, when a checkbox is dashed the standard (probably) is that clicking it causes everything to become enabled. This goes against that standard.

domain/ide/treeview/tree_scan_emitter_test.go
```diff
@@ -532,3 +605,50 @@ func TestFilterState_AggregateSeverityFilters_UsesFilters0AsBaseline(t *testing.
...
+		types.NewIssueViewOptions(false, false), // only OpenIssues differs
+	}
+	_, mixed2 := aggregateIssueViewOptions(disagree)
+	assert.True(t, mixed2.OpenIssues, "OpenIssues differs → mixed")
+	assert.False(t, mixed2.IgnoredIssues, "IgnoredIssues agrees → not mixed")
+
+	// A mixed option is pinned to its default, not opts[0]'s arbitrary value, so
+	// the rendered checkbox has a deterministic first-click direction. Here opts[0]
+	// has OpenIssues=false but the default is true, so the aggregate must report
+	// true; IgnoredIssues agrees (both true) and is preserved as-is.
```
rrama: See https://github.com/snyk/snyk-ls/pull/1362/changes#r3473746128

internal/types/folder_config.go
```diff
@@ -215,7 +215,16 @@ func (fc *FolderConfig) ToLspFolderConfig() *LspFolderConfig {
 			OriginScope: ev.OriginScope,
 			IsLocked:    strings.Contains(ev.Source, "locked"),
 		}
-		if !isMeaningfulValue(ev.Value) {
+		// Settings annotated AnnotationAlwaysSend are emitted even at their zero
+		// default (e.g. risk score 0 = "All"), so a reset syncs to an open settings
```
rrama, line 219: I am worried about this change, consider the scenario (needs manual verifying):
1. User opens HTML settings page for a project that is all defaults (so a blank risk score field).
2. User changes the org and causes an HTML settings page refresh.
3. I believe with this change this now causes the risk score field to be populated with a `0` for the project.
4. User changes some other setting and saves the page.
5. I believe this now may cause the risk score of `0` that got entered in step 3 to be saved as a folder override.

rrama (approved):

domain/ide/treeview/tree_scan_emitter.go
andrewrobinsonhodges-snyk: Good point - fixed

internal/types/folder_config.go
andrewrobinsonhodges-snyk: The risk score value will only be applied if it has changed (or a mixed value gets manually set), so I think this case is already handled

domain/ide/command/toggle_tree_filter.go
andrewrobinsonhodges-snyk: Done — renamed `filterValue` and went with your combined-token idea: args are now [token, value] where the token is "severity_high", "issueView_openIssues", "riskScore", or "reset" (so no more confusing filterValue, and the unused riskScore placeholder is gone). Dispatch goes through a single filterTokenToSetting map. Updated tree.js, docs/tree-view.md, the dev harness, and added a test pinning the token vocabulary so the HTML attributes / JS / Go map can't drift apart.

domain/ide/treeview/template/tree.js
andrewrobinsonhodges-snyk: The checkbox indeterminate ("dashed") state is a DOM *property* with no HTML-markup equivalent — there's no `indeterminate` attribute you can render server-side. So the server emits a plain `data-mixed="true"` attribute (which HTML can express) and this JS reads it on load to set the real `el.indeterminate = true` property. Same reason the slider uses `data-mixed`: we need a server-renderable marker that the client then translates into the actual DOM property / commit behaviour. If there were an HTML attribute for it I'd have just rendered that directly.

domain/ide/command/toggle_tree_filter.go
andrewrobinsonhodges-snyk: Done — addressed as part of the combined-token refactor: filterValue is removed entirely and riskScore is now 2 args (["riskScore", threshold]). See the rename thread above for the full shape.

domain/ide/treeview/tree_scan_emitter_test.go
andrewrobinsonhodges-snyk: Addressed via the linked thread — mixed checkboxes now pin to false so the first click enables everywhere (commit fixing aggregateIssueViewOptions), which is the dashed-checkbox convention you flagged. This test was updated to match.

PR #1364, opened 2026-06-25
fix(code): serialize AUTHENTICATION_ADDITIONAL_URLS writes to fix data race [IDE-2169]

basti-snyk:
## What & why [IDE-2169]

Fixes a data race on `AUTHENTICATION_ADDITIONAL_URLS` in the `code` package, surfaced as a flaky test.

**Root cause:** every `CodeConfig` instance and `updateCodeApiLocalEngine()` share one `engine.GetConfiguration()` object and each performed an unsynchronized read-modify-write on `AUTHENTICATION_ADDITIONAL_URLS`. A per-instance mutex was insufficient — distinct `*CodeConfig` instances each hold a different lock, leaving concurrent cross-instance and local-engine writes unprotected.

**Fix:** introduce a package-level `var authURLsMu sync.Mutex` in `codeconfig.go` that all writers of `AUTHENTICATION_ADDITIONAL_URLS` (`CodeConfig.SnykCodeApi()` and `updateCodeApiLocalEngine()`) acquire before their read-modify-write. The misleading per-instance `mu` field is removed. The `slices.Contains` idempotency guard is retained and also applied to `updateCodeApiLocalEngine` so repeated registrations don't accumulate duplicate URLs.

This is a real synchronization fix, not a test skip/retry.

## Tests
Replaced the single-instance test with three race-detector tests:
- `Test_SnykCodeApi_ConcurrentAccess_SameInstance` (50 goroutines)
- `Test_SnykCodeApi_ConcurrentAccess_CrossInstance` (two `*CodeConfig` sharing one engine — reproduces the production race)
- `Test_SnykCodeApi_ConcurrentWithLocalEngine` (`SnykCodeApi` racing `updateCodeApiLocalEngine`)

**Evidence:** RED (per-instance mutex / no guard) → `WARNING: DATA RACE` + count assertion fail; GREEN after fix at `-race -count=50`; full `make test` suite green at commit `2ee57b0`; lint 0, gofmt clean.

🤖 Opened by the automated flake-fix loop.

[IDE-2169]: https://snyksec.atlassian.net/browse/IDE-2169?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nick-y-snyk (approved):
Verified the fix, not just read it:

- Checked out the branch in a worktree, reverted `codeconfig.go`/`sast_local_engine.go` to the pre-fix state (parent of 2ee57b09) while keeping the new tests, and ran `go test -race` — reproduced the actual `WARNING: DATA RACE` plus duplicate-URL count mismatches (4, 14, 16 entries instead of 1) across all three new tests. Reapplied the fix and reran at `-race -count=20`: clean.
- Grepped this repo and go-application-framework for every writer of `AUTHENTICATION_ADDITIONAL_URLS`: only `CodeConfig.SnykCodeApi()` and `updateCodeApiLocalEngine()` write it, and both are now under `authURLsMu`. The only other consumers (GAF's `auth_header.go`/`response.go` middleware) are read-only, so no exposure outside the new lock's domain.
- Checked for reentrancy: no call path where one locked function calls the other, so no deadlock risk from the non-reentrant mutex.
- Confirmed all CI checks green (lint, race-test, smoke-race, integration on macOS/Linux/Windows).

One minor nit (non-blocking): the commit message/comment describe removing a "misleading per-instance mu field," but `main` never had one (checked `git show main:infrastructure/code/codeconfig.go` and the parent of 2ee57b09) — just a stale description, not a code issue.

Approving.

PR #1365, opened 2026-06-25
feat: add feedback banner to tree view [IDE-2008]

andrewrobinsonhodges-snyk:
### Description

Added a feedback banner to the tree view.

The banner appears 5 seconds after the user first interacts with the tree view.

Once dismissed, the banner does not show again.

The feedback link opens the web browser at the feedback form

image

### Checklist


application/config/config_test.go
```diff
@@ -976,6 +976,38 @@ func TestSetupStorage_FlushesPreInitOAuthTokenToNewStorage(t *testing.T) {
...
+	logger := zerolog.Nop()
+
+	storageFile := filepath.Join(t.TempDir(), "ls-config.json")
+	s, err := storage.NewStorageWithCallbacks(storage.WithStorageFile(storageFile))
+	require.NoError(t, err)
+
+	SetupStorage(conf, s, &logger)
+
+	SetFeedbackBannerDismissed(conf)
+	SetFeedbackBannerInteracted(conf)
```
rrama, line 993: Nit: the order of these two lines upsets me.

rrama: And it affects the test so it isn't a nit! Because the refresh is async, there is a race condition where the dismissed could be set but the interacted not. So they need to be flipped!

domain/ide/command/dismiss_feedback_banner.go
```diff
@@ -0,0 +1,42 @@
...
+	command types.CommandData
+	engine  workflow.Engine
+}
+
+func (cmd *dismissFeedbackBanner) Command() types.CommandData {
+	return cmd.command
+}
+
+func (cmd *dismissFeedbackBanner) Execute(_ context.Context) (any, error) {
+	config.SetFeedbackBannerDismissed(cmd.engine.GetConfiguration())
```
rrama, line 40: Why not just do this as a config update (command) and have it as a user global only config, then no command needed?

domain/ide/command/feedback_banner_interacted.go
```diff
@@ -0,0 +1,43 @@
...
+	command types.CommandData
+	engine  workflow.Engine
+}
+
+func (cmd *feedbackBannerInteracted) Command() types.CommandData {
+	return cmd.command
+}
+
+func (cmd *feedbackBannerInteracted) Execute(_ context.Context) (any, error) {
+	config.SetFeedbackBannerInteracted(cmd.engine.GetConfiguration())
```
rrama, line 41: Might be able to have this as a configuration command as well, as user global only, but I don't know if that forces it to be saved 🤔

rrama (approved):
Might be able to be done as just `snyk.configuration`, but this is fine.

application/config/config_test.go
andrewrobinsonhodges-snyk: fixed

domain/ide/command/dismiss_feedback_banner.go
andrewrobinsonhodges-snyk: Keeping the dedicated commands - refactoring requires a relatively large change and that's risky at this point.

domain/ide/command/feedback_banner_interacted.go
andrewrobinsonhodges-snyk: Keeping the dedicated commands - refactoring requires a relatively large change and that's risky at this point.

PR #1368, opened 2026-07-08
fix: stabilize Test_E2E_EndpointChangeAfterInitializationClearsToken flaky on windows [IDE-2179]

basti-snyk:
## Summary
Fixes flaky `Test_E2E_EndpointChangeAfterInitializationClearsToken` (windows-latest integration-tests, IDE-2179).

Root cause: the OAuth storage-bridge callback (`newOAuthStorageBridgeCallback`) propagated empty-token storage echoes through the serialized `credentialUpdateChan`. On slower runners a stale empty update could be drained by the worker *after* a later synchronous token write, clobbering the just-set token to empty and failing the assertion. The `writingToken` re-entrancy guard cannot catch empty echoes (empty is not a valid in-flight token).

Fix: `newOAuthStorageBridgeCallback` now early-returns on an empty token, keeping non-actionable empty echoes off the serialized channel at the boundary — mirroring the existing `updateCredentials` no-op guard. Credential clears remain correct: `logout` drives `updateCredentials("")` synchronously, independent of the bridge, so logout/clear propagation is unaffected. Real root-cause fix — no sleep/timeout/skip/retry workaround.

## Tests
- Two deterministic regression tests added (`auth_configuration_test.go`), confirmed RED before the fix / GREEN after.
- Stress: `go test ./application/server/ -run '^Test_E2E_EndpointChangeAfterInitializationClearsToken$' -race -count=100` → green.
- `go build ./...`, `golangci-lint run` (0 issues), and `make test` (unit) all green at HEAD.

## Known follow-up (non-blocking)
- Reviewers flagged a now-dead `token_empty` debug log field in `auth_configuration.go` (always false after the early return). Cosmetic only; left for a follow-up to keep this fix minimal.

_This fix was produced by an automated flake-fix loop._

### DIFF
--- a/infrastructure/authentication/auth_configuration.go
+++ b/infrastructure/authentication/auth_configuration.go
@@ -159,6 +159,23 @@ func newOAuthStorageBridgeCallback(authenticationService AuthenticationService) 
 	return func(_ string, value any) {
 		// an empty struct marks an empty token, so we stay with empty string if the cast fails
 		newToken, _ := value.(string)
+		// Drop empty-token storage updates. The bridge exists only to propagate a newly
+		// written/rotated (non-empty) OAuth token from storage to the auth service.
+		// Credential clears are always driven synchronously through updateCredentials /
+		// logout (service → storage), so an empty value arriving here (e.g. the OAuth key
+		// cleared during the init auto-auth logout) is only ever a redundant echo. Queuing
+		// it lets the async worker apply updateCredentials("") after a later, unrelated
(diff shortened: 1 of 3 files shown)


rrama (approved):

infrastructure/authentication/auth_configuration.go
rrama, line 182: Super minor nitpick: This `token_empty` log part can be removed, since it is now always false.

PR #1369, opened 2026-07-08
fix(auth): don't hold mutex during OAuth browser wait

nick-y-snyk:
## Summary

- `Authenticate()` held `a.m` for the full OAuth flow (up to 120 s browser wait). Re-clicking Authenticate did nothing until the first flow timed out.
- Root cause traced to GAF commit `d14fa1a` (Abdelrahman, Aug 2025) which added `globalRefreshMutex.Lock()` at the start of `CancelableAuthenticate` — before the `ctx.Done()` goroutine starts — creating a cancellation-blind window. Even past that window, `server.Shutdown(context.Background())` can block on live TCP connections, keeping `a.m` locked for the full 120 s.
- Fix: capture the provider under a brief `RLock`, run `provider.Authenticate` lock-free, re-acquire `a.m` only to save results. `ConfigureProviders` and a concurrent re-login now proceed immediately after `CancelOngoingAuth()`.

## Test plan

- [ ] 382 existing auth + command tests pass (`go test ./infrastructure/authentication/... ./domain/ide/command/...`)
- [ ] `TestLoginCommand_Execute_CanRestartAuthWhenPreviousInProgress` still passes (updated comment to reflect new mechanism)
- [ ] Manual: OAuth login → close browser → click Authenticate again → new browser link opens immediately (no 120 s wait)
- [ ] `make lint-fix` → 0 issues

### DIFF
--- a/infrastructure/authentication/auth_service_impl.go
+++ b/infrastructure/authentication/auth_service_impl.go
@@ -169,29 +169,40 @@ func (a *AuthenticationServiceImpl) provider() AuthenticationProvider {
 func (a *AuthenticationServiceImpl) Authenticate(ctx context.Context) (token string, err error) {
 	a.CancelOngoingAuth()
 
-	a.m.Lock()
-	defer a.m.Unlock()
-
 	a.previousAuthCtxCancelFuncMu.Lock()
 	ctx, a.previousAuthCtxCancelFunc = context.WithCancel(ctx)
 	a.previousAuthCtxCancelFuncMu.Unlock()
-
 	defer a.previousAuthCtxCancelFunc() // need to clean up resources if we weren't interrupted, impl should ensure its safe to double call
-	return a.authenticate(ctx)
-}
 
-func (a *AuthenticationServiceImpl) authenticate(ctx context.Context) (token string, err error) {
-	if a.authProvider == nil {
-		err = errors.New("authentication provider is not configured")
-		a.engine.GetLogger().Warn().Err(err).Msg("Failed to authenticate: auth provider is nil")
+	// Capture the provider under a brief read lock. ConfigureProviders takes a
+	// write lock, so releasing a.m before the slow browser wait lets a concurrent
+	// re-login reconfigure providers without waiting for the OAuth flow to time out.
+	a.m.RLock()
+	provider := a.authProvider
+	a.m.RUnlock()
+
+	if provider == nil {
+		a.engine.GetLogger().Warn().Msg("Failed to authenticate: auth provider is nil")
+		a.m.Lock()
 		a.authCache.RemoveAll()
-		return "", err
+		a.m.Unlock()
+		return "", errors.New("authentication provider is not configured")
 	}
 
-	token, err = a.authProvider.Authenticate(ctx)
+	// Run the auth flow without holding a.m. This lets ConfigureProviders and
+	// Logout proceed while waiting for the OAuth browser callback (up to 2 minutes).
+	// CancelOngoingAuth above signals any concurrent auth to abort.
+	token, err = provider.Authenticate(ctx)
 
+	// Re-acquire the write lock to save results.
+	a.m.Lock()
+	defer a.m.Unlock()
+	return a.authenticate(token, err, provider)
+}
+
+func (a *AuthenticationServiceImpl) authenticate(token string, err error, provider AuthenticationProvider) (string, error) {
 	if token == "" || err != nil {
-		a.engine.GetLogger().Warn().Err(err).Msgf("Failed to authenticate using auth provider %v", reflect.TypeOf(a.authProvider))
+		a.engine.GetLogger().Warn().Err(err).Msgf("Failed to authenticate using auth provider %v", reflect.TypeOf(provider))
 		a.authCache.RemoveAll()
 		return token, err
 	}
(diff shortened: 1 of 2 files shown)


infrastructure/authentication/auth_service_impl.go
```diff
@@ -212,29 +212,40 @@ func (a *AuthenticationServiceImpl) provider() AuthenticationProvider {
...
 	}
 
-	token, err = a.authProvider.Authenticate(ctx)
+	// Run the auth flow without holding a.m. This lets ConfigureProviders and
+	// Logout proceed while waiting for the OAuth browser callback (up to 2 minutes).
+	// CancelOngoingAuth above signals any concurrent auth to abort.
+	token, err = provider.Authenticate(ctx)
+
+	// Re-acquire the write lock to save results.
+	a.m.Lock()
```
basti-snyk, line 241: **Critical — result is persisted with no ordering guard, so the last goroutine to grab the lock wins.**

`provider.Authenticate(ctx)` now runs without `a.m`, then this block unconditionally re-acquires the lock and saves. Because `CancelOngoingAuth()` only *signals* (it does not wait for the older flow to finish), two consequences are newly possible that were structurally impossible when the write lock spanned the whole flow:

1. **Logout silently undone.** A login is parked in the browser wait. The user clicks Logout; `Logout` takes the lock, clears the token, returns. The login then completes, re-acquires the lock here, and re-persists a token — the user pressed Logout but ends up authenticated.
2. **Superseded / wrong-org login clobbers the newer one.** User starts login A (wrong org), restarts as B (right org). If A's browser callback already produced a valid token before cancellation propagates, A returns success; whichever of A/B grabs the lock *last* wins the save, so A can overwrite B.

Root fix: add an ordering guarantee before persisting — e.g. an auth generation/epoch counter so only the newest `Authenticate`'s result may write, plus bail on `ctx.Err() != nil` after the wait so a canceled/superseded flow discards its result and cannot resurrect a cleared session.

— AI review

infrastructure/authentication/auth_service_impl.go
```diff
@@ -212,29 +212,40 @@ func (a *AuthenticationServiceImpl) provider() AuthenticationProvider {
...
-		return "", err
+		a.m.Unlock()
+		return "", errors.New("authentication provider is not configured")
 	}
 
-	token, err = a.authProvider.Authenticate(ctx)
+	// Run the auth flow without holding a.m. This lets ConfigureProviders and
+	// Logout proceed while waiting for the OAuth browser callback (up to 2 minutes).
+	// CancelOngoingAuth above signals any concurrent auth to abort.
+	token, err = provider.Authenticate(ctx)
```
basti-snyk, line 238: **Should-Fix — two `provider.Authenticate` calls can now run concurrently on the same provider instance.**

With `a.m` released during the wait, `CancelOngoingAuth()` only signals the older flow; it does not drain it. So a second `Authenticate` can capture the same `provider` pointer and call `Authenticate` on it while the first is still blocked on its browser callback. If the OAuth provider is not single-flight safe (e.g. it binds a fixed loopback callback port or mutates shared flow state), the second listener fails to bind or the two callbacks cross wires. The removed full-duration lock previously guaranteed single-flight execution.

— AI review

infrastructure/authentication/auth_service_impl.go
```diff
@@ -212,29 +212,40 @@ func (a *AuthenticationServiceImpl) provider() AuthenticationProvider {
...
+	// Re-acquire the write lock to save results.
+	a.m.Lock()
+	defer a.m.Unlock()
+	return a.authenticate(token, err, provider)
+}
 
+func (a *AuthenticationServiceImpl) authenticate(token string, err error, provider AuthenticationProvider) (string, error) {
 	if token == "" || err != nil {
-		a.engine.GetLogger().Warn().Err(err).Msgf("Failed to authenticate using auth provider %v", reflect.TypeOf(a.authProvider))
+		a.engine.GetLogger().Warn().Err(err).Msgf("Failed to authenticate using auth provider %v", reflect.TypeOf(provider))
```
basti-snyk, line 248: **Should-Fix — a late-completing canceled flow wipes the winning flow's cache.**

On the failure path (`token == "" || err != nil`) this calls `a.authCache.RemoveAll()`. Now that flows overlap, a canceled/superseded login returning `context.Canceled` can reach this branch and acquire the lock *after* a successful concurrent login ran `authCache.Set(...)`, erasing the just-set validity entry (and all others). The next `IsAuthenticated()` misses the cache and is forced into a network re-check; offline or on a transient error it surfaces a false "not authenticated" for a validly logged-in user.

Suggest skipping the wipe for superseded flows, e.g. `if errors.Is(err, context.Canceled) { return token, err }` before `RemoveAll()`.

— AI review

infrastructure/authentication/auth_service_impl.go
```diff
@@ -212,29 +212,40 @@ func (a *AuthenticationServiceImpl) provider() AuthenticationProvider {
...
-		a.engine.GetLogger().Warn().Err(err).Msg("Failed to authenticate: auth provider is nil")
+	// Capture the provider under a brief read lock. ConfigureProviders takes a
+	// write lock, so releasing a.m before the slow browser wait lets a concurrent
+	// re-login reconfigure providers without waiting for the OAuth flow to time out.
+	a.m.RLock()
+	provider := a.authProvider
+	a.m.RUnlock()
+
+	if provider == nil {
+		a.engine.GetLogger().Warn().Msg("Failed to authenticate: auth provider is nil")
```
basti-snyk, line 228: **Suggestion — the nil-provider log dropped its `.Err(err)` field.**

The rewritten nil-provider branch logs `Warn().Msg("Failed to authenticate: auth provider is nil")`, whereas `main` attached the error with `.Err(err)`. Cosmetic (the error is still returned), but worth keeping for log parity.

— AI review

PR #1370, opened 2026-07-09
fix: keep settings config command serviceable during failing startup auth [IDE-2181]

basti-snyk:
## Summary
When the language server starts with an expired/revoked OAuth token, a failing token refresh at startup ran **synchronously inside the LSP `initialized` handler**, holding a jrpc2 dispatch worker and starving the `snyk.workspace.configuration` command that renders the IDE settings page — so the page timed out to its fallback and the user could not re-authenticate.

This moves the blocking scanner/auth initialization into a **background goroutine** so `initialized` returns promptly and the config-HTML command stays serviceable during a failing refresh (rendering config HTML needs no auth).

Because freeing dispatch widens the window where the LSP-initialized flag is `false`, the fix also:
- **Splits the init signal** — an early *handshake-acknowledged* flag (set synchronously) gates the endpoint-switch credential clear, so switching endpoint during startup still clears old-endpoint credentials (no session leakage); the late *scanner-ready* signal keeps scan-readiness waiters correct.
- Adds a **handler-level safety net** so the readiness signal still fires if a dependency resolution panics before the goroutine launches (no stranded `WaitForLspInitialized` waiters).
- Runs the background init on a **cancellable context** that shutdown cancels and **awaits** before disposing shared state, with **panic recovery** in the goroutine so a failing init cannot crash the server.

## Scope / PR size
Production change is **~230 lines** (`server.go`, `apply_auth_config.go`, `config_readers.go`, `ldx_sync_config.go`); the remaining ~690 lines are **TDD tests** (3 new test files) kept with their production code to preserve atomic-commit/TDD integrity — hence the total exceeds the usual 700-line target. Reviewer focus is the production delta.

## Verified (outside-in TDD, 4 review rounds)
- `dispatch_starvation_test.go` — config HTML serviced while startup token refresh fails (RED→GREEN, the ~60s starvation).
- `authentication_flows_e2e_test.go` — endpoint switch **during** the init window still clears old-endpoint credentials.
- `initialized_signal_safety_test.go` — readiness signal still fires when a dep resolution panics pre-launch.
- `background_init_lifecycle_test.go` — shutdown cancels+awaits background init; panic in init is recovered and the server survives.
- `make lint` 0 issues; `application/server` package GREEN under `-race` (unit + integration).

## Deferred follow-ups (non-blocking review suggestions)
- Thread `initCtx` into `command.HandleFolders` (or add an `initCtx.Err()` checkpoint before it) to shrink the 5s-shutdown-timeout window on that step.
- Make `cli.Initializer.Init` offline/install loops honor `ctx.Done()` so shutdown cancellation is fully cooperative.
- Optionally serialize the `backgroundInitCancel`/`backgroundInitDone` package vars (currently rely on LSP ordering + jrpc2 dispatch mutex, matching the pre-existing `cacheCheckCancel` pattern).

## Remaining for IDE-2181 (separate PRs, will stack on this)
- **snyk-ls**: add a working login control to the settings **fallback** page (`settings-fallback.html`).
- **snyk-intellij-plugin**: mirror that fallback and prove `snyk.login` dispatches from it.

## Test plan
- [ ] Start LS with an expired token; open IDE settings — settings page loads (not fallback) and re-auth works.
- [ ] Switch API endpoint during startup — old-endpoint credentials are cleared.
- [ ] Close the IDE during the startup init window — clean shutdown, no crash.

___

### **PR Type**
Bug fix, Enhancement

___

### **Description**
- Decouple heavy LSP initialization to background goroutine.

- Improve shutdown handling for background tasks.

- Clarify handshake and scanner-ready signals.

- Enhance resilience against startup auth failures.

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A[LSP Initialize/Initialized Call] --> B(Initialize Handler);
  B --> C{Launch Background Init};
  B -- Sync --> D[Acknowledge Handshake <br> (Set handshake ack)];
  C -- Async --> E[Background Init Goroutine <br> (Scanner Init, Folders, Scan)];
  E --> F[Complete Background Init <br> (Set scanner ready)];
  G[Shutdown Call] --> H{Cancel Background Init Context};
  H --> I[Await Background Init Done];
  J[Endpoint Change] --> K{Check Handshake Acknowledged};
  K -- True --> L[Clear Credentials];
  F --> D; # Scanner Ready implies Handshake Acked
```

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Tests</strong></td><td><details><summary>13 files</summary><table>
<tr>
  <td><strong>authentication_flows_e2e_test.go</strong><dd><code>Update E2E tests for async initialization</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1370/changes#diff-9deadd440a71de07926bf9936df138c9c5512696795d6b6544d4f40c01d06752">+84/-1</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>background_init_lifecycle_test.go</strong><dd><code>Add tests for background init lifecycle and recovery</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1370/changes#diff-9ac7991e169bbbd8c6d7308d51b9456ed79de34945463c8468ae4a6e53a56b58">+197/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>dispatch_starvation_test.go</strong><dd><code>Test config command serviceability during init</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1370/changes#diff-572416044efd8567b3ab04d8d958af3c4da2b396dee659fd9ba457705d895dcd">+183/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>execute_command_test.go</strong><dd><code>Wait for background init before clearing login token</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1370/changes#diff-62dafdad0c9b4059a5d99a1eb3f1fcf677e7720b1cac9f4f87ea711faeec86ab">+5/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>initialized_signal_safety_test.go</strong><dd><code>Test signal fires on dependency panic before init</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1370/changes#diff-c5959c2485893030fd2cb6de10c89c3152b196ccb1c11bd930bf6ec82eadadcb">+113/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>lsp_folder_config_notif_test.go</strong><dd><code>Add test for folder config notification matching validators</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1370/changes#diff-5825ab669a345a0ecb15dfa71158d1653dac3b4c02f7e42d9c22250fbd375eae">+60/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>notification_test.go</strong><dd><code>Stabilize progress listener tests for init races</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1370/changes#diff-09b80ca28f0c9f8db60c941c24e92dfc65ded48c06249e70e2004faec30b9891">+5/-4</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>precedence_smoke_test.go</strong><dd><code>Preserve scan payloads when clearing notifications</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1370/changes#diff-1c743b9b799604d498e5a5290961ffc135dec375c44150ad820cba81f3926d39">+4/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>server_smoke_test.go</strong><dd><code>Update smoke tests for async initialization</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1370/changes#diff-13396c212387ae79a286802e0885b1c86f1bc389f0f6e82de58d49d641e15e08">+37/-1</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>server_test.go</strong><dd><code>Add options for scanner override and concurrency</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1370/changes#diff-165bad981a0cc202462a2fea7f236eab9ca3a72bd47fe8bae6b3adad801351fb">+46/-8</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>apply_auth_config_test.go</strong><dd><code>Test handshake ack logic for endpoint change clear</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1370/changes#diff-1a300a09b5bd741eaaf57331b3a671e61e68bdd8f84a83b8d8333d497447e093">+30/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>scanner_test.go</strong><dd><code>Stabilize scan cancellation test</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1370/changes#diff-ca51a6c8a5aee5d9347b6425bca1629ea179a6722ad1bb0eae8fb52fdb490327">+7/-7</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>config_readers_test.go</strong><dd><code>Add tests for handshake acknowledged signal</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1370/changes#diff-88371adbfdb424175de543df2de8e581a18422b7c320e2e4129515f9871da7de">+24/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Enhancement</strong></td><td><details><summary>3 files</summary><table>
<tr>
  <td><strong>server.go</strong><dd><code>Implement background init and shutdown management</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1370/changes#diff-1b5a200b7dc3f37ed67632d0c2682eeb53762813f96edb15cb7660e076fa6031">+136/-25</a></td>

</tr>

<tr>
  <td><strong>apply_auth_config.go</strong><dd><code>Guard endpoint change clear on handshake acknowledgement</code>&nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1370/changes#diff-dcbb2ac73cd481e142a8a55439036be377feb70e23eb5bfd18f59909d7678074">+11/-8</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>config_readers.go</strong><dd><code>Introduce handshake acknowledged signal</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1370/changes#diff-3b315cd032c0bc1cd1052fd4bb3cc0c762129e8d5cd03c1ef15f42cd926c4588">+19/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Miscellaneous</strong></td><td><details><summary>1 files</summary><table>
<tr>
  <td><strong>json_rpc_recorder.go</strong><dd><code>Add method to clear specific notifications</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1370/changes#diff-0bac7761cd858c0179b4c5003f4cafce78ee5a32841e6cd26db5d68b76a32970">+14/-0</a>&nbsp; &nbsp; </td>

</tr>
<tr>
  <td><strong>ldx_sync_config.go</strong><dd><code>Add internal setting for handshake acknowledged</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1370/changes#diff-e42a6f2c9aece134bd052ddbd7e835aefd3661d3d5f9446f7d5b621c421eea4d">+18/-11</a>&nbsp; </td>

</tr>

</details>

___


application/server/background_init_lifecycle_test.go
```diff
@@ -0,0 +1,203 @@
...
+// signal must still fire.
+type panicInitScanner struct {
+	*scanner.TestScanner
+}
+
+func (s *panicInitScanner) Init(_ context.Context) error {
+	panic("simulated panic in background scanner init")
+}
+
+// Test_BackgroundInit_PanicRecovered_ServerSurvives_SignalFires asserts SHOULD-FIX 1.
```
rrama: Nit: AI left in comment.

application/server/background_init_lifecycle_test.go
```diff
@@ -0,0 +1,203 @@
...
+	return s.initReturned
+}
+
+func (s *lifecycleScanner) getInitCtx() context.Context {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	return s.initCtx
+}
+
+// Test_Shutdown_CancelsAndAwaitsBackgroundInit asserts SHOULD-FIX 2: shutdown must
```
rrama: Nit: AI left in comment.

application/server/background_init_lifecycle_test.go
```diff
@@ -0,0 +1,203 @@
...
+
+	// Cancel: the context handed to Init must be a cancellable one that shutdown
+	// cancels — not context.Background().
+	initCtx := sc.getInitCtx()
+	require.NotNil(t, initCtx)
+	require.Error(t, initCtx.Err(),
+		"the background init must run on a cancellable context that shutdown cancels, not context.Background()")
+}
+
+// panicInitScanner panics inside Init to exercise SHOULD-FIX 1: a panic in the
```
rrama: Nit: AI left in comment.

application/server/dispatch_starvation_test.go
```diff
@@ -0,0 +1,185 @@
...
+// When the stored OAuth token is expired/revoked, the scanner's Init path enters
+// a failing token-refresh storm. Historically initializedHandler ran that Init
+// synchronously, so it held a jrpc2 dispatch worker for the whole storm (~60s in
+// the field). With a saturated worker pool the later workspace/executeCommand that
+// asks for the settings-configuration HTML was never dispatched until Init unblocked
+// — even though rendering that HTML needs no auth and is fast.
+//
+// These tests inject a scanner whose Init blocks (a faithful stand-in for the
+// failing-refresh storm) and pin the server worker pool to 1 so the starvation is
+// deterministic regardless of host core count. They are RED before CP1 (config HTML
```
rrama: Nit: AI left in comment.

application/server/server.go
```diff
@@ -824,11 +840,26 @@ func getDownloadURL(conf configuration.Configuration, engine workflow.Engine, pr
...
 	return handler.New(func(ctx context.Context, params types.InitializedParams) (any, error) {
-		initialLogger := ctx2.LoggerFromContext(ctx)
+		// Readiness-signal safety net. The normal signal fires at the END of the
+		// background scanner-init goroutine launched below (invariant D1: scanner-readiness
+		// waiters must wait for real init completion). But every dependency resolution
+		// before that launch (mustConfigResolverFromContext, …, mustScannerFromContext,
+		// etc.) panics on a missing dep; jrpc2/withContext recover such a panic and keep
+		// the server alive, but the goroutine would never start, stranding any goroutine
+		// parked in WaitForLspInitialized (e.g. outbound-notification dispatch) forever.
+		// Registering this defer before the resolutions restores the pre-CP1 "signal
```
rrama: Nit: AI left in comment.

application/server/server.go
```diff
@@ -859,41 +890,106 @@ func initializedHandler(conf configuration.Configuration, engine workflow.Engine
...
+		backgroundInitCancel = initCancel
+		done := make(chan struct{})
+		backgroundInitDone = done
+		go func() {
+			// close(done) is registered first so it runs LAST: shutdown, which waits on
+			// this channel, only observes completion after the readiness signal below has
+			// fired.
+			defer close(done)
+			// Recover any panic from the init path (e.g. a nil-deref on the failing-token
+			// refresh cold path). Pre-CP1 this work ran inside the handler call stack, which
```
rrama: Nit: AI left in comment.

application/server/background_init_lifecycle_test.go
```diff
@@ -0,0 +1,203 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package server
+
+// IDE-2181 — background-init goroutine lifecycle.
+//
+// CP1 moved the blocking scanner Init + folder handling + first scan out of the
```
rrama: Nit: AI left in comment.

application/server/server.go
```diff
@@ -72,6 +72,22 @@ import (
...
 var cacheCheckCancel context.CancelFunc
 
+// Background scanner-init lifecycle (IDE-2181). initializedHandler launches the heavy
+// scanner init in a background goroutine; shutdownHandler must cancel and await it
+// before disposing the notifier/tree-emitter/timers it uses. These are written
+// synchronously in the initialized handler (mirroring cacheCheckCancel) and read by
+// shutdown, which runs later in the LSP lifecycle.
+var (
+	backgroundInitCancel context.CancelFunc
+	backgroundInitDone   chan struct{}
```
rrama, line 82: I am fine to keep these package level for now, since `cacheCheckCancel` already is, so it isn't adding too much tech debt.

application/server/server.go
```diff
@@ -859,41 +890,106 @@ func initializedHandler(conf configuration.Configuration, engine workflow.Engine
...
+			// unrecovered panic here would terminate the whole language server. Recovering it
+			// (and still firing the readiness signal) preserves the pre-CP1 behavior and the
+			// "signal always fires" guarantee so WaitForLspInitialized waiters never hang.
+			defer func() {
+				if r := recover(); r != nil {
+					logger.Error().Str("stack", string(debug.Stack())).Msgf("snyk-ls: recovered panic in background initialization: %v", r)
+				}
+				conf.Set(types.SettingIsLspInitialized, true)
+				types.SignalLspInitialized(conf)
+			}()
```
rrama, line 979: I don't like us continuing init as if nothing is bad when some of these stages fail (e.g. handling the folders).  But some are fine to just recover, e.g. first scan.

I had AI write it up a bit better:

> This single recover() wraps the whole goroutine body, so it treats a panic in Init, HandleFolders, and ScanWorkspace identically — recover, log, fire the readiness signal, keep running. I don't think a blanket "recover and continue" is right for all three stages:
> 
> Init panic (server.go:962): A returned error here is expected (bad/expired token) and handled gracefully — that's the whole point of this PR. But a panic is unexpected. At minimum we should warn the user the scanner is non-functional; stopping the server may be safer. Right now the only trace is a log line.
> HandleFolders panic (server.go:972): This is the one I'm most concerned about. It mutates persisted per-folder state; a panic mid-way leaves that half-written, and we then fire SettingIsLspInitialized = true on top of corrupted state. Recovering and continuing here risks propagating corruption. I'd lean toward stopping.
> ScanWorkspace panic (server.go:977): Fine to recover — it's the last step, non-corrupting, and re-triggerable.
> Could we split this into per-stage recovery instead of one blanket recover? Two things to watch if we do stop the server on Init/HandleFolders panics: (1) a bare re-panic in this goroutine crashes the process (unrecovered goroutine panic) rather than stopping cleanly, so we'd need an explicit clean-stop path; and (2) that path still has to release any WaitForLspInitialized waiters, otherwise shutdownHandler (which waits on backgroundInitDone at server.go:1174-1180) could itself block.

internal/types/ldx_sync_config.go
```diff
@@ -124,17 +124,24 @@ const (
...
+	SettingCachedOriginalPath      = "cached_original_path"
+	SettingUserSettingsPath        = "user_settings_path"
+	SettingIsLspInitialized        = "is_lsp_initialized"
+	// SettingIsLspHandshakeAcknowledged is set synchronously at the end of the
+	// initialized handler, BEFORE the background scanner init runs. It marks that the
+	// LSP initialize/initialized handshake has been acknowledged, so a subsequent
+	// config change is a real client action rather than part of startup. This is
+	// distinct from SettingIsLspInitialized, which flips only once background scanner
+	// init completes (scanner readiness). See IsLspHandshakeAcknowledged.
+	SettingIsLspHandshakeAcknowledged = "is_lsp_handshake_acknowledged"
```
rrama, line 144: Nit: Name is not self explanatory.

rrama (commented):
The recovery on any panics and just then blindly saying init is done worries me...
There is also quite a bit of AI left-in work commentary left in the comments.

I have not reviewed the tests yet, will do so after lunch.

application/server/background_init_lifecycle_test.go
basti-snyk: Cleaned up in e5cd1f2 — removed the dev-process / AI-authoring narration (CP1/SHOULD-FIX/RED-GREEN/invariant-D1 labels) from this comment while keeping the technical explanation.

application/server/background_init_lifecycle_test.go
basti-snyk: Cleaned up in e5cd1f2 — removed the dev-process / AI-authoring narration (CP1/SHOULD-FIX/RED-GREEN/invariant-D1 labels) from this comment while keeping the technical explanation.

application/server/background_init_lifecycle_test.go
basti-snyk: Cleaned up in e5cd1f2 — removed the dev-process / AI-authoring narration (CP1/SHOULD-FIX/RED-GREEN/invariant-D1 labels) from this comment while keeping the technical explanation.

application/server/background_init_lifecycle_test.go
basti-snyk: Cleaned up in e5cd1f2 — removed the dev-process / AI-authoring narration (CP1/SHOULD-FIX/RED-GREEN/invariant-D1 labels) from this comment while keeping the technical explanation.

application/server/dispatch_starvation_test.go
basti-snyk: Cleaned up in e5cd1f2 — removed the dev-process / AI-authoring narration (CP1/SHOULD-FIX/RED-GREEN/invariant-D1 labels) from this comment while keeping the technical explanation.

application/server/server.go
basti-snyk: Cleaned up in e5cd1f2 — removed the dev-process / AI-authoring narration (CP1/SHOULD-FIX/RED-GREEN/invariant-D1 labels) from this comment while keeping the technical explanation.

application/server/server.go
basti-snyk: Cleaned up in e5cd1f2 — removed the dev-process / AI-authoring narration (CP1/SHOULD-FIX/RED-GREEN/invariant-D1 labels) from this comment while keeping the technical explanation.

application/server/server.go
basti-snyk: Agreed — keeping backgroundInitCancel/backgroundInitDone package-level to mirror cacheCheckCancel. Thanks.

internal/types/ldx_sync_config.go
basti-snyk: Fair point. I've left `SettingIsLspHandshakeAcknowledged` as-is in this pass since I don't have a clearly-better name to hand and it's an exported const with refs across a few files. What would you like it to convey more clearly (e.g. that it marks the initialize/initialized handshake completing, distinct from scanner-readiness)? Happy to rename in a follow-up once we agree on the name.

application/server/server.go
basti-snyk: Good catch, and I agree the blanket recover-and-continue isn't right for all three stages. This is a behavioural change with real safety trade-offs (a self-terminating server on a HandleFolders panic vs. continuing), so I'd rather not fold it into this pass unattended. Capturing the plan for a dedicated follow-up: split into per-stage recovery — recover+continue for ScanWorkspace (last step, non-corrupting, re-triggerable); for Init and HandleFolders, stop cleanly rather than fire SettingIsLspInitialized on top of possibly half-written per-folder state. As you note, that clean-stop path must (1) avoid a bare re-panic that crashes the process as an unrecovered goroutine panic, and (2) still release WaitForLspInitialized waiters and close backgroundInitDone so shutdownHandler doesn't block. I'll open a follow-up ticket for this rather than rush it here — does deferring it to its own PR work for you?

rrama:
AI found bug:

> **Init-window config changes no longer fire diagnostics/analytics side-effects**
> 
> Moving scanner init into the background delays `SettingIsLspInitialized` from being set synchronously at the end of `initialized` to the end of the (possibly ~60s) background goroutine. Every config-change side-effect in `application/server/configuration.go` that is gated on `conf.GetBool(types.SettingIsLspInitialized)` (e.g. `applySeverityFilter:759`, `applyRiskScoreThreshold:818`, `applyIssueViewOptions:844`, `applyDeltaFindings:864`, `applyProductEnablement:695`, and the endpoint/auth-method analytics at `applyApiEndpoints:649` / `applyAuthenticationMethod:672`) will now evaluate false during the newly-widened init window. Previously such changes couldn't race this window (either init was fast so the flag was already true, or dispatch was starved so the change never ran). The PR intentionally migrated only the endpoint-switch credential clear (`domain/ide/command/apply_auth_config.go`) to the early handshake-ack signal, deliberately leaving these non-safety side-effects gated on scanner readiness. Impact is minor and self-heals: with auto-scan enabled the background init's final `ScanWorkspace` re-renders diagnostics reflecting the new settings; with auto-scan disabled there are no cached diagnostics to re-filter. Worth a reviewer confirming this scope (analytics for a real user config change during startup would be dropped, and `sendDiagnosticsForNewSettings` skipped) is acceptable.

Probably fine to do as a follow-up PR. Or maybe we just don't care as the gap is tiny.

application/server/background_init_lifecycle_test.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+// panicInitScanner panics inside Init to verify that a panic in the background init
+// goroutine is recovered, the server survives, and the readiness signal still fires.
+type panicInitScanner struct {
+	*scanner.TestScanner
+}
+
+func (s *panicInitScanner) Init(_ context.Context) error {
+	panic("simulated panic in background scanner init")
```
rrama, line 161: Small testing improvement: Verify the panic line is called in the test by saving a bool in the struct, setting it true here and confirming it at the end of the test.

application/server/server_test.go
```diff
@@ -115,6 +117,26 @@ func WithCallback(fn onCallbackFn) ServerTestOption {
 	}
 }
 
+// WithScanner replaces the Scanner dependency injected into request contexts.
+// It composes with WithRealDI, so a test can keep the real command service (needed
+// for commands that actually render, e.g. the settings-configuration HTML) while
+// still controlling scanner Init behavior.
+func WithScanner(s scanner.Scanner) ServerTestOption {
```
rrama, line 128: Hmm, I worry this will get out of hand and it may be better to make `WithDeps` work with `WithRealDI`.

rrama (approved):
I think this is better, but as said previously I worry about the recovery in all cases. I think we should either fix that now or do a follow up PR ASAP to only recover in some cases and shutdown in others.

application/server/server.go
basti-snyk: Filed the follow-up as IDE-2290 (https://snyksec.atlassian.net/browse/IDE-2290) — per-stage recovery redesign will be handled there in its own PR. — AI

application/server/background_init_lifecycle_test.go
basti-snyk: Done in 9cc57e06 — added a mutex-guarded `panicked` flag to `panicInitScanner` (set true immediately before the panic), and the test now asserts `getPanicked()` at the end, so the injected panic path is positively verified rather than inferred from survival + signal alone. Verified `-race` clean. — AI

internal/types/ldx_sync_config.go
basti-snyk: Tracking the rename as a follow-up: **IDE-2290-adjacent** — filed **IDE-2293** (Relates to IDE-2181). Keeping `SettingIsLspHandshakeAcknowledged` as-is here (exported const, cross-file refs) with the clarifying doc comment added in this PR; we can agree a clearer name and do the rename there. — AI

application/server/server_test.go
basti-snyk: Agreed the option surface shouldn't sprawl. Filed **IDE-2292** (Relates to IDE-2181) to make `WithDeps` compose on top of `WithRealDI` (selective override) instead of adding more one-off options — deferring the harness refactor to its own PR to keep this one focused. — AI

PR #1371, opened 2026-07-09
fix: replace fallback-page re-auth control with a Log out action [IDE-2181]

basti-snyk:
## Summary
The limited settings **fallback** page (shown when the full settings page can't load) had no login control — its Authentication section held only the Insecure/SSL-skip checkbox. A user dropped to the fallback was locked out of re-auth entirely, which is the user-facing half of IDE-2181.

This adds a **"Connect"** control to the fallback page that dispatches `snyk.login` through the IDE command bridge with **no args**, so re-auth reuses the user's existing configured method/endpoint (the limited page must not overwrite them). It is no-op-safe when the bridge is absent, disables while a login is in flight (no double-dispatch), and always recovers — via the bridge callback, a synchronous error, or a safety timeout — so it can never stick on "Connecting…". A generation latch makes a late callback from a superseded attempt inert. Status copy is state-neutral.

## PR Stack — Merge Order
```mermaid
flowchart LR
    main(["main"])
    PR1["#1370 CP1\nLS dispatch fix"]
    PR2["#1371 CP2 ← YOU ARE HERE\nfallback re-auth control"]
    main --> PR1 --> PR2
    style PR2 fill:#ffd700,color:#000
```

**Depends on:** #1370

## Verified (outside-in TDD)
`js-tests/settings-fallback.test.mjs` — asserts the control dispatches `snyk.login` (spy, not DOM-only), the disable/re-enable lifecycle, second-click suppression, the safety-timeout recovery, the cross-cycle stale-callback latch, and state-neutral copy. `make test` GREEN (146 JS tests pass).

## Deferred to later PR (CP3)
- `snyk-intellij-plugin`: mirror this fallback page (the plugin ships its own byte-identical copy) and prove `snyk.login` dispatches from it.

## Test plan
- [ ] Start the IDE with an expired token so the settings page falls back; click **Connect** → the OAuth flow starts and re-auth completes.
- [ ] Cancel the OAuth tab → the button recovers (not stuck on "Connecting…") and can be retried.

___

### **PR Type**
Enhancement, Tests

___

### **Description**
- Add 'Log out' button to settings fallback page.

- Dispatch `snyk.logout` command via IDE bridge.

- Implement and test new logout functionality.

___

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Tests</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>authentication_flows_e2e_test.go</strong><dd><code>Add E2E test for logout during scanner initialization</code>&nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

application/server/authentication_flows_e2e_test.go

<ul><li>Adds a new end-to-end test <br><code>Test_E2E_LogoutDuringScannerInitClearsCredentials</code>.<br> <li> This test verifies that the <code>snyk.logout</code> command correctly clears <br>credentials, even when the language server is in the midst of its <br>initialization phase.</ul>

</details>

  </td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1371/changes#diff-9deadd440a71de07926bf9936df138c9c5512696795d6b6544d4f40c01d06752">+65/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>settings-fallback.test.mjs</strong><dd><code>Unit tests for fallback page logout control</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

js-tests/settings-fallback.test.mjs

<ul><li>Removes unit tests related to the old 'Connect' (re-authentication) <br>button.<br> <li> Adds new unit tests for the 'Log out' button, covering its presence, <br>correct command dispatch (<code>snyk.logout</code>), no-op safety, and status <br>update behavior.</ul>

</details>

  </td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1371/changes#diff-6da8cfbb056402eea87adbb3081beda3776b81b201de3374c4d20a61b5362931">+78/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></td></tr><tr><td><strong>Enhancement</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>settings-fallback.html</strong><dd><code>Update fallback page UI for logout control</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

shared_ide_resources/ui/html/settings-fallback.html

<ul><li>Replaces the previous 'Connect' button for re-authentication with a <br>'Log out' button.<br> <li> Adds new CSS styles for the logout button and status messages.<br> <li> Implements the <code>startLogout</code> JavaScript function to dispatch the <br><code>snyk.logout</code> command and update the UI status.<br> <li> Updates the HTML structure and event listeners to support the new <br>logout functionality.</ul>

</details>

  </td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1371/changes#diff-3489a54a68d7b46f5b60db397b73c9f4e37526c68f41c78cee2b2e1c7da66eb2">+55/-0</a>&nbsp; &nbsp; </td>

</tr>

</details>

___

### DIFF
--- a/shared_ide_resources/ui/html/settings-fallback.html
+++ b/shared_ide_resources/ui/html/settings-fallback.html
@@ -59,6 +59,10 @@
       --border-color: var(--vscode-panel-border, #454545);
       --focus-border: var(--vscode-focusBorder, #007acc);
 
+      /* Buttons */
+      --button-background: var(--vscode-button-background, #0e639c);
+      --button-foreground: var(--vscode-button-foreground, #ffffff);
+
       /* Other */
       --border-radius: 4px;
       --border-radius-small: 2px;
@@ -213,6 +217,34 @@
       display: none !important;
     }
 
+    /* Re-authentication control */
+    .reauth-status {
+      margin: 0 0 var(--form-group-margin) 0;
+      color: var(--text-color);
+    }
+
+    .reauth-button {
+      align-self: flex-start;
+      padding: var(--input-padding) 16px;
+      background-color: var(--button-background);
+      color: var(--button-foreground);
+      border: 1px solid var(--button-background);
+      border-radius: var(--border-radius-small);
+      font-family: inherit;
+      font-size: inherit;
+      cursor: pointer;
+    }
+
+    .reauth-button:focus {
+      outline: 1px solid var(--focus-border);
+      outline-offset: 2px;
+    }
+
+    .reauth-button:disabled {
+      opacity: 0.6;
+      cursor: default;
+    }
+
     /* Info Message */
     .info-message {
       padding: var(--section-padding);
@@ -310,6 +342,11 @@
   <div class="section">
     <h2>Authentication</h2>
 
+    <div class="form-group">
+      <p id="reauth-status" class="reauth-status">Sign in or reconnect with Snyk.</p>
+      <button type="button" id="reauth-button" class="reauth-button">Connect</button>
+    </div>
+
     <div class="form-group half-width">
       <label class="checkbox-label">
         <input type="checkbox" name="proxy_insecure" id="proxy_insecure" {{INSECURE_CHECKED}}>
@@ -366,6 +403,9 @@
     var initialState = null;
     // Guard against re-entrance when blur() triggers getAndSaveIdeConfig
     var _isSaving = false;
+    // Monotonic re-auth cycle counter. Each startReauth() bumps it; a done()
+    // callback from a superseded cycle is inert (see startReauth).
+    var _reauthGen = 0;
 
     function get(id) {
       return document.getElementById(id);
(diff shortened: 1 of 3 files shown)


rrama:
We do not want all of this, we just want a "Log out" button or maybe "Clear auth" or something, just to get them out of the bad state or to allow them to clear credentials without a working LS. We do not need an ability for them to login without an LS.

rrama:
One gap worth flagging on the fallback page's logout wiring (`shared_ide_resources/ui/html/settings-fallback.html`, the `startLogout()` function calling `window.__ideExecuteCommand__('snyk.logout', [], done)`):

This dispatches through each IDE's generic command bridge straight to the LS `snyk.logout` command. `logoutCommand.Execute` (`domain/ide/command/logout.go`) clears LS-side state synchronously — confirmed by this PR's own E2E test. But the mechanism that clears the **IDE's own local credential store** is the `$/snyk.hasAuthenticated` push-back (`AuthenticationServiceImpl.updateCredentials` → `notifier.Send(AuthenticationParams{Token: "", ...})`), and that notification is gated behind `WaitForLspInitialized` in `registerNotifier` (`application/server/notification.go:92-97`) — i.e. it can queue behind the very init state this fallback page exists to route around. If init is genuinely stuck (not just slow), that push may never fire, so the IDE-local token can remain stale even though LS's copy is cleared.

Checked all four IDE clients for how their *native* command-palette logout compares:

- **VS Code**: native `snyk.initiateLogout` → `authenticationService.initiateLogout()` clears local token/context eagerly *before* calling LS — doesn't depend on the push. The fallback page currently dispatches `'snyk.logout'` directly instead of `'snyk.initiateLogout'`, so it misses that eager clear.
- **IntelliJ**: native `SnykLogoutAction` → `SnykCliAuthenticationService.logout()` clears `pluginSettings().token` locally first, same shape. The fallback bridge (`ExecuteCommandBridge.kt`) forwards any command generically to the LS with no path to that local service, so it also misses the eager clear.
- **Eclipse** / **Visual Studio**: neither has an eager local clear anywhere today, native path included — both rely solely on the `hasAuthenticated` push to clear `SecurePreferenceStore` / `Options.ApiToken` respectively. So they carry this same exposure even outside the fallback page, on their existing native logout action. That's a pre-existing gap, not something introduced here, but seems worth its own follow-up ticket.

For this PR's scope, worth considering whether the fallback page should route through each IDE's real local-clear path instead of relying entirely on the LS round-trip — e.g. VS Code swapping the dispatched command to `snyk.initiateLogout`, IntelliJ special-casing `snyk.logout` in `ExecuteCommandBridge.dispatch` to call `SnykCliAuthenticationService.logout()` directly. Alternatively/additionally, a dedicated `window.__ideLogout__` hook (falling back to today's `snyk.logout` dispatch when absent) would give an explicit contract instead of relying on IDE-specific command-string knowledge baked into shared HTML — happy to leave that choice to whoever picks up the IDE-side work.

shared_ide_resources/ui/html/settings-fallback.html
```diff
@@ -213,6 +217,29 @@
...
+      align-self: flex-start;
+      padding: var(--input-padding) 16px;
+      background-color: var(--button-background);
+      color: var(--button-foreground);
+      border: 1px solid var(--button-background);
+      border-radius: var(--border-radius-small);
+      font-family: inherit;
+      font-size: inherit;
+      cursor: pointer;
+    }
```
rrama, line 236: Consider matching the real page's CSS for log out. E.g. grey instead of blue, as this isn't an action we want the user taking often.

Unhovered:
Image
Hovered:
Image
(The colour change is move obvious when you hover...)

shared_ide_resources/ui/html/settings-fallback.html
```diff
@@ -310,6 +337,11 @@ <h1>Snyk Settings</h1>
   <div class="section">
     <h2>Authentication</h2>
 
+    <div class="form-group">
+      <p id="logout-status" class="logout-status">If Snyk is stuck, clear your stored credentials to recover, then reopen Settings to sign in again.</p>
```
rrama, line 341: I will defer the wording choice review to Matt. IMO seems a bit wordy and doesn't look the nicest. I also wonder if a "Clear Credentials" label is a better option. And maybe the text should be alongside the button?

Currently:
Image

rrama (approved):
We will re-consider doing a local clear approach if we hit situations that this can't clear. But until then, the button is better than no button.

PR #1377, opened 2026-07-16
fix: expired OAuth token still shows user as authed [IDE-2178]

nick-y-snyk:
## Summary
- Fix singleflight self-deadlock: the OAuth refresher's reentrant `IsAuthenticated()` call for the same in-flight token blocked forever on `authCheckGroup.Do`, so a failed refresh never completed and logout never fired.
- Fix `shouldCauseLogout` misclassifying real permanent 401s as transient: `http.Client.Do` always wraps RoundTripper errors in `*url.Error`, so `isTransientNetworkError` matched unconditionally before string-based permanent-error checks ran. Now checks GAF's structured `snyk_errors.Error.StatusCode` first.

Root-caused via manual repro against a real CLI build with an expired/revoked OAuth token (balloon showed "Could not retrieve authentication status... Authentication error" with no auto-logout).

## Test plan
- [x] `go test ./infrastructure/authentication/...` — 158 pass
- [x] `make lint-fix` — 0 issues
- [ ] CI full suite

### DIFF
--- a/infrastructure/authentication/auth_configuration.go
+++ b/infrastructure/authentication/auth_configuration.go
@@ -81,9 +81,6 @@ func Default(engine workflow.Engine, authenticationService AuthenticationService
 		refreshToken, err := auth.RefreshToken(ctx, oauthConfig, token)
 		if err != nil {
 			logger.Err(err).Msg("failed to refresh oauth2 token")
-			// call authservice to handle notifications and such
-			// we don't need the returned values, as we know it will either return false, nil or false, err
-			_ = authenticationService.IsAuthenticated()
 		}
 		return refreshToken, err
 	}
--- a/infrastructure/authentication/auth_configuration.go
+++ b/infrastructure/authentication/auth_configuration.go
@@ -73,7 +73,13 @@ func Default(engine workflow.Engine, authenticationService AuthenticationService
 		types.DefaultOpenBrowserFunc(url)
 	}
 
-	// this doesn't have any effect
+	// This closure is only wired into this OAuth2Provider's own authenticator, used by the
+	// explicit login flow (Authenticate -> CancelableAuthenticate). Real API traffic (scans,
+	// whoami) goes through a separate authenticator GAF builds internally for the network
+	// client, which never receives this refresherFunc - so it has no effect on that path.
+	// It used to reentrantly call authenticationService.IsAuthenticated() on failure, which
+	// would deadlock the singleflight group in IsAuthenticated() if this closure were ever
+	// invoked from within it; that call was removed for IDE-2178.
 	refresherFunc := func(ctx context.Context, oauthConfig *oauth2.Config, token *oauth2.Token) (*oauth2.Token, error) {
 		logger := engine.GetLogger().With().Str("method", "oauth.refresherFunc").Logger()
 		logger.Info().Msg("refreshing oauth2 token")
--- a/infrastructure/authentication/auth_service_impl.go
+++ b/infrastructure/authentication/auth_service_impl.go
@@ -88,12 +88,6 @@ type AuthenticationServiceImpl struct {
 	// authCheckGroup coalesces concurrent auth API calls so only one in-flight request
 	// is made at a time; all waiters share the same result.
 	authCheckGroup singleflight.Group
-	// inFlightAuthChecks tracks tokens currently inside authCheckGroup.Do. A reentrant
-	// IsAuthenticated() call for the same token (e.g. the OAuth refresher invoking it again
-	// after a failed refresh, while the original check for that token is still running)
-	// would block forever on singleflight.Group, since the "in-flight" call is the very
-	// goroutine waiting on it. Detect that case here and short-circuit instead of deadlocking.
-	inFlightAuthChecks sync.Map
 	// credentialUpdateChan serializes credential updates from the OAuth storage bridge
 	// to prevent race conditions where older tokens overwrite newer ones during rapid rotations.
 	credentialUpdateChan chan credentialUpdate
@@ -715,12 +709,6 @@ func (a *AuthenticationServiceImpl) doAuthCheck(conf configuration.Configuration
 
 	// Coalesce concurrent auth API calls: all in-flight callers share one result.
 	token := config.GetToken(conf)
-	if _, alreadyInFlight := a.inFlightAuthChecks.LoadOrStore(token, struct{}{}); alreadyInFlight {
-		logger.Debug().Msg("auth check already in flight for this token, skipping reentrant call")
-		return false
-	}
-	defer a.inFlightAuthChecks.Delete(token)
-
 	v, _, _ := a.authCheckGroup.Do(token, func() (interface{}, error) {
 		u, e := a.authProvider.GetCheckAuthenticationFunction()(a.engine)
 		return &authCheckResult{user: u, err: e}, nil
--- a/infrastructure/authentication/auth_service_impl.go
+++ b/infrastructure/authentication/auth_service_impl.go
@@ -70,7 +70,10 @@ type AuthenticationServiceImpl struct {
 	// key = token, value = isAuthenticated
 	authCache *imcache.Cache[string, bool]
 	// Last token that was successfully used for authentication. It might have expired (so not be present in authCache).
+	// Uses its own mutex (not m) because doAuthCheck runs under m.RLock, which allows
+	// concurrent writers here.
 	lastUsedToken               string
+	lastUsedTokenMu             sync.Mutex
 	m                           sync.RWMutex
 	previousAuthCtxCancelFunc   context.CancelFunc
 	previousAuthCtxCancelFuncMu sync.Mutex
@@ -745,7 +748,7 @@ func (a *AuthenticationServiceImpl) doAuthCheck(conf configuration.Configuration
 			return false
 		}
 
-		invalidOAuth2Token, isLegacyTokenErr := config.ParseOAuthToken(config.GetToken(conf), a.engine.GetLogger())
+		invalidOAuth2Token, isLegacyTokenErr := config.ParseOAuthToken(token, a.engine.GetLogger())
 		isLegacyToken := isLegacyTokenErr != nil
 
 		a.handleEmptyUser(logger, isLegacyToken, invalidOAuth2Token)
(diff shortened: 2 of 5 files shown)


infrastructure/authentication/auth_service_impl.go
```diff
@@ -707,6 +715,12 @@ func (a *AuthenticationServiceImpl) doAuthCheck(conf configuration.Configuration
 
 	// Coalesce concurrent auth API calls: all in-flight callers share one result.
 	token := config.GetToken(conf)
+	if _, alreadyInFlight := a.inFlightAuthChecks.LoadOrStore(token, struct{}{}); alreadyInFlight {
+		logger.Debug().Msg("auth check already in flight for this token, skipping reentrant call")
+		return false
```
rrama: AI review comment:

> **Simultaneous sign-in checks can wrongly report an authenticated user as signed out**
> 
> Every simultaneous authentication check except the first one for the same token is immediately answered with "not authenticated" (`LoadOrStore` at `infrastructure/authentication/auth_service_impl.go:718`) instead of waiting to share the successful result, so an authenticated user is momentarily reported as signed out.
> Impact: During the window while the very first check is still contacting the API (e.g. at startup or after the 1-minute cache expires), concurrent callers such as folder scans see a false negative and silently skip scanning even though the user is signed in.
> 
> <details>
> <summary>Token-keyed in-flight map cannot distinguish reentrancy from concurrency</summary>
> 
> The new guard was added to fix a genuine same-goroutine reentrant deadlock (the OAuth refresher calling `IsAuthenticated()` from inside `authCheckGroup.Do`). However it is keyed only by the token string in a `sync.Map`, so it cannot tell apart a same-goroutine reentrant call from a legitimate different-goroutine concurrent call.
> 
> `IsAuthenticated()` takes only `a.m.RLock()` (`infrastructure/authentication/auth_service_impl.go:682`), which allows multiple concurrent readers. Previously `singleflight.Group.Do` (`infrastructure/authentication/auth_service_impl.go:724`) coalesced concurrent callers for the same token: the first executed the check and all other waiters blocked and then received the *same* `authCheckResult`, so all of them returned `true` for a valid token — exactly what the comment at `auth_service_impl.go:716` still documents ("all in-flight callers share one result").
> 
> With the new `LoadOrStore` short-circuit, only the first goroutine runs the check; every other concurrent goroutine sees `alreadyInFlight == true` and returns `false` without waiting for or sharing the result. For an authenticated but not-yet-cached token this produces a spurious `false`.
> 
> A concrete downstream effect: `domain/snyk/scanner/scanner.go:264-269` returns early ("Not authenticated, not scanning") when `IsAuthenticated()` returns false, so concurrent folder scans triggered while the token is uncached can be silently skipped.
> 
> The reentrancy deadlock should be detected in a way that only short-circuits the same goroutine (e.g. tracking the reentrant call per-goroutine / within the Do closure) rather than short-circuiting all concurrent callers.
> 
> </details>

infrastructure/authentication/auth_service_impl.go
```diff
@@ -827,6 +841,14 @@ func shouldCauseLogout(err error, logger *zerolog.Logger) bool {
 		return true
 	}
 
+	// A structured Snyk API error with a definitive auth status code is a permanent failure,
+	// even though http.Client.Do always wraps it in a *url.Error (which isTransientNetworkError
+	// would otherwise treat as transient network noise).
+	var snykErr snyk_errors.Error
+	if errors.As(err, &snykErr) && (snykErr.StatusCode == http.StatusUnauthorized || snykErr.StatusCode == http.StatusBadRequest) {
```
rrama, line 852: AI comment:

> **snyk_errors.Error matched as value type relies on GAF wrapping values not pointers**
> 
> The new permanent-error check uses `var snykErr snyk_errors.Error; errors.As(err, &snykErr)` (`infrastructure/authentication/auth_service_impl.go:847-848`). This only matches when the error chain contains a `snyk_errors.Error` *value*, not a `*snyk_errors.Error` pointer. This pattern (value target) is used consistently elsewhere in the repo (`infrastructure/secrets/errors.go:44`, `infrastructure/cli/cli_extension_executor.go:143`, `domain/ide/workspace/analytics_errors.go:70`), which strongly implies GAF returns the error as a value, so the fix should match production errors. Worth a quick confirmation against a real repro that the whoami workflow chain actually carries the value form (the added test constructs the value form directly rather than exercising the real GAF path).

infrastructure/authentication/auth_service_impl.go
nick-y-snyk: Confirmed the regression — fixed at the root by removing the redundant reentrant `IsAuthenticated()` call in the OAuth refresher (the in-flight leader call already observes the same refresh error and runs the notification/logout path itself), and deleted the `inFlightAuthChecks` guard entirely. Added `TestIsAuthenticated_ConcurrentCallsAllReturnSharedResult` (proves concurrent callers get the correct shared singleflight result, not a hardcoded `false`) and `TestIsAuthenticated_PermanentAuthErrorStillTriggersReAuthNotification` (proves the re-auth notification side effect is preserved without the reentrant call).

infrastructure/authentication/auth_service_impl.go
nick-y-snyk: Confirmed against GAF v0.3.1 source: `errFromStatusCode`/`NewUnauthorisedError`/`NewBadRequestError` return `snyk_errors.Error` by value, and GAF's own `netstack_error_handler.go` uses the identical value-target `errors.As` pattern. The wrap chain (pkg/errors -> fmt.Errorf %w -> errors.Join -> *url.Error) all support Unwrap, so this reaches the value fine. No change needed.

infrastructure/authentication/auth_configuration.go
```diff
@@ -81,9 +81,6 @@ func Default(engine workflow.Engine, authenticationService AuthenticationService
 		refreshToken, err := auth.RefreshToken(ctx, oauthConfig, token)
 		if err != nil {
 			logger.Err(err).Msg("failed to refresh oauth2 token")
-			// call authservice to handle notifications and such
-			// we don't need the returned values, as we know it will either return false, nil or false, err
-			_ = authenticationService.IsAuthenticated()
 		}
 		return refreshToken, err
```
rrama, line 91: AI comment (honestly, I have no idea if it is a valid concern or not, I haven't looked into it):

> **Removed refresher-side auth-check drops notification/logout for non-IsAuthenticated refresh paths**
> 
> The reentrant `authenticationService.IsAuthenticated()` call removed from `refresherFunc` (`infrastructure/authentication/auth_configuration.go:82-84`) previously fired on ANY OAuth refresh failure, including refreshes that GAF triggers during ordinary API calls (e.g. a scan), not just during an explicit `IsAuthenticated()` whoami check. The PR correctly removes it to break the singleflight self-deadlock, and relies on the outer `doAuthCheck()` observing the same failure. However, when a refresh fails during a code path that is NOT an `IsAuthenticated()` call, the immediate notification/logout side effect is now only produced on the next `IsAuthenticated()` poll rather than proactively. For the reported deadlock scenario this is fine (the refresh happens within the whoami check), but the stale `// this doesn't have any effect` comment at `auth_configuration.go:76` is now clearly incorrect (the reentrant call demonstrably DID have effect — it caused the deadlock). Worth confirming that periodic auth checks reliably cover the non-IsAuthenticated refresh-failure case.

infrastructure/authentication/auth_service_impl_test.go
```diff
@@ -380,6 +381,100 @@ func TestIsAuthenticated_ConcurrentCallsSendOnlyOneNotification(t *testing.T) {
...
+	ts.SetToken(engine.GetConfiguration(), "invalidCreds")
+	provider := &errorFakeAuthProvider{err: buildWhoamiErr(fmt.Errorf("API request failed (status: 401)"))}
+	service := NewAuthenticationService(engine, ts, provider, errorReporter, notifier, testutil.DefaultConfigResolver(engine))
+
+	var mu sync.RWMutex
+	messageRequestReceived := false
+	go notifier.CreateListener(func(params any) {
+		if _, ok := params.(types.ShowMessageRequest); ok {
+			mu.Lock()
+			messageRequestReceived = true
```
rrama, line 467: This is a weak check.

AI says:

> This listener only checks that some types.ShowMessageRequest was received — any unrelated one would satisfy it. Since the test's purpose is to prove the re-authenticate request still fires, consider asserting on the action shape like TestHandleInvalidCredentials (auth_service_impl_test.go:649-664) does — i.e. verify the first action's CommandId is types.LoginCommand. That makes it a precise regression guard instead of relying on TestHandleInvalidCredentials for that coverage:
> 
> ```go
> callback := func(params any) {
> 	if p, ok := params.(types.ShowMessageRequest); ok {
> 		keys := p.Actions.Keys()
> 		loginAction, ok := p.Actions.Get(keys[0])
> 		require.True(t, ok)
> 		require.Equal(t, types.LoginCommand, loginAction.CommandId)
> 		mu.Lock()
> 		messageRequestReceived = true
> 		mu.Unlock()
> 	}
> }
> ```

rrama (approved):
Assuming the AI found bug isn't too critical I will approve. But, I do think there is a lack of testing.

AI had this to say:

> The gaps
> 1. No test exercises the actual reentrant path that caused the deadlock. The deleted code was in auth_configuration.go, but neither new test goes through the OAuth refresher (refresherFunc). TestIsAuthenticated_PermanentAuthErrorStillTriggersReAuthNotification (:448-476) uses errorFakeAuthProvider returning a plain error — it does not trigger a real OAuth refresh. So if someone re-adds the reentrant IsAuthenticated() call to the refresher, no test catches the deadlock directly. The tests guard the symptom shape (concurrency + notification), not the deleted-line's actual code path. That's the weakest point of the coverage.
> 
> 2. StatusCode == 400 is untested. shouldCauseLogout treats 400 as permanent (:836), but only the 401 snyk_errors.Error case is tested. There's a string-based (status: 400) test but no structured 400 test.
> 
> 3. The notification-shape assertion is weak — a reviewer already flagged this (unresolved thread at auth_service_impl_test.go:463). The listener only checks some ShowMessageRequest fired, not that it's the re-auth/LoginCommand action. Any unrelated ShowMessageRequest would pass.
> 
> 4. Open design question is untested — the unresolved thread at auth_configuration.go:85 raises that the deleted reentrant call fired on any refresh failure (including non-IsAuthenticated paths like scans), and now those only notify on the next poll. No test covers the "refresh fails outside an IsAuthenticated call" case. Whether that matters is a real behavioral question the tests don't touch.
> 
> Bottom line
> Coverage is adequate to prevent the two bugs from silently regressing at the unit level, but I'd want: (a) a test that drives a real OAuth refresh failure through refresherFunc to guard the deleted line directly, (b) a structured-400 case, and (c) tightening the notification assertion per the reviewer's suggestion. So your instinct is right — it's a bit lacking, mainly around the reentrancy path itself.

infrastructure/authentication/auth_service_impl_test.go
nick-y-snyk: Good catch — tightened the assertion to mirror `TestHandleInvalidCredentials`: now verifies `loginAction.CommandId == types.LoginCommand`, not just that some `ShowMessageRequest` fired.

infrastructure/authentication/auth_configuration.go
nick-y-snyk: Dug into this properly, including checking the actual Jira ticket (IDE-2178) and running an empirical test through the real production entry point — worth the full picture:

**Is the comment stale / is there a real coverage gap?** No, on both counts, but for a different reason than either of us assumed. `Default()'s `refresherFunc` is only ever wired into `OAuth2Provider's own authenticator, used by the explicit login flow (`Authenticate()`/`CancelableAuthenticate()`). Real `IsAuthenticated()`/scan/whoami traffic goes through a completely separate authenticator GAF builds internally (`OAuth2Provider.GetCheckAuthenticationFunction()` returns a free function that ignores its receiver) — confirmed empirically with a new test (`Test_IsAuthenticated_DoesNotUseOAuth2ProviderCustomRefresherFunc`) that drives the real `IsAuthenticated()` path against an httptest server and asserts the closure's distinctive log line never fires, even though a real refresh attempt and whoami call both happen.

**So was the deadlock ever real?** Checked the actual Jira ticket — its root cause is entirely different from singleflight reentrancy: GAF doesn't clear the OAuth token on `invalid_grant`, and our string-matching in `shouldCauseLogout` could miss the marker because GAF passes through a plain 401. That's what the structured-error-detection commit in this PR actually fixes. The singleflight-deadlock fix removed a reentrant call from a closure that's provably unreachable from real traffic today — harmless defense-in-depth (if that closure ever gets wired to real traffic in the future, the reentrant call would be a real landmine), but it wasn't what fixed the reported incident.

Reworded the stale comment above `refresherFunc` to explain this, and added a test pinning down that a refresh failure reached outside `IsAuthenticated()` produces no proactive notification today (matches current behavior, not a regression from this PR — that path was never wired to notifications either way).

PR #1380, opened 2026-07-21
fix: cancel login requests when auth method changed, login button pressed, or logout command sent [IDE-2247]

andrewrobinsonhodges-snyk:
### Description

If a user clicks log in, log out, or changes auth method, any in-flight login requests should be cancelled.

This fix also makes oauth browser login non-blocking; if the user closes the browsers before logging in, the LS can now process other login/logout.config commands.

### Checklist


domain/ide/command/login.go
```diff
@@ -116,6 +116,21 @@ func (cmd *loginCommand) Execute(ctx context.Context) (any, error) {
...
+		if util.IsCancellation(err) {
+			logger.Debug().Str("method", "loginCommand.Execute").Msg("login canceled")
+			return nil, err
+		}
+		// A login timeout (e.g. the user didn't complete the browser flow) is a user-facing, retryable
+		// outcome, not a defect: tell the user, but return no error so the command handler doesn't
+		// report it to Sentry. Scoping this here keeps genuine timeouts elsewhere reportable.
+		if util.IsTimeout(err) {
+			logger.Debug().Str("method", "loginCommand.Execute").Msg("login timed out")
+			cmd.notifier.SendError(err)
```
basti-snyk: Should Fix — the user-facing message here leaks developer jargon, which works against IDE-2247's own goal (the comment just above calls a timeout "a user-facing, retryable outcome, not a defect"). `SendError` renders as `"Snyk encountered an error: %v"`, and because the OAuth provider normalizes the timeout to `fmt.Errorf("oauth authentication timed out: %w", context.DeadlineExceeded)`, the user sees *"Snyk encountered an error: oauth authentication timed out: context deadline exceeded."* Consider surfacing a purpose-built friendly message for the timeout branch (e.g. "Snyk sign-in timed out. Please try again.") instead of the wrapped error string. — AI review

infrastructure/authentication/cli_provider.go
```diff
@@ -214,8 +226,13 @@ func (a *CliAuthenticationProvider) runCLICmd(ctx context.Context, cmd *exec.Cmd
...
 	err := cmd.Run()
-	if err == nil && ctx.Err() != nil {
-		err = ctx.Err()
+	// When the context is done (the IDE cancels the login via $/cancelRequest, or a deadline fires),
+	// exec kills the CLI subprocess and cmd.Run() returns "signal: killed". Return the context error
+	// so callers can detect the cancellation or timeout with errors.Is; the subprocess error is
+	// logged at debug for diagnosability.
+	if ctxErr := ctx.Err(); ctxErr != nil {
+		a.engine.GetLogger().Debug().Err(err).Str("method", "runCLICmd").Msgf("Snyk CLI command aborted (%v)", ctxErr)
+		return ctxErr
```
basti-snyk, line 235: Should Fix — this reorder now returns `ctxErr` **before** the `err != nil` branch runs, so whenever the context is done the real `cmd.Run()` error is discarded (demoted to a Debug log) and a bare cancellation/timeout is returned instead. Failure mode: the CLI auth subprocess fails for a genuine reason (bad config, server 5xx) at the same moment an inbound request deadline expires — the real cause is swallowed. Prefer `ctxErr` only when `err` is nil or is the process-kill signal (`signal: killed`); otherwise return the real `err`. At minimum, log the discarded `err` at Err level rather than Debug. — AI review

infrastructure/authentication/auth_service_impl.go
```diff
@@ -217,29 +218,77 @@ func (a *AuthenticationServiceImpl) provider() AuthenticationProvider {
...
+		a.engine.GetLogger().Debug().Str("method", "Authenticate").Msg("authentication superseded before applying result; discarding")
+		return "", ctxErr
+	}
+
+	return a.finishAuthenticate(provider, token, err)
+}
+
+// finishAuthenticate applies the result of the provider's Authenticate call — caching, endpoint
+// discovery, credential update and provider reconfiguration. The caller must hold a.m.
+func (a *AuthenticationServiceImpl) finishAuthenticate(provider AuthenticationProvider, token string, err error) (string, error) {
```
basti-snyk, line 281: Suggestion (latent, not an active bug) — supersession safety currently rests only on the `ctx.Err()` guard plus an unenforced convention that every provider-mutating caller invokes `CancelOngoingAuth` before touching provider state under `a.m`. `finishAuthenticate` never re-checks that the snapshotted `provider` is still `a.authProvider`. Today all callers honor the rule, so it is safe — but a future direct `ConfigureProviders`/`SetProvider` caller that forgets to cancel would let a stale in-flight login apply a token against the wrong provider, with `ctx.Err()==nil` offering no protection. The `provider` param is already threaded in here, so a one-line structural guard (`if a.authProvider != provider { discard }`) is cheap defense-in-depth for a race that would be painful to debug. — AI review

infrastructure/sentry/sentry_error_reporter.go
```diff
@@ -55,7 +62,15 @@ func (s *GDPRAwareSentryErrorReporter) FlushErrorReporting() {
 }
 
 func (s *GDPRAwareSentryErrorReporter) CaptureError(err error) bool {
-	s.notifier.SendError(err)
+	// A canceled operation is an expected outcome, not a failure: never surface it to the user or
+	// report it to Sentry, regardless of where it originated.
+	if util.IsCancellation(err) {
```
basti-snyk, line 67: Should Fix (consistency) — this guard suppresses `IsCancellation` but not `IsTimeout`, and the CLI provider's error path (`cli_provider.go`) likewise only guards `IsCancellation` before calling `CaptureError`. Net effect: an OAuth login timeout is kept out of Sentry (intercepted in `login.go`/`initializer.go`), but a CLI-provider timeout (`context.DeadlineExceeded`) would fall through to Sentry — contradicting the PR's "uniform classification via shared helpers" thesis. Currently unreachable (snyk-ls sets no deadline on the CLI login ctx), so this is latent, but worth deciding one timeout policy and applying it in both providers. Also note: this guard now suppresses `context.Canceled` from **any** subsystem app-wide (scans, hovers, etc.), not just auth — confirm that broad scope is intended. — AI review

infrastructure/authentication/auth_service_impl.go
```diff
@@ -217,29 +218,77 @@ func (a *AuthenticationServiceImpl) provider() AuthenticationProvider {
...
+	//
+	// Supersession invariant: a caller that supersedes an in-flight login by clearing or reconfiguring
+	// provider state cancels the auth context first (via CancelOngoingAuth), so the ctx.Err() guard
+	// below discards this login's now-stale result. Logout does this directly; ConfigureProviders
+	// relies on its caller ApplyAuthMethodChange to cancel before reconfiguring. This is about
+	// correctness, not deadlock avoidance — releasing a.m above means nothing ever waits on the
+	// interactive provider call while holding a.m.
+	a.m.RLock()
+	provider := a.authProvider
+	a.m.RUnlock()
```
rrama: Slightly better would be:
```go
a.m.Lock()
provider := a.authProvider
if provider == nil {
    // log, RemoveAll — still under lock
    a.m.Unlock()
    return "", err
}
a.m.Unlock()

token, err = provider.Authenticate(ctx)
```

avoids a slight race.

infrastructure/authentication/oauth_provider.go
```diff
@@ -52,8 +53,17 @@ func (p *OAuth2Provider) Authenticate(ctx context.Context) (string, error) {
...
+		// GAF's ErrAuthCanceled is a plain sentinel that does not wrap context.Canceled. Normalize it
+		// to context.Canceled so the shared util.IsCancellation helper classifies a canceled OAuth
+		// login as an expected cancellation across the auth/login stack — debug log, no Sentry report,
+		// no user notification.
+		return "", fmt.Errorf("oauth authentication canceled: %w", context.Canceled)
+	case errors.Is(err, auth.ErrAuthTimedOut):
+		p.logger.Debug().Msg("authentication timed out")
+		// Normalize GAF's ErrAuthTimedOut to context.DeadlineExceeded so util.IsTimeout recognizes it:
+		// the timeout is surfaced to the user (so they can retry) but kept out of Sentry.
+		return "", fmt.Errorf("oauth authentication timed out: %w", context.DeadlineExceeded)
```
rrama, line 66: To consider: Should we make the GAF ErrAuthTimedOut be an extension of context.DeadlineExceeded? Or maybe just make IsTimeout check the GAF ones as well.

I have no strong feelings either way.

infrastructure/authentication/cli_provider.go
rrama: I very much disagree, it is a millisecond scenario and the user already cancelled it and so wouldn't care for the result anyway.

domain/ide/command/apply_auth_config.go
```diff
@@ -64,20 +66,40 @@ func ApplyInsecureSetting(conf gafConfig.Configuration, insecure bool) {
...
 
 	previousMethod := config.GetAuthenticationMethodFromConfig(conf)
+	changed := authMethod != previousMethod
 	logger.Info().
 		Str("old_auth_method", string(previousMethod)).
 		Str("new_auth_method", string(authMethod)).
 		Msg("auth method change requested")
+
+	// When the method actually changes, cancel any login still in flight for the previous method so
+	// a stuck auth is aborted immediately. This is also required for correctness:
```
rrama: AI nit pick (unchecked by me):

> Deadlock rationale is now stale: Authenticate no longer holds a.m during the provider call (auth_service_impl.go:249-262), so ConfigureProviders can't deadlock — drop the deadlock claim; only the supersession invariant justifies the cancel.

domain/ide/command/login.go
```diff
@@ -91,9 +91,9 @@ func (cmd *loginCommand) Execute(ctx context.Context) (any, error) {
 
 	if n == 3 {
 		// Cancel any in-progress auth before reconfiguring providers. ConfigureProviders
-		// (called inside applyAuthConfig) acquires the same mutex as Authenticate. Without
-		// canceling first, a blocking Authenticate holds the mutex and ConfigureProviders
-		// deadlocks — preventing the new login from ever starting.
+		// (called unconditionally inside applyAuthConfig → ApplyAuthMethodChange) acquires the same
+		// mutex as Authenticate. Without canceling first, a blocking Authenticate holds the mutex and
```
rrama: AI nit pick (unchecked by me):

> Same stale deadlock rationale: Authenticate releases a.m before the provider call (auth_service_impl.go:249-262), so ConfigureProviders no longer deadlocks — reword to supersession correctness, not deadlock avoidance.

rrama (approved):
LGTM. Just some AI found bits.
NGL, I didn't review the test code, I just skimmed them and they seemed to be enough cover.

domain/ide/command/login.go
andrewrobinsonhodges-snyk: Addressed — the timeout branch now surfaces a purpose-built warning ('Snyk sign-in timed out. Please try again.') via SendShowMessage instead of the wrapped error string, so no context.DeadlineExceeded jargon reaches the user. Test updated accordingly.

infrastructure/authentication/auth_service_impl.go
andrewrobinsonhodges-snyk: Leaving as latent for now — every provider-mutating caller (Logout, ApplyAuthMethodChange→ConfigureProviders) cancels the auth context first, so the ctx.Err() guard covers all current paths. Agreed the provider-identity guard is cheap defense-in-depth; noting it here for a follow-up if a direct SetProvider/ConfigureProviders caller is ever added without a preceding cancel.

infrastructure/sentry/sentry_error_reporter.go
andrewrobinsonhodges-snyk: This asymmetry is intentional. Cancellation is universally expected, so it's suppressed app-wide here (broad scope is deliberate — 'regardless of where it originated'). Timeout is only expected in the interactive login flow, so it's suppressed narrowly at login.go/initializer.go; keeping it out of CaptureError is what keeps genuine scan/API timeouts reportable to Sentry. Suppressing IsTimeout here would regress that. The CLI-provider timeout path is latent (no deadline is set on the CLI login ctx today).

infrastructure/authentication/oauth_provider.go
andrewrobinsonhodges-snyk: I think the current approach of converting it at the boundary works - there might be scenarios where the GAF timeout is needed for other things, and we don't want to go the other way and expose GAF internals here

domain/ide/command/apply_auth_config.go
andrewrobinsonhodges-snyk: Done — dropped the stale deadlock rationale. Authenticate releases a.m before the interactive provider call, so ConfigureProviders can no longer deadlock behind it; the comment now justifies the cancel by the supersession invariant (aborting the stuck login and discarding its stale result via the ctx.Err() guard).

domain/ide/command/login.go
andrewrobinsonhodges-snyk: Done — reworded to supersession correctness. Authenticate releases a.m before the provider call so there's no deadlock; the unconditional cancel here reflects that a login always supersedes in-flight auth (aborting it and discarding its stale result via the ctx.Err() guard), regardless of whether the config changed.

infrastructure/authentication/auth_service_impl.go
andrewrobinsonhodges-snyk: Done — folded the provider snapshot, nil-check and RemoveAll into a single a.m critical section as suggested, closing the RUnlock→Lock gap. Still releases a.m before the interactive provider.Authenticate call. go test -race on the auth package is green.

PR #1384, opened 2026-07-23
fix: only dedupe secrets issues [IDE-2140]

andrewrobinsonhodges-snyk:
### Description

Secrets issues are classed as "one or more instances of a secret in a file". Other scan types treat each instance as an individual issue.

Our deduplication code was incorrectly collapsing some non-secrets issues down (e.g., credentials picked up by the Code scanner, or repeated strings containing issues). 

This fix only collapses on the issue fingerprint if it is a Secrets issue, else it uses the unique identifiers of the issues to count instances.

I've not checked this in, but I did write a test to verify this using a real customer dataset that was incorrectly being deduplicated, and have confirmed that the fix gives the correct totals.

### Checklist


domain/ide/treeview/tree_builder.go
```diff
@@ -630,16 +630,20 @@ type issueStats struct {
...
+// are collapsed by fingerprint only for products where a shared fingerprint means
+// the same finding (product.CollapsesByFingerprint — currently only Secrets);
+// every other product's issues are counted individually.
 func computeIssueStats(issues []types.Issue) issueStats {
 	seen := make(map[string]bool, len(issues))
 	stats := issueStats{severityCounts: &SeverityCounts{}}
 	for _, issue := range issues {
 		fp := issue.GetFingerprint()
-		if fp != "" && seen[fp] {
+		collapses := fp != "" && issue.GetProduct().CollapsesByFingerprint()
```
basti-snyk, line 642: This does not solve the root cause, I think. It is ok for now, but we should investigate why the fingerprints are double.

basti-snyk (approved):
integration tests are failing. but else lgtm.

PR #1386, opened 2026-07-23
chore: fallback settings clear credentials button styling and label

rrama:
### Description

Align the fallback settings authentication action with the full settings page secondary button style and requested copy.

- Renamed fallback authentication button from `Log out` to `Clear credentials`
- Removed the initial helper text shown above the button
- Updated fallback button styling to use the same secondary (grey) color token and hover token as the full settings page Log out button
- Simplified post-action status feedback to exactly: `Signed out.`
- Updated JS tests for the new label and concise post-action status behavior

### Checklist

### Screenshot

Screenshot 2026-07-23 at 17 07 16

Live view [link](https://raw.githack.com/snyk/snyk-ls/refs/heads/cursor/fallback-clear-credentials-836f/shared_ide_resources/ui/html/settings-fallback.html)

<div><a href="https://cursor.com/agents/bc-1486636c-52fd-435d-87b5-2e199686836f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-1486636c-52fd-435d-87b5-2e199686836f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/shared_ide_resources/ui/html/settings-fallback.html
+++ b/shared_ide_resources/ui/html/settings-fallback.html
@@ -551,7 +551,7 @@
       function done() {
         if (statusEl) {
           statusEl.className = statusEl.className.replace(/\bhidden\b/, '').trim();
-          statusEl.textContent = 'Signed out. Your stored credentials were cleared — reopen Settings to sign in again.';
+          statusEl.textContent = 'Signed out.';
         }
       }
       try {
--- a/shared_ide_resources/ui/html/settings-fallback.html
+++ b/shared_ide_resources/ui/html/settings-fallback.html
@@ -62,6 +62,10 @@
       /* Buttons */
       --button-background: var(--vscode-button-background, #0e639c);
       --button-foreground: var(--vscode-button-foreground, #ffffff);
+      --button-hover-background: var(--vscode-button-hoverBackground, #1177bb);
+      --button-secondary-background: var(--vscode-button-secondaryBackground, #3a3d41);
+      --button-secondary-foreground: var(--vscode-button-secondaryForeground, #ffffff);
+      --button-secondary-hover-background: var(--vscode-button-secondaryHoverBackground, #45494e);
 
       /* Other */
       --border-radius: 4px;
@@ -226,13 +230,18 @@
     .logout-button {
       align-self: flex-start;
       padding: var(--input-padding) 16px;
-      background-color: var(--button-background);
-      color: var(--button-foreground);
-      border: 1px solid var(--button-background);
+      background-color: var(--button-secondary-background);
+      color: var(--button-secondary-foreground);
+      border: none;
       border-radius: var(--border-radius-small);
       font-family: inherit;
       font-size: inherit;
+      font-weight: 600;
       cursor: pointer;
+    }
+
+    .logout-button:hover {
+      background-color: var(--button-secondary-hover-background);
     }
 
     .logout-button:focus {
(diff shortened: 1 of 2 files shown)


shared_ide_resources/ui/html/settings-fallback.html
```diff
@@ -338,8 +347,8 @@ <h1>Snyk Settings</h1>
     <h2>Authentication</h2>
 
     <div class="form-group">
-      <p id="logout-status" class="logout-status">If Snyk is stuck, clear your stored credentials to recover, then reopen Settings to sign in again.</p>
-      <button type="button" id="logout-button" class="logout-button">Log out</button>
+      <p id="logout-status" class="logout-status hidden" aria-live="polite"></p>
```
andrewrobinsonhodges-snyk: I don't think this will do anything - did you test it? AIUI screen readers ignore hidden elements. If we wantt this, it should be in the parent div.

shared_ide_resources/ui/html/settings-fallback.html
```diff
@@ -541,7 +550,8 @@ <h2>CLI Configuration</h2>
       var statusEl = get('logout-status');
       function done() {
         if (statusEl) {
-          statusEl.textContent = 'Signed out. Your stored credentials were cleared — reopen Settings to sign in again.';
+          statusEl.className = statusEl.className.replace(/\bhidden\b/, '').trim();
```
andrewrobinsonhodges-snyk, line 553: Would strongly prefer that we used classList or similar instead of manipulating the names. This feels a bit fragile.

andrewrobinsonhodges-snyk (approved):

shared_ide_resources/ui/html/settings-fallback.html
rrama: Oh right, we do have access to `classList` now, with the the Visual Studio upgrade 😁
Ooo, we can start re-writing with moden JS features 😍

shared_ide_resources/ui/html/settings-fallback.html
rrama: Good catch — fixed in c9babd1c. `aria-live="polite"` on the `<p>` was inert since the element started `display:none`; screen readers generally won't pick up a live region that isn't in the a11y tree at load. Moved `aria-live` to the wrapping `.form-group` (always visible) instead — the `<p>`'s hidden/text toggle is unchanged, and only the changed text node gets announced (aria-atomic defaults to false), so the button label won't get re-announced.

shared_ide_resources/ui/html/settings-fallback.html
rrama: Agreed it's fragile, but this exact `className.replace(/\bhidden\b/, '')` pattern is used 4 other times in this same file (and elsewhere in the repo), so switching just this occurrence would make the file inconsistent. Filed IDE-2312 to modernise the JS/HTML across the repo now that all IDE hosts render through a Chromium-based webview (no more IE11 constraint) — leaving this as-is for now.

PR #1390, opened 2026-07-27
fix: return installed CLI path from download instead of re-discovering it via Find [IDE-2322]

basti-snyk:
## Summary

Fixes the flaky `infrastructure/oss` CI failure where `TestMain` aborted the whole package on a CLI cache-miss run.

`installRelease` discarded the destination the downloader had just written and re-derived the CLI path through `Find()`. `Find()`'s three-tier lookup covers the configured path, the XDG data dir and `$PATH` — but not the CLI cache dir. So on a cache-miss run the binary was downloaded, checksummed, moved and made executable, and the rediscovery still failed with `unable to find snyk-linux in PATH`, aborting the package via `TestMain`'s `log.Fatalf` and silently losing its coverage.

The rediscovery was unreliable because the configured CLI path is read through the config resolver, and that read is not guaranteed to return the same value twice: a value written under a `user:global:` scoped key only round-trips when the resolver carries a scope annotation for it, and it can also be rewritten concurrently.

## What changed

- The configured CLI path is resolved **exactly once per logical operation** and threaded through, instead of re-read.
- `Downloader` no longer holds a `ConfigResolverInterface`; `Download` takes the resolved path and returns the destination it actually wrote.
- `installRelease` returns that path directly. `updateFromRelease` passes its one resolved value to both `Download` and `replaceOutdatedCli`, so the download and rename destinations cannot diverge.
- An empty configured path is a hard error rather than a silent write to a cwd-relative filename.
- `Find()` is unchanged and still locates an already-installed CLI.
- `FakeInstaller.Install` returns the path it wrote instead of an empty string.

## Correction — a regression this PR introduced, and fixed

The first commit also changed `moveToDestination`'s remove-failure branch, which had returned a nil error (silently skipping the rename). An earlier revision of this description called that "an incidental latent bug fix, not a regression". **That was wrong**, and CI proved it: `smoke (1, windows-latest)` and `smoke (2, windows-latest)` failed with

```
couldn't remove old CLI at C:\...\snyk-win.exe ...: Access is denied.
FAIL	github.com/snyk/snyk-ls/infrastructure/oss
```

That swallowed error was load-bearing. Within one `go test ./...` run packages execute in parallel, so `infrastructure/oss`'s `TestMain` and the `application/server` smoke helpers install the CLI into the same cache dir concurrently; when one holds the destination open, the other's `os.Remove` — and equally its `os.Rename` — fails. POSIX unlink-while-open hides this on Linux and macOS entirely, which is why a full local 4-shard smoke run was green and only 2 of 4 Windows shards caught it.

The second commit treats "the requested binary is already there" as success rather than failure: on a remove **or** rename failure, the file already at the destination is compared against the release's expected checksum — match means another installer wrote exactly what we wanted, so its path is returned; mismatch or unreadable returns the original wrapped error. Both call sites share one helper. A genuinely failed operation still reports failure.

## Verification

| Check | Result |
|---|---|
| `go build ./...` / `go vet ./...` | pass |
| `golangci-lint` | pass, 0 issues |
| `go test ./infrastructure/cli/install/... -race -count=20` | pass |
| `go test ./... -race` | pass |
| `make test` | pass, 0 failures |
| `make test-smoke-parallel` (4 shards) | pass, 0 failures |

The primary regression test was proven to discriminate: reverting `installRelease` to `return i.Find()` makes `TestInstallRelease_ReturnsInstalledPath_WhenResolverCannotRediscoverIt` fail with the exact CI error. A real cache-miss install was also exercised end-to-end against an empty cache dir with a genuine ~163 MB download.

**No local tier can exercise the Windows failure path** — POSIX semantics hide it. Windows CI is the only real arbiter for that part, which is precisely how the regression above slipped through review.

## Known gaps

- ~~`TestInstallRelease_ReturnsInstalledPath_WhenCacheDirOutsideDiscovery` does not discriminate against the pre-fix code.~~ **Fixed** in `35cdaa63` — renamed to `TestInstallRelease_ReturnsConfiguredDestination_OnSuccessfulInstall`, which describes what it actually asserts. The genuine regression guard remains `..._WhenResolverCannotRediscoverIt`.
- Reviewer suggestions accepted but not acted on: the checksum-match fallback does not re-assert the executable bit (moot on Windows; a low-probability compound race on POSIX), and the debug log does not distinguish a checksum mismatch from an unreadable file.
- Snyk SCA via MCP did not complete (its trust step timed out); it was run via the CLI instead — 9 pre-existing dependency vulnerabilities, none introduced here, no manifests changed. Snyk Code reports 8 pre-existing findings in the unchanged `benchmark/testdata` fixture.

## Test plan

- [ ] CI green, **including all Windows smoke shards**
- [ ] Confirm a cache-miss smoke run no longer aborts `infrastructure/oss`

---

This fix was produced by an automated flake-fix loop.

___

### **PR Type**
Bug fix, Enhancement

___

### **Description**
- Fixes flaky CI by improving CLI path resolution.

- CLI path resolved once and passed explicitly.

- Enhances handling of concurrent CLI installations.

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A[Installer] -->|Get CLI Path| B(Config Resolver)
  A -->|Download CLI| C{Downloader}
  C -->|Uses CLI Path| D[File Operations]
  D -->|Handles Conflicts| E[File System]
  A -->|Returns Installed Path| F[Caller]
  subgraph Refactor
    C -- "Explicit Path" --> D
    D -- "Concurrency Handling" --> E
  end
  B -- "Resolved Once" --> A
```

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Tests</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>server_smoke_test.go</strong><dd><code>Smoke Test Adjusts Preview CLI Path Handling</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

application/server/server_smoke_test.go

<ul><li>Updates the test to pass the specific <code>previewCLIPath</code> to the <br><code>Downloader.Download</code> method.<br> <li> Aligns the test with the refactored downloader API that expects an <br>explicit path.</ul>

</details>

  </td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1390/changes#diff-13396c212387ae79a286802e0885b1c86f1bc389f0f6e82de58d49d641e15e08">+2/-3</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>downloader_test.go</strong><dd><code>Update Downloader Tests for New API</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

infrastructure/cli/install/downloader_test.go

<ul><li>Updates existing tests to accommodate the refactored <br><code>Downloader.Download</code> method signature.<br> <li> Ensures tests correctly pass the destination path and verify the <br>returned installed path.</ul>

</details>

  </td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1390/changes#diff-bc987779bba3bce6c86614e040335b0da96085c7dc9e7ce403ad6e5a6c4fa13c">+3/-7</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>installer_test.go</strong><dd><code>Add Extensive Tests for CLI Install and Download</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

infrastructure/cli/install/installer_test.go

<ul><li>Adds new unit tests to cover various scenarios for CLI installation <br>and download, including:<br>   - Empty CLI path configuration.<br>   - <br>Concurrent installation conflicts and resolutions.<br>   - Correct path <br>resolution and return values.<br>   - <code>FakeInstaller</code>'s behavior.<br> <li> Introduces mocks for <code>ConfigResolverInterface</code> to isolate test cases.</ul>

</details>

  </td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1390/changes#diff-1708a6a27d97bc0f7269e4994bf6e7c8f82df1cbbd91d93cce862aa579f9ecdb">+361/-0</a>&nbsp; </td>

</tr>
</table></td></tr><tr><td><strong>Enhancement</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>downloader.go</strong><dd><code>Refactor Downloader, Improve Concurrency Handling</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

infrastructure/cli/install/downloader.go

<ul><li>Refactors the <code>Downloader</code> to no longer depend on <br><code>ConfigResolverInterface</code>.<br> <li> The <code>Download</code> method now accepts the target <code>cliPath</code> and returns the <br>installed path.<br> <li> Introduces more robust handling for concurrent CLI installations by <br>checking existing files against checksums.<br> <li> Adds dependency injection for <code>os.Remove</code> and <code>os.Rename</code> for improved <br>testability.</ul>

</details>

  </td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1390/changes#diff-7ba28fb2e4a170e8d1517e55166f1731bf866b6b0ce21759b8cca002b0b7fc06">+69/-36</a>&nbsp; </td>

</tr>
</table></td></tr><tr><td><strong>Bug fix</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>installer.go</strong><dd><code>Integrate Refactored Downloader in Installer</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

infrastructure/cli/install/installer.go

<ul><li>Integrates the refactored <code>Downloader</code> in <code>installRelease</code> and <br><code>updateFromRelease</code>, passing the resolved CLI path.<br> <li> Ensures the <code>installRelease</code> method now returns the actual installed CLI <br>path.<br> <li> Fixes <code>FakeInstaller.Install</code> to return the path it writes to, aligning <br>with the new contract.<br> <li> Removes redundant CLI path lookups within <code>updateFromRelease</code> and <br><code>replaceOutdatedCli</code>.</ul>

</details>

  </td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1390/changes#diff-8421d797f924e31a4db10413bf4cdac016fad3dae3f424d1035ac9377c56fbd5">+17/-16</a>&nbsp; </td>

</tr>

</details>

___

### DIFF
--- a/infrastructure/cli/install/downloader.go
+++ b/infrastructure/cli/install/downloader.go
@@ -36,6 +36,8 @@ type Downloader struct {
 	errorReporter   error_reporting.ErrorReporter
 	httpClient      func() *http.Client
 	engine          workflow.Engine
+	removeFile      func(string) error
+	renameFile      func(string, string) error
 }
 
 func NewDownloader(engine workflow.Engine, errorReporter error_reporting.ErrorReporter, httpClientFunc func() *http.Client) *Downloader {
@@ -44,6 +46,8 @@ func NewDownloader(engine workflow.Engine, errorReporter error_reporting.ErrorRe
 		errorReporter:   errorReporter,
 		httpClient:      httpClientFunc,
 		engine:          engine,
+		removeFile:      os.Remove,
+		renameFile:      os.Rename,
 	}
 }
 
@@ -194,7 +198,7 @@ func (d *Downloader) Download(r *Release, cliPath string, isUpdate bool) (string
 	}
 
 	_ = cliTmpFile.Close() // close file to allow moving it on Windows
-	destinationPath, err := d.moveToDestination(cliPath, executableFileName, cliTmpFile.Name())
+	destinationPath, err := d.moveToDestination(cliPath, executableFileName, cliTmpFile.Name(), expectedChecksum)
 
 	if isUpdate {
 		d.progressTracker.EndWithMessage("Snyk CLI has been updated.")
@@ -224,7 +228,30 @@ func (d *Downloader) destinationPath(cliPath string, destinationFileName string)
 	return filepath.Join(filepath.Dir(cliPath), destinationFileName)
 }
 
-func (d *Downloader) moveToDestination(cliPath string, destinationFileName string, sourceFilePath string) (string, error) {
+func (d *Downloader) remove(path string) error {
+	if d.removeFile != nil {
+		return d.removeFile(path)
+	}
+	return os.Remove(path)
+}
+
+func (d *Downloader) rename(sourceFilePath string, destinationFilePath string) error {
+	if d.renameFile != nil {
+		return d.renameFile(sourceFilePath, destinationFilePath)
+	}
+	return os.Rename(sourceFilePath, destinationFilePath)
+}
+
+func (d *Downloader) existingDestinationMatchesChecksum(expectedChecksum HashSum, destinationFilePath string) bool {
+	compareErr := compareChecksum(d.engine.GetLogger(), expectedChecksum, destinationFilePath)
+	if compareErr == nil {
+		return true
+	}
+	d.engine.GetLogger().Debug().Err(compareErr).Str("path", destinationFilePath).Msg("existing Snyk CLI does not match requested checksum")
+	return false
+}
+
+func (d *Downloader) moveToDestination(cliPath string, destinationFileName string, sourceFilePath string, expectedChecksum HashSum) (string, error) {
 	// Defend this file-moving boundary when it is called independently of Install.
 	if cliPath == "" {
 		return "", fmt.Errorf("CLI path is not configured")
@@ -244,25 +271,33 @@ func (d *Downloader) moveToDestination(cliPath string, destinationFileName strin
 
 	// for Windows, we have to remove original file first before move/rename
 	if fileInfo, statErr := os.Stat(destinationFilePath); statErr == nil {
-		removeErr := os.Remove(destinationFilePath)
+		removeErr := d.remove(destinationFilePath)
 		if removeErr != nil {
 			returnErr := errors.Wrap(
 				removeErr,
 				fmt.Sprintf("couldn't remove old CLI at %s. FileInfo: %v", destinationFilePath, fileInfo),
 			)
+			if d.existingDestinationMatchesChecksum(expectedChecksum, destinationFilePath) {
+				logger.Info().Str("path", destinationFilePath).Msg("another installer already wrote the requested Snyk CLI")
+				return destinationFilePath, nil
+			}
 			logger.Err(returnErr).Send()
 			return "", returnErr
 		}
 	}
 
 	logger.Debug().Str("tempFilePath", sourceFilePath).Msg("tempfile path")
-	err = os.Rename(sourceFilePath, destinationFilePath)
+	err = d.rename(sourceFilePath, destinationFilePath)
 	if err != nil {
 		returnErr :=
 			errors.Wrap(
 				err,
 				fmt.Sprintf("couldn't rename Snyk CLI from %s to %s", sourceFilePath, destinationFilePath),
 			)
+		if d.existingDestinationMatchesChecksum(expectedChecksum, destinationFilePath) {
+			logger.Info().Str("path", destinationFilePath).Msg("another installer already wrote the requested Snyk CLI")
+			return destinationFilePath, nil
+		}
 		logger.Err(returnErr).Send()
 		return "", returnErr
 	}
--- a/infrastructure/cli/install/downloader.go
+++ b/infrastructure/cli/install/downloader.go
@@ -82,12 +82,15 @@ func (d *Downloader) lockFileName() (string, error) {
 	return config.CLIDownloadLockFileName(d.engine.GetConfiguration())
 }
 
-func (d *Downloader) validateDownloadPreconditions(r *Release) error {
+func (d *Downloader) validateDownloadPreconditions(r *Release, cliPath string) error {
 	if r == nil {
 		return fmt.Errorf("release cannot be nil")
 	}
 	if d.httpClient == nil {
 		return fmt.Errorf("http client function is not configured")
+	}
+	if cliPath == "" {
+		return fmt.Errorf("CLI path is not configured")
 	}
 	return nil
 }
@@ -99,8 +102,8 @@ func downloadKind(isUpdate bool) string {
 	return "download"
 }
 
-func (d *Downloader) Download(r *Release, cliPath string, isUpdate bool) (string, error) {
-	if err := d.validateDownloadPreconditions(r); err != nil {
+func (d *Downloader) Download(r *Release, cliPath string, isUpdate bool) (destinationPath string, err error) {
+	if err = d.validateDownloadPreconditions(r, cliPath); err != nil {
 		return "", err
 	}
 
@@ -127,6 +130,22 @@ func (d *Downloader) Download(r *Release, cliPath string, isUpdate bool) (string
 		d.progressTracker.BeginWithMessage("Downloading Snyk CLI...", "We download Snyk CLI to run security scans.")
 	}
 
+	// Begin was just sent above, so from this point on every return path
+	// (early or final) must close the progress indicator exactly once. A
+	// deferred function keyed off the named return values guarantees this
+	// regardless of which return statement is hit.
+	defer func() {
+		if err != nil {
+			d.progressTracker.EndWithMessage(fmt.Sprintf("Failed to %s Snyk CLI.", kindStr))
+			return
+		}
+		if isUpdate {
+			d.progressTracker.EndWithMessage("Snyk CLI has been updated.")
+		} else {
+			d.progressTracker.EndWithMessage("Snyk CLI has been downloaded.")
+		}
+	}()
+
 	doneCh := make(chan struct{}, 1)
 
 	var resp *http.Response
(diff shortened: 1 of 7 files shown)


infrastructure/cli/install/downloader.go
```diff
@@ -98,10 +99,11 @@ func downloadKind(isUpdate bool) string {
 	return "download"
 }
 
-func (d *Downloader) Download(r *Release, isUpdate bool) error {
+func (d *Downloader) Download(r *Release, cliPath string, isUpdate bool) (string, error) {
 	if err := d.validateDownloadPreconditions(r); err != nil {
```
nick-y-snyk: 🟡 **[important]** `Download()` never validates `cliPath` before doing the network GET and writing a temp dir. `validateDownloadPreconditions` only checks `r == nil` and `d.httpClient == nil`.

Both current callers (`installRelease`, `updateFromRelease`) guard `cliPath == ""` before calling `Download`, but `Download` is exported and already called directly elsewhere (`server_smoke_test.go`). A future direct caller with an empty path hits `filepath.Dir("")` -> `"."`, so `os.MkdirAll(".", ...)` and `os.MkdirTemp(".", "downloads")` succeed and create a `downloads*` dir in the CWD — the full release download (~163MB per this PR's own verification) happens before `moveToDestination` finally rejects the empty path. That's exactly the "silent write to a cwd-relative filename" this PR says it fixed, just one layer further out.

Suggest adding the `cliPath == ""` check to `validateDownloadPreconditions` so it's enforced at the top of `Download` itself, not only at the two current call sites.

nick-y-snyk (commented):
Solid fix overall — the regression test (`TestInstallRelease_ReturnsInstalledPath_WhenResolverCannotRediscoverIt`) genuinely discriminates against the old `Find()`-based rediscovery, and the Windows concurrent-install race the second commit fixes is well documented and tested. Two small items plus one pre-existing observation below; none of these block merge.

infrastructure/cli/install/installer.go
```diff
@@ -125,40 +131,35 @@ func (i *Install) updateFromRelease(r *Release) (bool, error) {
 	}
 
 	cliPath := i.configResolver.GetString(types.SettingCliPath, nil)
-	if cliPath != "" {
-		cliPath = filepath.Clean(cliPath)
+	if cliPath == "" {
```
nick-y-snyk, line 134: 🟢 **[nit]** This new empty-`cliPath` hard-error guard has no direct test, unlike the equivalent guard in `installRelease` (`TestInstallRelease_ReturnsErrorWhenCliPathIsEmpty`). `TestInstallerUpdate_ResolvesCliPathOnce` and `TestInstaller_Update_DoesntUpdateIfNoLatestRelease` both always configure a non-empty path, so a future refactor could silently drop this check without any red test.

infrastructure/cli/install/downloader.go
```diff
@@ -145,7 +147,7 @@ func (d *Downloader) Download(r *Release, isUpdate bool) error {
 
 	if resp.StatusCode != http.StatusOK {
 		d.errorReporter.CaptureError(err)
-		return fmt.Errorf("failed to %s Snyk CLI from %q: %s", kindStr, downloadURL, resp.Status)
+		return "", fmt.Errorf("failed to %s Snyk CLI from %q: %s", kindStr, downloadURL, resp.Status)
```
nick-y-snyk: 📚 **[learning, pre-existing]** Not introduced by this diff (only the return arity changed here), but flagging since this function is being touched: on this `resp.StatusCode != http.StatusOK` branch, `Download` returns before the `defer func(Body io.ReadCloser) { ... doneCh <- struct{}{} ... }` a few lines below gets registered. The goroutine spawned right after `d.httpClient().Get(downloadURL)` blocks in `d.progressTracker.CancelOrDone(cancel, doneCh)` waiting on `doneCh`, which now never fires — that goroutine and the open response body leak until process exit (unless the tracker is externally cancelled). Might be worth a follow-up to register the body-close defer immediately after the successful `Get()`, before the status check.

nick-y-snyk (approved):

infrastructure/cli/install/downloader.go
nick-y-snyk: Fixed in 8a72818e: `cliPath == ""` check moved into `validateDownloadPreconditions`, so `Download()` itself rejects an empty path before the network GET, regardless of caller.

infrastructure/cli/install/installer.go
nick-y-snyk: Fixed in 8a72818e: added `TestUpdateFromRelease_ReturnsErrorWhenCliPathIsEmpty`, which exercises this exact guard on the `updateFromRelease` path.

infrastructure/cli/install/downloader.go
nick-y-snyk: Fixed in 8a72818e: the body-close defer is now registered immediately after a successful `Get()`, before the status-code check, so a non-200 response can no longer leak the goroutine/body. Covered by `TestDownloaderDownload_ReportsFailureEndWhenHttpGetFails` and the new non-200 close test.

PR #1396, opened 2026-08-04
fix: stabilize Test_IsAuthenticated_DoesNotUseOAuth2ProviderCustomRefresherFunc flaky test [IDE-2402]

basti-snyk:
## What
Fixes a flaky test caused by a real cross-goroutine ordering race in `infrastructure/authentication`.

## Root cause
A queued async credential update (from a stalled OAuth2 storage-bridge callback) could be applied by `credentialUpdateWorker` *after* a synchronous `logout()` had already cleared the token — reviving an expired/cleared token nondeterministically depending on goroutine scheduling. This is not a Go memory-model data race (not caught by `-race`), but an application-level ordering race between two independently-locked writers.

## Fix
Added a generation counter (`credentialGeneration`) guarded by `credentialWriteMu`:
- Every direct/synchronous credential write bumps the counter.
- Queued async updates are stamped with the counter value at enqueue time.
- The worker drops a queued update if a newer direct write has occurred since it was queued.

`logout()`, `UpdateCredentials()`, and `finishAuthenticate()` now all route through a shared `updateCredentialsDirect` so the counter is bumped consistently on every synchronous write.

## Verification
- Added `Test_QueueCredentialUpdate_StaleQueuedTokenDoesNotClobberSynchronousLogout`, a deterministic regression test reproducing the race.
- Before the fix: reliably reproduced within ~1300 `-count=1` stress iterations.
- After the fix: `go test -race -count=100` — 200/200 runs pass.
- Full unit suite (`make test`) green at this commit.

_Produced by an automated flake-fix loop (Jira IDE-2402)._

### DIFF
--- a/infrastructure/authentication/auth_configuration_test.go
+++ b/infrastructure/authentication/auth_configuration_test.go
@@ -483,6 +483,62 @@ func Test_RegisterOAuthStorageBridge_EmptyStorageUpdateDoesNotClearAppliedToken(
 	)
 }
 
+// Test_QueueCredentialUpdate_StaleQueuedTokenDoesNotClobberSynchronousLogout is the
+// deterministic regression test for IDE-2402.
+//
+// Root cause: while doAuthCheck attempts to verify a token, GAF's OAuth2 authenticator
+// tries to refresh it; the refresh fails, but syncTokenRefresh's storage.Refresh() still
+// re-Sets the unchanged, still-expired token from disk into config, and that Set fires
+// the OAuth storage bridge callback, which queues an async credential update carrying
+// the SAME stale token via QueueCredentialUpdate. Moments later, once the auth check
+// definitively fails, doAuthCheck synchronously calls logout(), which clears the token.
+// Whether credentialUpdateWorker drains the queued stale update before or after that
+// synchronous clear is an unsynchronized ordering race: if the worker runs after the
+// clear, it reapplies the stale (expired) token, undoing the clear — the token
+// reappears in storage byte-for-byte, exactly as observed in IDE-2402.
+//
+// This test reproduces the adversarial ordering directly: queue the stale token first,
+// then synchronously clear via UpdateCredentials(""), mirroring logout(). require.Never
+// polls continuously for 2 seconds: without the fix the worker eventually (and often
+// immediately) drains the queued update after the clear and the token reappears
+// (reliable RED); with the generation-counter guard the queued update is recognized as
+// stale (superseded by the direct clear) and dropped, so the token stays cleared
+// (reliable GREEN), across -race and all GOMAXPROCS values.
+func Test_QueueCredentialUpdate_StaleQueuedTokenDoesNotClobberSynchronousLogout(t *testing.T) {
+	engine, tokenService := testutil.UnitTestWithEngine(t)
+	conf := engine.GetConfiguration()
+	conf.Set(configresolver.UserGlobalKey(types.SettingAuthenticationMethod), string(types.OAuthAuthentication))
+
+	service := NewAuthenticationService(
+		engine,
+		tokenService,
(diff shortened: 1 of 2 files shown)


infrastructure/authentication/auth_service_impl.go
```diff
@@ -156,6 +175,19 @@ func (a *AuthenticationServiceImpl) credentialUpdateWorker(ctx context.Context)
...
+					// back by GAF's OAuth2 authenticator while it attempted a refresh —
+					// must not clobber it. See IDE-2402.
+					a.engine.GetLogger().Debug().
+						Str("method", "AuthenticationService.credentialUpdateWorker").
+						Msg("dropping stale queued credential update superseded by a direct write")
+					return
+				}
 				a.writingToken.Store(&update.token)
 				defer a.writingToken.Store(nil)
 				a.updateCredentials(update.token, update.sendNotification, update.updateApiUrl)
```
nick-y-snyk, line 193: 🟡 **[important]** `credentialWriteMu` scope too broad — this call runs the *entire* `updateCredentials` body (token write + `postCredentialUpdateHook()` + `notifier.Send`) under the lock, not just the generation check.

`postCredentialUpdateHook` can do real I/O: `login.go:110-115` sets a hook that calls `ldxSyncService.RefreshConfigFromLdxSync(...)` (network call). Before this PR the worker held zero service-level lock while calling `updateCredentials`. Now: worker mid-processing a queued update (running a slow hook) blocks `Logout()`/`UpdateCredentials()`/`finishAuthenticate()` — all take `credentialWriteMu` via `updateCredentialsDirect` (line 657) — until the hook returns. User-facing symptom: Logout hangs on an unrelated in-flight token-refresh's network call. Coupling runs both directions: a slow hook during login also blocks the worker from draining other queued updates.

Only the generation-check + token write need the mutex — that's the actual guarded invariant. Suggest releasing `credentialWriteMu` before calling the hook/notifier, or splitting `updateCredentials` into a guarded core (token write) and an unguarded tail (hook, notify).

Related, not currently reachable: hook now runs while `credentialWriteMu` is held (non-reentrant `sync.Mutex`). If a future hook ever calls back into `UpdateCredentials`/`logout`/`finishAuthenticate` on the same goroutine, that's a self-deadlock. Today's hook (`login.go:110`) only calls `FlushCache`/`RefreshConfigFromLdxSync`/`populateFolderFeatureFlagsAndSastSettings`, none loop back — but worth a warning comment on `SetPostCredentialUpdateHook` for future hook authors.

infrastructure/authentication/auth_configuration_test.go
```diff
@@ -483,6 +483,62 @@ func Test_RegisterOAuthStorageBridge_EmptyStorageUpdateDoesNotClearAppliedToken(
...
+	serviceImpl, ok := service.(*AuthenticationServiceImpl)
+	require.True(t, ok)
+
+	// Simulate GAF's storage-bridge echo of the unchanged, still-expired token during a
+	// failed refresh attempt.
+	serviceImpl.QueueCredentialUpdate(staleToken, true, false)
+	// Simulate doAuthCheck's subsequent synchronous logout clearing the token.
+	service.UpdateCredentials("", true, false)
+
+	require.Never(t,
```
nick-y-snyk, line 535: 🟢 **[nit]** Outcome here is deterministic under the mutex fix (no timing race left to poll for), so this pays a full 2s wall-clock on every CI run for one immediate check's worth of confidence. Not blocking.

nick-y-snyk (commented):
Fix itself sound: generation counter + `credentialWriteMu` correctly serializes worker's check-and-apply against direct writes, closes the IDE-2402 ordering race, regression test is deterministic by construction. One important finding below before approve, plus a nit.

PR #1397, opened 2026-08-05
fix: delta findings fail open when no baseline exists [IDE-2418]

basti-snyk:
## Summary

This PR started as CP0 (BDD scaffolding only) for IDE-2418. The sibling PRs originally planned as separate stacked PRs on top of it (#1399, #1400, #1401, #1402, #1404) were squash-merged into this branch instead, so it now carries the full delta-fail-open fix plus the feature work that rode alongside it:

- **Godog BDD harness**: runner + shared step registry reusing the existing `application/server` test harness, a requirement-mapping lint enforcing every tracked requirement has a `# maps:` anchor, a `bdd` CI job, and a coverage gate that hard-fails `feat:` commits missing feature coverage (warns on `fix:`)
- **IDE-2418 fix**: delta/net-new findings fail open (show every current finding) instead of silently returning zero when no baseline exists yet — per-product baseline availability is now tracked (`Folder.IsBaselineAvailable`/`IsDeltaAppliedForProduct`/`DeltaStatusForProduct`), defaulting to "no baseline" so an unscanned or corrupted-snapshot product fails open rather than hiding everything
- **Ambient Canary autonomy** (IDE-2426): per-folder `ambient_canary_autonomy` setting (`notify_only` / `autonomous_fixes`), with UI, reset-field, and source-indicator coverage
- **LLM provider settings for autonomous remediation** (IDE-2274): `llm_provider` / `llm_base_url` / `llm_model` settings, wired through to the actual Remy fix workflow invocation (IDE-2430), never touching the API key (still developer-environment-only)
- Configuration dialog UI updates for all of the above

## Test plan
- [x] `make lint` clean
- [x] `make test` clean
- [x] `make test-bdd` clean — 12/12 BDD scenarios passing (harness, delta fail-open, Ambient Canary autonomy, LLM provider dialog, remediation wiring)
- [x] Ghost-check: removed a `# maps:` anchor, confirmed the lint test fails, restored it

___

### **PR Type**
Enhancement, Documentation, Tests

___

### **Description**
- Introduce Ambient Canary Autonomy settings for per-folder configuration.

- Implement LLM provider settings for autonomous remediation features.

- Enhance delta findings to "fail open" when no baseline exists.

- Add a comprehensive BDD acceptance test suite for core features.

___

### Diagram Walkthrough

```mermaid
flowchart LR
  subgraph "Core Logic"
    A[Delta Findings Logic Update] --> B{Folder Baseline State};
    B --> C[Fail Open on Missing Baseline];
  end
  subgraph "Configuration"
    D[New LLM Provider Settings] --> E{UI & Persistence};
    F[New Ambient Canary Autonomy Settings] --> E;
    E --> G[Configuration Dialog Update];
  end
  subgraph "Testing & CI"
    H[BDD Harness & Scenarios] --> I{CI Coverage Gate};
    J[New Test Setup & Fixtures] --> H;
  end
  subgraph "API & Protocol"
    K[Update LSP Scan Params] --> L{DeltaStatus Field};
  end
  B -- Manages --> M[Folder Object Baseline Status];
  E -- Configures --> N[LLM Remediation Agent];
  F -- Configures --> O[Ambient Canary Autonomy];
  G -- Impacts --> P[Configuration UI];
  I -- Enforces --> Q[Commit Quality];
```

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Tests</strong></td><td><details><summary>25 files</summary><table>
<tr>
  <td><strong>bdd_coverage_gate.sh</strong><dd><code>BDD coverage gate script for commit quality</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-7f925f90eecfcc8b6f986edd50130bf466216f804da4a0160e0283db68b71c1d">+47/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>build.yaml</strong><dd><code>Add BDD job and coverage gate to CI workflow</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-d0777657fa3fd81d23aaf7273e58aee453b04e67882517900c56daeef9b3e4c1">+32/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>Makefile</strong><dd><code>Add 'test-bdd' target for running BDD suite</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52">+5/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>test_init.go</strong><dd><code>Update test initialization for ScanSnapshotPersister</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-fc304a8ceac9f55997a1ed4f30020de933d3ada1a11eacd32c236acdce0cec81">+6/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>bdd_lint.go</strong><dd><code>Implement feature file parsing and mapping lint logic</code>&nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-29ccc0d2c485b045f993f772d9dedb0963ae896a83070eeb91f999d894006e2c">+166/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>bdd_lint_test.go</strong><dd><code>Add unit tests for BDD feature file parsing and linting</code>&nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-f9db4e840aa1a6ab9023373576c500997eb8c8560eb756a73fc72d01de7ab272">+104/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>bdd_steps_test.go</strong><dd><code>Implement BDD step definitions for core features and delta fail-open</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-5fcf20b0c028ab75ebe77c10364d1656a95f054114f46d6a34e4b9057048c693">+1344/-0</a></td>

</tr>

<tr>
  <td><strong>bdd_test.go</strong><dd><code>Scaffold BDD test suite runner using godog</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-7a28de939e1b93109f6cb9df4dc1df943021e2e9dd04db85789fcd03b090d9c4">+51/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>configuration_test.go</strong><dd><code>Add tests for LLM provider settings and Ambient Canary Autonomy</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-e141faafb3f0507956b85e92ac3b584c7c706bd96e7f428c7ef95d0d8b2df564">+583/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>server_smoke_test.go</strong><dd><code>Update smoke test to account for DeltaStatus in scan params</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-13396c212387ae79a286802e0885b1c86f1bc389f0f6e82de58d49d641e15e08">+8/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>configuration_command_test.go</strong><dd><code>Add test for Ambient Canary Autonomy in effective config</code>&nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-4e6fcd9606a6e827d54cddeaf143ea61e51516a7e24ebcee13d0968284546bb0">+15/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>folder_test.go</strong><dd><code>Add extensive tests for baseline availability and delta filtering </code><br><code>logic</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-bb03ee76aabde8be49355b245229584151a114e49ec791407738dcaeb631acaf">+278/-6</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>workspace_test.go</strong><dd><code>Update workspace constructor and add tests for folder management</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-8b15a4314c4b8fec249174af4d164a17c6e0f8686690a5a03db609fc40cb9afd">+17/-16</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>nop_scan_persister.go</strong><dd><code>Change NopScanPersister to return ErrBaselineDoesntExist</code>&nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-65c268300051850a966f051700dbc27826a301a7577e2f5999ef8a5ac279ba50">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>remy_config_test.go</strong><dd><code>Add tests for LLM provider and model forwarding in fix config</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-e85d72f83c8cd6e545da4d1e71f158b9dfe6a7e6dac9cba34170ba471118079e">+109/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>config_html_test.go</strong><dd><code>Add tests for LLM provider UI rendering and Ambient Canary Autonomy</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-1309718000d7ccd2382924bb48641b806331f719524d77103362a9458fa2f322">+158/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>config_resolver_test.go</strong><dd><code>Test Ambient Canary Autonomy persistence and rendering</code>&nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-9d6e7bc1195e5dec96d9355a8c2226943a7655521daa1eb60b7419a7acff08a8">+41/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>ldx_sync_config_test.go</strong><dd><code>Test exact name of Ambient Canary Autonomy setting</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-caeb581622ddc107fcb0467465f30a48b3caee2b535249111c5283e9ad1faa14">+6/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>workspace_mock.go</strong><dd><code>Mock IsDeltaAppliedForProduct for Folder interface</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-330c7206d108fa248f17c11655d6cf416932af116b2c98500f672e88803b28a3">+14/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>register_configurations_test.go</strong><dd><code>Update setting counts and add tests for new config registrations</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-8afddaa47efa6142660c7883d0f111283a11e421822a627c6fa6776910bb3971">+63/-1</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>delta-fail-open.feature</strong><dd><code>New scenarios for delta findings fail-open logic</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-d71608dc92b98be3c0bc0b252ccbf3e868e6fe3f8da73e50c7d60c7e7962e685">+61/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>harness.feature</strong><dd><code>Add Ambient Canary Autonomy scenario to harness feature</code>&nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-c05813c9cbdc56cc97957d47e5044d04ef3c33c78826afff3b6ccd8c28f008fa">+19/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>llm-provider-setting.feature</strong><dd><code>New scenarios for LLM provider settings and usage</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-98c2a71aa153dfaf4cbcba8123ebe1040fb5bcb542ec877091ba3ece5163b5b9">+39/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>go.mod</strong><dd><code>Add godog dependency for BDD testing</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6">+7/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>go.sum</strong><dd><code>Update dependency checksums</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-3295df7234525439d778f1b282d146a4f1ff6b415248aaac074e8042d9f42d63">+23/-0</a>&nbsp; &nbsp; </td>

</tr>
<tr>
  <td><strong>CONTRIBUTING.md</strong><dd><code>Document BDD acceptance test suite and requirements mapping</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055">+13/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Enhancement</strong></td><td><details><summary>14 files</summary><table>
<tr>
  <td><strong>config.go</strong><dd><code>Add LsServerName constant</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-d9a33e7e08d7d8fb8292de9d94d5bab14156125922662d7448de0951894ec11d">+1/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>configuration.go</strong><dd><code>Apply LLM provider configuration settings and manage environment</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-eb3fe6fd1956591838595b08979653ec23c0b0c16f5340e9e89a05c30e921607">+12/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>llm_provider_env.go</strong><dd><code>Manage LLM provider environment variables and concurrency</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-bdf3cc2b1153bdd1bd1762a060a74a2530f278b926212626283caafa71a9eb10">+174/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>server.go</strong><dd><code>Use LsServerName constant in initialize handler</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-1b5a200b7dc3f37ed67632d0c2682eeb53762813f96edb15cb7660e076fa6031">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>configuration_command.go</strong><dd><code>Include LLM and Ambient Canary settings in configuration output</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-33091090cd00338147252f2f1da05b8f20249df3e0118866bbd80412803378ce">+4/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>folder.go</strong><dd><code>Implement baseline availability checks for delta findings fail-open</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-92ee5447150d64c4b9a6c8c8e8834577ada9e739d1a0a2842ef5bc79fd014f26">+76/-23</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>remy.go</strong><dd><code>Integrate LLM provider environment variable management into </code><br><code>remediation runner</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-3ad54b822be438a8117c900dfa6cdcf87b84569bfb85139751524795371d6274">+44/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>config.html</strong><dd><code>Add UI for LLM provider settings and Ambient Canary Autonomy</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-700a5eb160e325cc14adc0153f9fb936b41e4667abaf10337d6e2696d7644762">+67/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>form-handler.js</strong><dd><code>Add Ambient Canary Autonomy to folder reset fields</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-631e55de1a3395dd284c5a8bb15cc71fa9de087f03b3442858eae129883538ee">+1/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>reset-handler.js</strong><dd><code>Add default Autonomous Remediation settings to reset handler</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-4b26124d7fbe25fbe7b9c22f55d72d52dc3457ef6dc56b70e967cec69bed7987">+5/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>ldx_sync_config.go</strong><dd><code>Add constants for LLM provider and Ambient Canary Autonomy settings</code></dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-e42a6f2c9aece134bd052ddbd7e835aefd3661d3d5f9446f7d5b621c421eea4d">+4/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>lsp.go</strong><dd><code>Add DeltaStatus type and include in SnykScanParams</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-c16b9b2fc8b4479b07708b28ebdaa473e8f3a73eb62d9402538790b2aed3f978">+9/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>register_configurations.go</strong><dd><code>Register new LLM and Ambient Canary Autonomy settings</code>&nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-0048261bf7228f3c64fba7ab2d62bd98827fd0169045cbd19bb7f0e84d05e72a">+23/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>workspace.go</strong><dd><code>Add IsDeltaAppliedForProduct to Folder interface</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-982c4a5a61d388000a2e40c2d1d1b3b125505eaea41616ef5e5edc7d72165670">+3/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Additional files</strong></td><td><details><summary>7 files</summary><table>
<tr>
  <td><strong>configuration.md</strong></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-17ed18489a956f326ec0fe4040850c5bc9261d4631fb42da4c52891d74a59180">+2/-2</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>folder_handler_test.go</strong></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-657ad7fa06e97a16b4e8eb7cddc64cd5e53343e6513ef874563e83ffd694416b">+54/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>folder-reset.test.mjs</strong></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-e0ce3c3efaeb4dc15cf95634e4d489793840c566a5186e03864c7b6905093c32">+5/-4</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>form-payload.json</strong></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-f2f527afa56136c15aa90f2f1f93583632a8557947343671f7c650d75a39b28a">+8/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>config_output_multi_project.html</strong></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-7341a4c8b11a2e7e32215f48cda97caaafcd5faee07bdb8e3c1a44b7e2339e60">+115/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>config_output_no_projects.html</strong></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-7f80710ed2abc11c038fb7e756dfb95c1225bee7e380cd741655c84c868c72ed">+59/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>config_output_single_solution.html</strong></td>
  <td><a href="https://github.com/snyk/snyk-ls/pull/1397/changes#diff-40f815485d4acec22847d1036425f9b1a5ebf38e8043df9856e600654eb44f37">+73/-0</a>&nbsp; &nbsp; </td>

</tr>

</details>

___


application/server/bdd_steps_test.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func (s *bddSteps) aRunningLanguageServer() error {
+	engine, tokenService := testutil.UnitTestWithEngine(s.t)
+	loc, jsonRPCRecorder, _ := setupServer(s.t, engine, tokenService)
+	s.engine = engine
+	s.loc = loc
+	s.jsonRPCRecorder = jsonRPCRecorder
+	return nil
+}
```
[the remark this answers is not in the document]
basti-snyk: Fixed in `e0c5d526` — added `beforeScenario`/`afterScenario` godog hooks that run each scenario as its own `t.Run` subtest, so `setupServer`'s `t.Cleanup` now fires after each scenario instead of once at suite end. New test `Test_BDDSteps_PerScenarioCleanup` pins this (asserts scenario one's client is stopped before scenario two starts).

application/server/bdd_lint.go
```diff
@@ -0,0 +1,148 @@
...
+		for j := i - 1; j >= 0; j-- {
+			prev := strings.TrimSpace(lines[j])
+			if prev == "" {
+				continue
+			}
+			if mm := mapsLineRe.FindStringSubmatch(prev); mm != nil {
+				ids = splitRequirementIDs(mm[1])
+			}
+			break
+		}
```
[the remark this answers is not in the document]
basti-snyk: Fixed in `e0c5d526` — the backward scan now also skips `@`-prefixed Gherkin tag lines, not just blank lines. New subtest in `bdd_lint_test.go` proves a `@sometag` line between the anchor and `Scenario:` no longer breaks detection.

.github/workflows/build.yaml
```diff
@@ -44,6 +44,38 @@ jobs:
...
+          go-version-file: 'go.mod'
+          cache: true
+
+      - name: Setup tools (`make tools`)
+        uses: ./.github/actions/setup-tools
+
+      - name: Run BDD feature suite
+        run: make test-bdd
+
+      - name: Enforce BDD coverage for feat:/fix: commits
```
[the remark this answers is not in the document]
basti-snyk: Confirmed and fixed in `52c60300` — quoted the step name. Re-validated `yaml.safe_load` against the original file (fails exactly as you described) and the fix (parses clean), and scanned the rest of the file for the same unquoted-colon risk in other `name:` fields — found none. Thanks for catching this with an actual CI run, not just a static guess.

application/server/bdd_lint.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+var (
+	scenarioLineRe = regexp.MustCompile(`^(?:Scenario|Scenario Outline):\s*(.+)$`)
+	mapsLineRe     = regexp.MustCompile(`^#\s*maps:\s*(.+)$`)
+)
+
+// parseFeatureDir parses every *.feature file directly under dir.
+func parseFeatureDir(dir string) ([]featureScenario, error) {
+	entries, err := os.ReadDir(dir)
```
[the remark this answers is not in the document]
basti-snyk: Fixed in `086cd901` — switched `parseFeatureDir` to `filepath.WalkDir` so it recurses the same way godog's loader does. New test `Test_ParseFeatureDir_RecursesIntoSubdirectories` (confirmed RED against the old `os.ReadDir` implementation first) proves a nested `.feature` file is now found.

application/server/bdd_lint.go
```diff
@@ -0,0 +1,154 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("reading feature file %q: %w", path, err)
+		}
+		scenarios = append(scenarios, parseFeatureScenarios(entry.Name(), content)...)
+	}
+	return scenarios, nil
+}
+
+// parseFeatureScenarios scans the given feature file content for
+// Scenario/Scenario Outline lines and the `# maps:` comment anchored
```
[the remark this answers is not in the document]
basti-snyk: Fixed in `086cd901` — comment now accurately says the scan skips blank lines, @tags, and # comments.

application/server/bdd_steps_test.go
```diff
@@ -0,0 +1,233 @@
...
+
+func (s *bddSteps) afterScenario(ctx context.Context, sc *godog.Scenario, err error) (context.Context, error) {
+	close(s.scenarioDone)
+	<-s.scenarioDied
+	return ctx, err
+}
+
+// reported stops the deferred send from double-reporting when fn returns
+// normally instead of exiting via Goexit.
+func (s *bddSteps) runStep(fn func() error) {
```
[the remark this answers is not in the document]
basti-snyk: Fixed in `086cd901` — recover() added to runStep's deferred function, converting a genuine panic into an error on stepResult instead of crashing the suite. New test Test_BDDSteps_RunOnScenarioGoroutine_SurvivesPanic (confirmed it crashed the test binary before the fix). This also resolves the "Deadlock Risk" finding below — the send is now guaranteed on every path (return, Goexit, panic), so no separate timeout was needed.

application/server/bdd_steps_test.go
```diff
@@ -0,0 +1,233 @@
...
+		return fmt.Errorf("unmarshalling initialize result failed: %w", err)
+	}
+	return nil
+}
+
+func (s *bddSteps) theServerRespondsWithItsCapabilities() error {
+	if s.initResult.ServerInfo.Name == "" {
+		return fmt.Errorf("expected server info to be populated, got %+v", s.initResult.ServerInfo)
+	}
+	if s.initResult.ServerInfo.Version != config.LsProtocolVersion {
```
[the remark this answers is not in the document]
basti-snyk: Name assertion fixed in `88a0bf81` — now checks the literal "snyk-ls" from server.go:689. Investigated the version assertion: config.LsProtocolVersion is an ldflags-injected build var with no independent source of truth in a test binary, so a "stronger" assertion would mean mutating the same global the handler reads — no real coverage gain, just coupling to build-injection mechanics. This repo already has the identical accepted pattern elsewhere (Test_initialize_containsServerInfo, server_test.go:410), so left as-is rather than trading a real problem for a more brittle one.

.github/scripts/bdd_coverage_gate.sh
rrama, line 1: You will have to give us the skills, etc. to use this.

application/server/bdd_steps_test.go
rrama, line 1: More testing surfaces can't hurt, and it isn't worth my time to review this properly, so with a 60 second skim 👍 
It will be interesting to see how AIs handle editing such a long file, but I guess we shall see and address issues as they come.

application/server/bdd_test.go
```diff
@@ -0,0 +1,51 @@
+/*
+ * © 2022-2026 Snyk Limited
```
rrama: Not created in 2022. Check other files added too.

application/server/bdd_test.go
```diff
@@ -0,0 +1,51 @@
...
+// afterScenario already run every scenario as its own t.Run subtest of t, so
+// that setupServer's t.Cleanup fires per scenario. Also setting TestingT
+// would make godog wrap the same scenario in a second, colliding subtest.
+func TestBDD(t *testing.T) {
+	suite := godog.TestSuite{
+		ScenarioInitializer: func(sc *godog.ScenarioContext) {
+			newBDDSteps(t).register(sc)
+		},
+		Options: &godog.Options{
+			Format: "pretty",
```
rrama, line 42: Future enhancement: Maybe we could detect if AI run or human run and change format, as I imagine a less pretty format would save on tokens.

domain/ide/workspace/folder.go
```diff
@@ -683,11 +736,17 @@ func getFlatIssueList(issueByFile snyk.IssuesByFile) []types.Issue {
...
+// hiding every finding behind an IsNew flag that was never computed.
+func (f *Folder) filterByIsNew(issues snyk.IssuesByFile) snyk.IssuesByFile {
 	filtered := snyk.IssuesByFile{}
 	for path, issueSlice := range issues {
 		for _, issue := range issueSlice {
-			if issue != nil && issue.GetIsNew() {
+			if issue == nil {
+				continue
+			}
+			if issue.GetIsNew() || !f.IsBaselineAvailable(issue.GetProduct()) {
```
rrama, line 749: I disagree with the fundamental premise of this PR. As a developer, if my baseline scan failed, I do not want it to show me any issues, I want to see the scan error. If the baseline scan just hasn't run yet, I want to see some placeholder text saying "Reference scan not yet finished...".
I do not want to be told I introduced vulns.

rrama: Spoke with Basti, I have come around. We need to by default surface these as they are security issues. We can't rely on user's having PR checks, etc. AC especially needs something rather than giving an all-clear. The error should still be surfaced in the tree view in the Net New tab, should the user look. And maybe for AC we do it as a notification of the failure. Plus, the issues aren't tagged as new, so won't get the icon.

application/server/configuration.go
```diff
@@ -1358,6 +1362,145 @@ func applyCliReleaseChannel(conf configuration.Configuration, settings map[strin
...
+// Remy invocation holds the read lock queues at most one goroutine rather
+// than one per save.
+var llmEnvReconcilePending atomic.Bool //nolint:gochecknoglobals // coalesces concurrent reconciliation requests
+
+// osSetenv and osUnsetenv are indirections over os.Setenv/os.Unsetenv so tests
+// can exercise applyLlmProviderEnvLocked's failure handling.
+var (
+	osSetenv   = os.Setenv   //nolint:gochecknoglobals // test indirection over os.Setenv
+	osUnsetenv = os.Unsetenv //nolint:gochecknoglobals // test indirection over os.Unsetenv
+)
```
rrama: I wonder if all these LLM bits and the funcs below should be its own struct, even in its own file, then we just wrap it all (the functions and struct creation) in `config`. I wonder if this would allow for easier testing.

For the bot's reading, AI's take on what I'm asking to provide some more clarity:
> I wonder if all these LLM bits and the funcs below should be their own struct, in its own file, with os.Setenv/os.Unsetenv as injected fields — then configuration just holds one long-lived instance (it has to be a singleton, since appliedLlmEnvVar tracks real process-global env state) and wraps it. That'd drop the four //nolint:gochecknoglobals here and let the failure-injection tests run in parallel instead of swapping package globals.

infrastructure/configuration/template/config.html
```diff
@@ -161,6 +161,42 @@ <h2>CLI configuration</h2>
...
+                                    <option value="anthropic" {{if eq (index .Settings "llm_provider") "anthropic"}}selected{{end}}>Anthropic</option>
+                                    <option value="openai" {{if eq (index .Settings "llm_provider") "openai"}}selected{{end}}>OpenAI</option>
+                                    <option value="ollama" {{if eq (index .Settings "llm_provider") "ollama"}}selected{{end}}>Ollama</option>
+                                    <option value="vertex" {{if eq (index .Settings "llm_provider") "vertex"}}selected{{end}}>Vertex AI</option>
+                                    <option value="litellm" {{if eq (index .Settings "llm_provider") "litellm"}}selected{{end}}>LiteLLM</option>
+                                </select>
+                            </div>
+
+                            <div class="form-group half-width">
+                                <label for="llm_model"><span data-toggle="tooltip" title="Model id. Required for Ollama and LiteLLM; an optional override for the other providers.">Model</span> <span class="config-scope-slot" data-config-scope-slot="true" data-setting-key="llm_model"></span></label>
```
rrama: Personal preference, I think side-by-side looks nicer.

Current:
Image

My proposal:
Image

infrastructure/configuration/template/config.html
```diff
@@ -492,6 +528,20 @@ <h3>Scanners</h3>
...
+                                        <span data-toggle="tooltip" title="Choose whether Ambient Canary may apply fixes automatically for this folder, or only notify you.">Ambient Canary autonomy</span>
+                                        {{sourceIndicator $folder.EffectiveConfig "ambient_canary_autonomy"}}
+                                    </label>
+                                    {{$aca := getEffectiveValue $folder.EffectiveConfig "ambient_canary_autonomy"}}
+                                    <select id="folder_{{$index}}_ambient_canary_autonomy" name="folder_{{$index}}_ambient_canary_autonomy" data-setting="ambient_canary_autonomy" {{if isLocked $folder.EffectiveConfig "ambient_canary_autonomy"}}disabled{{end}}>
+                                        <option value="notify_only" {{if and $aca (eq $aca.Value "notify_only")}}selected{{end}}>Notify only</option>
+                                        <option value="autonomous_fixes" {{if and $aca (eq $aca.Value "autonomous_fixes")}}selected{{end}}>Allow autonomous fixes</option>
+                                    </select>
+                                </div>
+                            </div>
```
rrama, line 560: Why ONLY at the per-folder level? Why not in Project Defaults as well?

internal/types/ldx_sync_config_test.go
```diff
@@ -129,6 +129,12 @@ func testFm(t *testing.T) workflow.ConfigurationOptionsMetaData {
 	return workflow.ConfigurationOptionsFromFlagset(fs)
 }
 
+// Test_SettingAmbientCanaryAutonomy_ExactName pins the wire key string that ambient-canary reads
+// as a plain string, since it cannot import a Go constant across repos.
+func Test_SettingAmbientCanaryAutonomy_ExactName(t *testing.T) {
+	assert.Equal(t, "ambient_canary_autonomy", SettingAmbientCanaryAutonomy)
```
rrama, line 135: Hmm, I see the problem since AC is private. But doesn't AC import LS, so should it not just use the const from LS?

internal/types/lsp.go
rrama, line 1: Q: What are you hoping to use this for?

internal/types/register_configurations.go
```diff
@@ -356,6 +373,12 @@ func RegisterAllConfigurations(fs *pflag.FlagSet) {
 		configresolver.AnnotationDisplayName: {"SAST Settings"},
 		configresolver.AnnotationDescription: {"SAST configuration from Snyk API (autofix, local code engine)"},
 	})
+	registerFlag(fs, SettingAmbientCanaryAutonomy, "", "Ambient Canary autonomy for this folder", map[string][]string{
```
rrama, line 376: Hmm, the default being `""` feels wrong, but I'm guessing there is a reason?

internal/types/ldx_sync_config.go
```diff
@@ -85,6 +85,9 @@ const (
 	SettingCliReleaseChannel               = "cli_release_channel"
 	SettingOrganization                    = "organization"
 	SettingAutomaticAuthentication         = "automatic_authentication"
+	SettingLlmProvider                     = "llm_provider"
+	SettingLlmBaseUrl                      = "llm_base_url"
+	SettingLlmModel                        = "llm_model"
```
rrama, line 90: These settings are named just `"llm_*"` not `"remediation_llm_*"`.
I am assuming intentional, but just wanting to call it out.

What this means is if we add future uses for LLMs, then we would either just use the same settings OR we would end up with like `"explanation_llm_model"` and just `"llm_model"`. Or we do a Snyk Code Security and Snyk Code Quality settings, where we had `"snyk_code"` as well as `"snyk_code_security"` and `"snyk_code_quality"`, where the less specific is used as a fallback if the more specific isn't set. <- I'd imagine we'd do this approach. So, I think we are all good.

infrastructure/configuration/template/config.html
rrama, line 560: Should this only be visible for the Ambient Canary integration? Does it make sense for this to be configurable in the IDEs when the IDEs don't share a common configuration file?

features/delta-fail-open.feature
```diff
@@ -0,0 +1,52 @@
...
+  A developer who enables "only show new findings" must never see an empty
+  result just because no baseline has been recorded yet. Until a baseline
+  exists, every current finding is shown; once a baseline exists, only newly
+  introduced findings are shown.
+
+  # maps: M2
+  Scenario: A developer's first scan with delta findings enabled shows every finding
+    Given a running language server
+    And the editor sends the initialize request
+    And delta findings are enabled for the workspace
```
rrama: Nit: I'm not keen on the use of "workspace", I think it could be misinterpreted (someone may read it as meaning "workspace folder"). Can we call it "globally" instead please?

Maybe we should say "delta findings switched to "New new issues" globally" to be even clearer, as "enabled" makes it sound like you can disable doing delta scans entirely.

rrama: Of note, if we had a duplicate BDD scenario where instead of "workspace" it was "workspace folder" (set at the folder level) then we would have spotted https://snyksec.atlassian.net/browse/IDE-2493

features/delta-fail-open.feature
```diff
@@ -0,0 +1,52 @@
...
+    Then the editor is notified of only the newly introduced issue
+
+  # maps: M6
+  Scenario: A developer with a baseline for one product but not another sees new findings from both
+    Given a running language server
+    And the editor sends the initialize request
+    And delta findings are enabled for the workspace
+    And the developer has an established baseline for one product with one known issue
+    When the developer saves a file that produces a new issue in that product and an issue in a product without a baseline
+    Then the editor is notified of the newly introduced issue and of the issue from the product without a baseline
```
rrama: There are getting way too specific toward the end (the last 3) and feel like they are not really reusable.

I asked AI to comment:
> These steps feel over-specific — they read like a paraphrased test case rather than reusable Gherkin. Each line encodes the whole scenario setup/assertion in one sentence, and each maps 1:1 to a bespoke, non-parameterized step function in `application/server/bdd_steps_test.go` (`developerHasEstablishedBaseline`, `developerSavesFileWithMixedProductIssues`, `editorNotifiedOfNewAndUnbaselinedIssues`). Nothing else reuses them, so the specificity buys us no coverage.
> 
> A couple of concerns:
> - **Not reusable**: "an issue in a product without a baseline" bakes the product topology into prose instead of a parameter/table.
> - **Restates the assertion**: line 52 narrates the expected diff (new issue kept + unbaselined issue kept) rather than describing what the user observes.
> 
> Consider parameterizing product/baseline state so the steps compose across scenarios, e.g.:
> 
> ```gherkin
> Given a baseline exists for "Code"
> And no baseline exists for "Open Source"
> When a scan finds a new "Code" issue and an "Open Source" issue
> Then the editor is notified of both issues
> ```
> 
> This keeps each step small and lets other scenarios reuse `a baseline exists for "<product>"`. Not blocking — it's consistent with the rest of the delta scenarios in this file — but more granular steps would give the suite more leverage as it grows.

features/delta-fail-open.feature
```diff
@@ -0,0 +1,52 @@
+Feature: Delta findings fail open when no baseline exists
+
+  A developer who enables "only show new findings" must never see an empty
+  result just because no baseline has been recorded yet. Until a baseline
+  exists, every current finding is shown; once a baseline exists, only newly
+  introduced findings are shown.
+
+  # maps: M2
```
rrama: Should the mapping be like `IDE-####-M#`? And where are these requirements defined? Are they meant to live in the code or in the Jira or what?

features/llm-provider-setting.feature
```diff
@@ -0,0 +1,39 @@
...
+  configuration dialog. The setting exists, persists, is reflected back in
+  the dialog, and is actually used by autonomous remediation when it runs.
+
+  # maps: IDE-2274-M1, IDE-2274-M2
+  Scenario: The chosen provider and endpoint survive reopening the dialog
+    Given a running language server
+    When a developer saves "ollama" as the LLM provider with custom API endpoint "http://localhost:11434"
+    And the developer reopens the Snyk configuration dialog
+    Then the configuration dialog shows "ollama" as the selected LLM provider
+    And the configuration dialog shows "http://localhost:11434" as the custom API endpoint
```
rrama: "as the custom LLM API endpoint"* We do not want it confused with the Snyk endpoint.

application/server/bdd_steps_test.go
```diff
@@ -0,0 +1,1279 @@
...
+	}
+	selectHTML := s.dialogHTML[idx : idx+selectEnd]
+	if !strings.Contains(selectHTML, `value="" selected`) {
+		return fmt.Errorf("expected no provider (the empty \"Automatic\" option) to be selected, got: %s", selectHTML)
+	}
+	return nil
+}
+
+func (s *bddSteps) theDialogContainsNoLlmApiKeyField() error {
+	if strings.Contains(strings.ToLower(s.dialogHTML), "llm_api_key") {
```
rrama, line 521: Asserting a negative is a bit silly, especially since it is hardcoded (not a real setting in the first place).

application/server/configuration_test.go
```diff
@@ -3707,3 +3709,585 @@ func Test_ProcessConfigSettings_GlobalReset(t *testing.T) {
...
+	assert.Equal(t, "llama3.1", m[types.SettingLlmModel])
+
+	renderer, err := infraconfig.NewConfigHtmlRenderer(engine, cr)
+	require.NoError(t, err)
+	html := renderer.GetConfigHtml(m, folderConfigs)
+	require.NotEmpty(t, html)
+	assert.Contains(t, html, "llama3.1", "the saved model must come back in the re-rendered dialog")
+}
+
+func TestApplyLlmProviderConfig_PersistsProviderAndBaseUrl(t *testing.T) {
```
rrama, line 3714: Super petty nit: Put the more unit-style tests before the closer to integration-style tests above.

application/server/configuration_test.go
```diff
@@ -3707,3 +3709,585 @@ func Test_ProcessConfigSettings_GlobalReset(t *testing.T) {
...
+		types.SettingLlmModel:    {Value: "llama3.1", Changed: true},
+	})
+
+	assert.Equal(t, "llama3.1", types.GetGlobalString(conf, types.SettingLlmModel))
+}
+
+// TestApplyLlmProviderConfig_ModelOnlySaveIsNotSwallowedByM5Guard proves the
+// M5 "nothing touched" guard checks all three fields, not just provider and
+// base URL - a save that only changes the model must not be dropped.
+func TestApplyLlmProviderConfig_ModelOnlySaveIsNotSwallowedByM5Guard(t *testing.T) {
```
rrama, line 3748: What is the M5 guard? Similar to the other file, where do these requirements live?

application/server/configuration_test.go
```diff
@@ -3707,3 +3709,585 @@ func Test_ProcessConfigSettings_GlobalReset(t *testing.T) {
...
+	assert.Equal(t, "llama3.1", types.GetGlobalString(conf, types.SettingLlmModel))
+}
+
+func TestApplyLlmProviderConfig_AbsentSettingsLeaveConfigAndEnvUntouched(t *testing.T) {
+	engine, _ := testutil.UnitTestWithEngine(t)
+	conf := engine.GetConfiguration()
+	logger := engine.GetLogger()
+
+	t.Setenv("ANTHROPIC_BASE_URL", "")
+	require.NoError(t, os.Unsetenv("ANTHROPIC_BASE_URL"))
```
rrama, line 3768: Nit, leave the classic comment of why we do this as 2 steps.

application/server/configuration_test.go
```diff
@@ -3707,3 +3709,585 @@ func Test_ProcessConfigSettings_GlobalReset(t *testing.T) {
...
+		start := time.Now()
+		applyLlmProviderConfig(conf, logger, map[string]*types.ConfigSetting{
+			types.SettingLlmProvider: {Value: "anthropic", Changed: true},
+			types.SettingLlmBaseUrl:  {Value: u, Changed: true},
+		})
+		assert.Less(t, time.Since(start), promptSaveThreshold, "each queued settings save must return promptly")
+	}
+
+	close(proceed)
+	<-fixDone
```
rrama: Nit: Consider `RequireEventuallyClosed`, to avoid hangs.

I asked AI to find others:

> Here are the other spots worth a comment, ranked by how much they matter:
> 
> **1. Second `<-fixDone` — same fix as the first**
> `application/server/configuration_test.go:4241` has the identical unguarded blocking receive in `TestApplyLlmProviderConfig_MultipleSavesDuringInFlightInvocation_ConvergeToLatestValue`. Same swap to `testsupport.RequireEventuallyClosed`.
> 
> **2. `<-monitorDone` in the concurrency test**
> `application/server/configuration_test.go:4067` — another bare blocking receive. Here it's lower-risk because `monitorDone` is closed by a goroutine whose only blocking op is a lock the writers release, and it only runs after `writers.Wait()`. Still, for consistency `RequireEventuallyClosed` avoids a hang if the monitor goroutine ever wedges.
> 
> **3. `time.Sleep(50ms)` anti-pattern**
> `application/server/configuration_test.go:4251-4253` — a fixed sleep followed by an assert to check the env "doesn't flap back". This is exactly what `testsupport.RequireNeverReceive` (`internal/testsupport/channels.go:72-82`) exists for, but that needs a channel. Since this is polling a value not a channel, the closest idiomatic tool is `require.Never` on the condition:
> ```go
> require.Never(t, func() bool {
>     return os.Getenv("ANTHROPIC_BASE_URL") != latest
> }, 50*time.Millisecond, 5*time.Millisecond, "env must not flap back to a stale queued value")
> ```
> The bare `time.Sleep` will either flake on slow CI or silently pass if a flap happens after the window — the comment on `RequireNeverReceive` calls out this exact failure mode.
> 
> **Not worth commenting on:**
> - The two `invocationStarted` `select`/`time.After` blocks (`application/server/configuration_test.go:4161-4165` and `:4224-4228`) are already timeout-guarded and safe. Swapping to `RequireEventuallyReceive` is pure consistency, optional.
> - The `require.Eventually` convergence checks (`:4182`, `:4247`, `:4079`) are already using the right pattern.
> 
> So the highest-value comments are: the second `<-fixDone` (`:4241`) and the `time.Sleep` flap check (`:4251`). The first `<-fixDone` (`:4176`) you already flagged, and `<-monitorDone` (`:4067`) is a minor consistency nit.

domain/snyk/remediation/remy_config_test.go
```diff
@@ -44,3 +46,110 @@ func TestBuildRemyFixConfig_SelectsSastAgenticFlow(t *testing.T) {
...
+	types.SetGlobalUser(base, types.SettingLlmProvider, "anthropic")
+
+	conf := buildRemyFixConfig(base, "/work/repo-root")
+
+	assert.Equal(t, "anthropic", conf.GetString("provider"))
+	assert.False(t, conf.IsSet("model"), "model must not be set when the developer never chose one")
+}
+
+// TestBuildRemyFixConfig_ModelWithoutProvider covers the (unlikely but
+// possible) case of a persisted model with no provider selected: the model
```
rrama: Nit: "unlikely but possible", this isn't really that "unlikely" IMO.

domain/snyk/remediation/remy_config_test.go
```diff
@@ -44,3 +46,110 @@ func TestBuildRemyFixConfig_SelectsSastAgenticFlow(t *testing.T) {
...
+}
+
+// TestBuildRemyFixConfig_NoProviderChosen guards the no-forced-default
+// requirement: a developer who never chose a provider must not
+// have "provider"/"model" keys set at all, matching the empty-string-key
+// discipline the other flags already follow.
+func TestBuildRemyFixConfig_NoProviderChosen(t *testing.T) {
+	conf := buildRemyFixConfig(configuration.NewWithOpts(), "/work/repo-root")
+
+	assert.False(t, conf.IsSet("provider"), "provider must not be set when the developer never chose one")
```
rrama, line 75: Nit: When doing an `assert.False(t, conf.IsSet(` I prefer we refer to the const rather than hardcode the string, as the test passes if someone changes the prod string.

rrama (approved):
Few bits that may need fixing, few things need clarifying, then some nits.

Thought (for a future enhancement): If I have the env var set for my Anthropic base URL, should we display the value of it in the settings HTML page. With maybe a note saying it is from the env var, like when we say the source for VSC is User or Workspace.

internal/types/ldx_sync_config_test.go
basti-snyk: No, it's in internal. That's why it doesn't work.

internal/types/ldx_sync_config.go
basti-snyk: Yep, it's intentional.

application/server/bdd_test.go
basti-snyk: Deferred as a future enhancement per your own framing — not implementing now, will revisit if/when it's worth the added complexity.

application/server/configuration.go
basti-snyk: Done — extracted into an llmProviderEnvManager type in application/server/llm_provider_env.go, with injectable setenv/unsetenv fields instead of swapping package-level function vars for tests, per your suggestion.

infrastructure/configuration/template/config.html
basti-snyk: Added it to Project Defaults too, mirroring the existing scan_automatic dual-scope pattern — so it's no longer per-folder only.

features/delta-fail-open.feature
basti-snyk: Parameterized the last scenario's steps by product name (developerSavesFileWithProductIssues / editorNotifiedOfBothProductIssues) instead of the bespoke non-reusable functions, per your suggestion.

features/delta-fail-open.feature
basti-snyk: Renamed to the IDE-2418-M# format, matching the convention already used by features/llm-provider-setting.feature.

application/server/configuration_test.go
basti-snyk: Both fixed: swapped the second bare <-fixDone for testsupport.RequireEventuallyClosed (mirroring the earlier fix in this file), and replaced the time.Sleep+assert with require.Never.

features/delta-fail-open.feature
basti-snyk: Reworded to "switched to ... globally", and added a new scenario (IDE-2418-M7) covering the folder-level ("workspace folder") case you flagged in your reply.

application/server/configuration_test.go
basti-snyk: Reordered — the unit-style TestApplyLlmProviderConfig_*/TestApplyLlmProviderEnvLocked_* tests now come before the integration-style TestUpdateSettings_Llm* round-trip tests.

application/server/configuration_test.go
basti-snyk: Restored the comment explaining why this is deliberately done as 2 steps.

domain/snyk/remediation/remy_config_test.go
basti-snyk: Reworded to "edge case" — fair point.

domain/snyk/remediation/remy_config_test.go
basti-snyk: Switched to remyProviderConfigKey instead of the hardcoded string.

infrastructure/configuration/template/config.html
basti-snyk: Added it to Project Defaults too, mirroring the existing scan_automatic dual-scope pattern, so it isn't per-folder-only anymore.

internal/types/register_configurations.go
basti-snyk: Confirmed intentional — it's consistent with sibling folder-scoped settings (e.g. scan_net_new) that also default to "" to distinguish "not configured" from an explicit choice. Also just added the setting to Project Defaults, so there's now an easy way to set a real value globally instead of leaving every folder unset.

application/server/configuration_test.go
basti-snyk: Added a pointer comment: see "IDE-2274-M5" in features/llm-provider-setting.feature for the mapped requirement — that's a different M5 than delta-fail-open's own M5, unrelated namespace.

application/server/bdd_steps_test.go
basti-snyk: 🙏

application/server/bdd_test.go
basti-snyk: Fixed the copyright year on bdd_test.go and bdd_lint.go, and checked every other new file in this PR — no other stale years found.

infrastructure/configuration/template/config.html
basti-snyk: Done — switched to the existing Bootstrap row/col-md-6 grid pattern already used elsewhere in this file, so the two fields sit side by side.

features/llm-provider-setting.feature
basti-snyk: Reworded to "as the custom LLM provider's API endpoint".

application/server/bdd_steps_test.go
basti-snyk: Simplified — now just checks for any selected option instead of the hardcoded empty-value negative assertion.

PR #1405, opened 2026-08-07
feat(server): expose structured Code AI Fix result via $/snyk.aiFix [IDE-2451]

basti-snyk:
## Summary
- Adds `types.AiFixNotification` / `types.AiFixResult` DTOs and a new `$/snyk.aiFix` case in `registerNotifier` (`application/server/notification.go`), following the exact existing pattern used for `$/snyk.scan`, `$/snyk.scanSummary`, and `$/snyk.treeView`.
- Emits the notification from the existing `setStateCallback` seam in `domain/ide/command/code_fix_diffs.go`, alongside (not instead of) the existing `SendShowDocumentRequest` call, whenever the real `AiFixHandler` state machine transitions (`AiFixInProgress` / `AiFixSuccess` / `AiFixError`).
- Payload is `{issueId, status, fixes: [{fixId, filePath}]}`, built from `AiFixHandler`'s real `AiStatus` and `[]llm.AutofixUnifiedDiffSuggestion` results via a new pure `aiFixResultsFrom` helper (one `AiFixResult` per file per suggestion, sorted for determinism).
- Purely additive: `fixDiffs`'s fire-and-forget contract, the `showDocument` request/webview behavior, and all other `$/snyk.*` wire formats are unchanged.

## Test plan
- [x] Integration test on the real notifier plane (`application/server/notification_test.go`, `Test_AiFixNotification`) — real `di.Notifier()`, real `registerNotifier`, real JSON-RPC harness; confirmed RED before the `registerNotifier` case existed.
- [x] Integration test through the real production entrypoint (`domain/ide/command/code_fix_diffs_test.go`, `Test_codeFixDiffs_Execute_SendsAiFixNotification`) — real `Execute()` → `handleResponse()` → real `AiFixHandler` + real `Notifier`, driven to the deterministic network-free ERROR path (no bundle hash registered); confirmed RED before the wiring existed.
- [x] Unit tests for the pure `aiFixResultsFrom` payload helper (no suggestions, single file, multiple suggestions spanning multiple files) — `Test_aiFixResultsFrom`.
- [x] `go build ./...` and `go vet ./...` clean (whole tree)
- [x] `GOOS=windows go build ./...` clean
- [x] `go test -race ./...` clean (full repo, no regressions)
- [x] `make lint` (golangci-lint) — 0 issues
- [x] Snyk Code and Snyk SCA scans — clean (no new issues, no new vulnerable dependencies)

### DIFF
--- a/application/server/bdd_lint_test.go
+++ b/application/server/bdd_lint_test.go
@@ -11,7 +11,7 @@ import (
 
 // trackedFeatureRequirements is a manually-maintained ratchet: extend it alongside
 // each new feature file so it can't silently drift from real coverage.
-var trackedFeatureRequirements = []string{"M1", "M2", "IDE-2418-M2", "IDE-2418-M3", "IDE-2418-M4", "IDE-2418-M5", "IDE-2418-M6", "IDE-2418-M7", "IDE-2274-M1", "IDE-2274-M2", "IDE-2274-M4", "IDE-2274-M5"}
+var trackedFeatureRequirements = []string{"M1", "M2", "IDE-2418-M2", "IDE-2418-M3", "IDE-2418-M4", "IDE-2418-M5", "IDE-2418-M6", "IDE-2418-M7", "IDE-2274-M1", "IDE-2274-M2", "IDE-2274-M4", "IDE-2274-M5", "IDE-2451-M1"}
 
 func Test_FeatureFiles_MapEveryRequirement(t *testing.T) {
 	scenarios, err := parseFeatureDir("../../features")
--- a/application/server/bdd_steps_test.go
+++ b/application/server/bdd_steps_test.go
@@ -35,6 +35,9 @@ import (
 	"github.com/snyk/snyk-ls/infrastructure/authentication"
 	"github.com/snyk/snyk-ls/infrastructure/code"
 	"github.com/snyk/snyk-ls/infrastructure/featureflag"
+	"github.com/snyk/snyk-ls/infrastructure/snyk_api"
+	"github.com/snyk/snyk-ls/internal/observability/error_reporting"
+	"github.com/snyk/snyk-ls/internal/observability/performance"
 	"github.com/snyk/snyk-ls/internal/product"
 	"github.com/snyk/snyk-ls/internal/testsupport"
 	"github.com/snyk/snyk-ls/internal/testutil"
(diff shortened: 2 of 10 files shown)


nick-y-snyk (changes requested):
Reviewed the identifier used in the new `$/snyk.aiFix` notification against the rest of this exact code path.

**Blocking — inconsistent identifier.** `code_fix_diffs.go` line ~107 sets `IssueId: issue.GetID()`, which returns the static rule ID. Everything else in this same flow keys by the occurrence-unique `AdditionalData.GetKey()`:
- `Folder.Issue(key)` (domain/ide/workspace/folder.go:100) resolves issues by this key
- `AiFixHandler.resetAiFixCacheIfDifferent` tracks state by this key
- `SendShowDocumentRequest`, fired from the *same* `setStateCallback` a few lines above, builds its identifier via the `issueId()` helper (infrastructure/code/issue_enhancer.go:213), which prefers this key

When a file has two findings for the same rule, both get the same `IssueId` here, so a consumer (this unblocks ambient-canary's IDE-2448 acceptance test per the Jira ticket) can't tell which occurrence a given `$/snyk.aiFix` notification belongs to. Suggest using the same `issueId()` helper for consistency.

**Minor — partial determinism.** `aiFixResultsFrom` sorts file paths within a suggestion but not the suggestions slice itself, so `fixes` ordering isn't fully deterministic if the scanner returns suggestions in varying order.

**Test coverage note.** `Test_codeFixDiffs_Execute_SendsAiFixNotification`'s fixture (`snyk.Issue{ID: uuid.NewString()}`) never sets `AdditionalData`, so `GetKey()` falls back to `GetID()` — the assertion would pass regardless of which identifier the code reads. A fixture with a `Key` distinct from `ID` would have caught the bug above.

nick-y-snyk (approved):
Approving.

All three points from my earlier review are addressed:

- **Identifier** — `sendAiFixNotification` now uses `code.IssueId(issue)` (`issueId` exported as `IssueId` in `infrastructure/code/issue_enhancer.go`), so `$/snyk.aiFix` carries the occurrence-unique key rather than the static rule ID, consistent with `Folder.Issue`, `resetAiFixCacheIfDifferent` and `SendShowDocumentRequest`.
- **Determinism** — `aiFixResultsFrom` now sorts suggestions by `FixId` as well as file paths within a suggestion.
- **Test coverage** — `Test_codeFixDiffs_Execute_SendsAiFixNotification` now uses a fixture with `AdditionalData.Key` distinct from `ID` and asserts on the key, so it actually exercises the identifier.

Two follow-ups handled on top of that:

- `c316e9f9` fixes a data race the determinism fix introduced: `aiFixResultsFrom` sorted the slice returned by `GetAiFixDiffResult()` in place, but that is `AiFixHandler`s internal `aiFixDiffState.result`, not a copy — so it mutated shared state with no lock held while `AiFixHandler.GetResults` iterates the same backing array under `RLock` for the webview. Now sorts a copy via `slices.SortedFunc`; emitted ordering is unchanged.
- Branch history was rebuilt so every commit is authored and committed by `basti-snyk`. The original root commit had an empty author email (`Bastian Doetsch <>`), unlinked to any GitHub account, which was the sole reason the CLA check failed. `license/cla` is green now. The rewritten tree was verified byte-identical to the pre-rewrite tip before the race fix was added.

PR #1408, opened 2026-08-13
refactor: read consistent ignores from the folder config the scan holds

rrama:
### Description

A Snyk Code scan resolved the consistent ignores feature flag by folder path while already holding that folder's config. Resolving by path builds a second folder config: a go-git enrichment of the repository and a write back to storage, for a value the config in hand already answers.

`Scan` cannot reach this point without a folder config — `ResolveFolderAndScanType` fails first with `ErrFolderConfigNotInContext`, and the folder path it returns comes from that same config — so both forms read the same key on the same configuration, and the flag value is unchanged.

`Test_Scan` listed the expected upload path per case in its table and never asserted it, so forcing the flag either way left the suite green. The fake code scanner now records which of the two upload paths ran and the cases assert it, which is what pins the flag to the folder config it is staged on. Mutation-checked: forcing the flag on fails the legacy case, forcing it off fails the consistent ignores case.

Independent of the CLI-1721 gitignore filtering stack, which makes the same move for the two file filter flags.

### Checklist

### DIFF
--- a/infrastructure/code/code.go
+++ b/infrastructure/code/code.go
@@ -291,7 +291,7 @@ func internalScan(ctx context.Context, sc *Scanner, folderPath types.FilePath, l
 		return []types.Issue{}, nil
 	}
 
-	codeConsistentIgnoresEnabled := sc.featureFlagService.GetFromFolderConfig(folderPath, featureflag.SnykCodeConsistentIgnores)
+	codeConsistentIgnoresEnabled := folderConfig.GetFeatureFlag(featureflag.SnykCodeConsistentIgnores)
 	results, err = sc.UploadAndAnalyze(ctx, folderPath, folderConfig, files, filesToBeScanned, codeConsistentIgnoresEnabled, t)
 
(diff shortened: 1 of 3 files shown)


rrama:
Thanks bot — checking the nil dereference finding: `FolderConfig.GetFeatureFlag` starts with `if fc == nil { return false }` ([internal/types/folder_config.go](https://github.com/snyk/snyk-ls/blob/main/internal/types/folder_config.go#L127)), so a nil folder config cannot panic here.

The reachable-but-different behaviour would be a silent `false` rather than a crash, and that is also unreachable: `Scan` calls `scannercommon.ResolveFolderAndScanType`, which returns `ErrFolderConfigNotInContext` before `internalScan` runs, and derives the folder path from that same config. So the by-path lookup this replaces and the folder-config read resolve the same key on the same configuration.

No change made.

andrewrobinsonhodges-snyk (approved):

PR #1409, opened 2026-08-13
refactor: resolve GAF config keys from a list

rrama:
## Stack Context

Two PRs delivering CLI-1721, tracked-file-aware `.gitignore` filtering for Snyk Code:

1. **this PR** — the feature flag service resolves GAF configuration keys from a list rather than one hardcoded special case
2. [feat: filter files with GAF's configuration-driven file filter](https://github.com/snyk/snyk-ls/pull/1410) — the filtering itself, built on that

### Description

The service resolved exactly one GAF configuration key, and did so through a name comparison in the middle of the per-flag fetch loop. Adding a second key meant adding a second branch.

Flags whose value comes from a GAF configuration key now live in `gafConfigResolvedFlags` and are resolved together by `getConfigValues`, separate from the flags the feature flag service resolves by platform name. The keys share one configuration clone: GAF caches what it resolves on the configuration object it is handed, including one batched flag evaluation covering the keys it registers together, so a clone per key would throw that cache away. A key that cannot be resolved is reported as `false` alongside a joined error, so a failed lookup can never be mistaken for an enabled flag.

### Why a configuration key rather than a flag name

GAF owns the mapping from its configuration keys to platform feature flag names, and those names are unexported string literals inside GAF. A copy of one in snyk-ls is unverifiable: no compiler error, no test failure and no runtime error if GAF renames it — the flag simply reads `false` forever and the rollout reaches nobody. Naming the configuration key keeps that second copy out of this repository. `Test_fetch_resolvesConfigKeysAsConfigKeys` pins the route each key takes, since taking the wrong one is silent.

### No behaviour change

`ignore_workflow.ConfigIgnoreApprovalEnabled` is `"internal_iaw_enabled"` in both GAF v0.11.0 and v0.14.1 — the same string as the `IgnoreApprovalEnabled` constant it replaces — so the key written to folder config is unchanged. `IgnoreApprovalEnabled` and the exported `Flags` slice have no consumers outside this package.

### Checklist

### DIFF
--- a/infrastructure/featureflag/featureflag.go
+++ b/infrastructure/featureflag/featureflag.go
@@ -17,6 +17,7 @@
 package featureflag
 
 import (
+	"errors"
 	"fmt"
 	"maps"
 	"sync"
@@ -24,6 +25,7 @@ import (
 
 	"github.com/erni27/imcache"
 	"github.com/rs/zerolog"
+
 	"github.com/snyk/code-client-go/pkg/code"
 	"github.com/snyk/code-client-go/pkg/code/sast_contract"
 	"github.com/snyk/go-application-framework/pkg/configuration"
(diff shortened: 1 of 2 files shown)


rrama:
* **#1410** <a href="https://app.graphite.com/github/pr/snyk/snyk-ls/1410?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1409** <a href="https://app.graphite.com/github/pr/snyk/snyk-ls/1409?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snyk/snyk-ls/1409?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

rrama:
Thanks bot — two notes on this one.

**Interface Contract Change:** agreed the shape changed, but it was never externally implementable: `ExternalCallsProvider` was an exported type whose methods (`getConfigValues`, `getFeatureFlag`, `getSastSettings`) are all unexported, so nothing outside this package could satisfy it. Rather than argue the point, the interface is now unexported too — `externalCalls`, with `WithProvider` → `withProvider` to match, since it is only reachable from this package's own tests. Done in 5b417410.

**Unnecessary Mutex Contention:** fair, the debug logging is now outside the lock, matching the platform flag path below it. Done in 5b417410.

rrama:
On the renaming finding: nothing outside this package can implement the interface, before or after — every method on it (`getConfigValues`, `getFeatureFlag`, `getSastSettings`) is unexported, so an external implementation was never possible and unexporting the type cannot break one. `go build ./...` is clean across the repo. Same rationale as [the earlier reply](https://github.com/snyk/snyk-ls/pull/1409#issuecomment-5281619584).

infrastructure/featureflag/featureflag.go
```diff
@@ -38,30 +40,37 @@ import (
...
 	UseExperimentalRiskScoreInCLI,
 	UseExperimentalRiskScore,
 	UseOsTest,
 	SnykSecretsEnabled,
 }
 
+// gafConfigResolvedFlags are GAF configuration keys rather than platform feature flag names. GAF
+// registers a resolver for each and owns whatever answering it takes — an org settings lookup for
+// ignore approval, a batched flag evaluation for others — so snyk-ls holds no second copy of a name
+// or endpoint that could drift out of step with it.
```
andrewrobinsonhodges-snyk: Nit: It took me three read-throughs of that before I could parse it. Granted it is 27 degrees in my flat right now, but I feel this could be worded better.

andrewrobinsonhodges-snyk (approved):

infrastructure/featureflag/featureflag.go
rrama: Reworded — thanks, that one was doing three jobs in a single sentence. Now:

```go
// gafConfigResolvedFlags are GAF configuration keys, not the feature flag names themselves.
// Using the GAF configuration key ensures we use the same resolution logic as GAF, plus keeps LS
// free of a second copy of the flag name that could drift. GAF decides how each key is answered and
// we don't need to worry about the implementation details.
```

Done in 587ea6de. The stack is also rebased onto main now that #1408 has merged.

rrama:
On the **Logic Error** finding: this one is already safe, and the test cited as evidence proves it rather than exposing it.

`GetBoolWithError` normalises the value before we ever see it — [`pkg/configuration/configuration.go:531`](https://github.com/snyk/go-application-framework/blob/v0.14.1/pkg/configuration/configuration.go#L531) in GAF v0.14.1:

```go
func (ev *extendedViper) GetBoolWithError(key string) (bool, error) {
	result, err := ev.GetWithError(key)
	if err != nil || result == nil {
		return false, err          // value discarded whenever there is an error
	}
	return toBool(result)
}
```

So when the resolver in `Test_getConfigValues_reportsErrorsWithoutLosingValues` returns `(true, error)`, the `true` never reaches `getConfigValues` — `value` is already `false`. That is exactly what the test asserts, and it passes:

```go
assert.False(t, values[testConfigKeyA], "a failed lookup must be false")
assert.True(t, values[testConfigKeyB], "one failing key must not discard the others")
```

Worth recording the half of this that is fair: the fail-closed guarantee comes from GAF, not from our own code. We are deliberately not adding a defensive `if err != nil { value = false }` here, because with GAF normalising upstream that branch is unreachable and no test can exercise it — a line that looks load-bearing but never runs is worse than the inherited guarantee, which at least has a test pinning the observable behaviour.

The **Breaking Change** finding is the same one answered [here](https://github.com/snyk/snyk-ls/pull/1409#issuecomment-5281619584) and [here](https://github.com/snyk/snyk-ls/pull/1409#issuecomment-5281902618): every method on that interface is unexported, so no package outside `featureflag` could ever have implemented it, and `go build ./...` is clean repo-wide.

No change made.

PR #1410, opened 2026-08-13
feat: filter files with GAF's configuration-driven file filter [CLI-1721]

rrama:
## Stack Context

Second of two PRs delivering CLI-1721, tracked-file-aware `.gitignore` filtering for Snyk Code. Stacked on [refactor: resolve GAF config keys from a list](https://github.com/snyk/snyk-ls/pull/1409), which is what lets this PR name two GAF configuration keys instead of hardcoding platform flag names. **Review that one first; this PR's base is that branch.**

### Description

Snyk Code scans skipped files that `.gitignore` lists but git tracks, because the file filter treated every ignore rule as authoritative regardless of what is in the index. GAF's file filter fixes this behind `FF_GITIGNORE_RESPECT_TRACKED_FILES`.

Scanning now dispatches per folder. With the flag on, the filter is built from a configuration clone carrying the folder's organization and both file filter flag values, so GAF resolves nothing over the network mid-scan. With the flag off, the previous implementation runs unchanged — it is the control arm of the rollout and the rollback target, and `legacyFilteredFiles` is the shipped body verbatim apart from the default globs moving to a shared `lsDefaultIgnoredRules`.

Both flags and the organization are read from the folder config the scan already carries. Resolving any of them by path builds a fresh folder config, which opens the repository with go-git to enrich it and writes it back to storage, so a scan would otherwise repeat that work per flag — on the legacy path too, where the flags cannot change what the filter does. A missing folder config is refused rather than defaulted, because reading its flags as off would filter the folder as though the rollout had never reached it.

### Per-folder resolution

snyk-ls is multi-root and a folder can belong to an organization other than the global one. The engine configuration carries only the global organization, so both flags are resolved per folder and the filter's configuration clone is scoped to that folder's organization before GAF sees it. `Test_filteredFiles_resolvesFlagPerFolder` fails if resolution ever regresses to the global engine configuration.

### Dependency bump

GAF `v0.11.0` → `v0.14.1` and code-client-go `v1.26.1` → `v1.31.3`. Note that main's `chore: correct GAF version to v0.11.0` (#1395) corrected an accidental `v1.0.0`; `v0.14.1` is the first release whose file filter actually reads `FF_GITIGNORE_RESPECT_TRACKED_FILES`.

### Testing

`code_file_filter_test.go` covers the dispatcher over every combination of the two flags, per-folder resolution, the metacharacter path shape that breaks `.gitignore` matching entirely, scan roots below a repository root, caching behaviour on each arm, error paths, and one row per `lsDefaultIgnoredRules` pattern. Mutation-checked: forcing the dispatcher either way, unplumbing the metacharacter flag, dropping the configuration clone, and resolving the organization globally each turn the suite red.

### Checklist

- [x] License file updated, if new 3rd-party dependency is introduced (dependency bump only, no new direct dependency)


rrama:
* **#1410** <a href="https://app.graphite.com/github/pr/snyk/snyk-ls/1410?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snyk/snyk-ls/1410?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1409** <a href="https://app.graphite.com/github/pr/snyk/snyk-ls/1409?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

rrama:
Thanks bot — both findings are intentional, flagging why for reviewers.

**Breaking Behavioral Change:** refusing on a missing folder config is deliberate. Defaulting reads every feature flag as `false`, which would filter the folder as though the rollout had never reached it and leave nothing behind to explain why — a silent wrong answer is worse here than a loud failure. It is also unreachable in practice: `Scan` fails with `ErrFolderConfigNotInContext` via `scannercommon.ResolveFolderAndScanType` before `internalScan` is called.

**Stale Cache Risk:** the caching difference between the two paths is the point, not an oversight. `legacyFilteredFiles` is the control arm of a live A/B rollout and the rollback target, so it stays byte-for-byte as shipped; `newFilteredFiles` deliberately caches nothing so a flag flip takes effect on the next scan. `Test_filteredFiles_reReadsTrackedFilesFlagPerCall` pins exactly that difference.

No change made.

rrama:
Thanks bot — both of these need correcting for anyone reviewing.

**Performance Regression:** the premise does not hold. GAF v0.14.1's `FileFilter` caches nothing — its fields are `path`, `defaultRules`, `logger`, `max_threads`, `dotSnykSections`, `config`, `metricsRecorder`, `metricScopeID` — and `GetAllFiles()` runs a fresh `filepath.WalkDir` on every call. `legacyFilteredFiles` calls `GetAllFiles()` once per scan as well, so both paths walk the tree on every scan and always have. What `sc.fileFilters` reuses is the struct, not the IO.

Not caching on the new path is deliberate and tested: a folder scanned before the rollout reached it must pick up the flag on its next scan rather than keep serving a filter built with the old flag values. `Test_filteredFiles_reReadsTrackedFilesFlagPerCall` pins that.

**Increased Failure Risk:** the folder config is not injected into the context by `Scan` — it is read *out* of it. `scannercommon.ResolveFolderAndScanType(ctx)` returns `ErrFolderConfigNotInContext` when the context has none, and `Scan` aborts there, before `internalScan`; the folder path it scans comes from that same config. So the nil branch is unreachable, and refusing rather than defaulting is the deliberate choice described in the comment on it.

No change made.

infrastructure/code/code.go
```diff
@@ -297,6 +287,82 @@ func internalScan(ctx context.Context, sc *Scanner, folderPath types.FilePath, l
 	return results, err
 }
 
+// lsDefaultIgnoredRules is prepended to every rule set. Both call sites append to it, which allocates
+// a new slice rather than writing into this one only because a composite literal has no spare
+// capacity, so it must never be rebuilt with room to grow.
+var lsDefaultIgnoredRules = []string{"**/.git/**", "**/.svn/**", "**/.hg/**", "**/.bzr/**", "**/.DS_Store/**"} //nolint:gochecknoglobals // a fixed rule set Go cannot express as a const
```
andrewrobinsonhodges-snyk, line 291: Might be worth a comment saying what these rules represent? Or define them as string constants with sensible names. 

Just a suggestion though - up to you whether to action it.

infrastructure/code/code.go
```diff
@@ -297,6 +287,82 @@ func internalScan(ctx context.Context, sc *Scanner, folderPath types.FilePath, l
...
+
+// filteredFiles picks the file filtering implementation for the folder being scanned. Flags and
+// organization both come from the folder config the scan already holds, because resolving either by
+// path builds a fresh folder config, which opens the repository with go-git to enrich it and writes
+// it back to storage — work a scan would otherwise repeat per flag, on the legacy path too.
```
andrewrobinsonhodges-snyk: "legacy path?" Please either remove or explain.

infrastructure/code/code.go
```diff
@@ -297,6 +287,82 @@ func internalScan(ctx context.Context, sc *Scanner, folderPath types.FilePath, l
...
+
+	if folderConfig.GetFeatureFlag(gafUtils.FF_GITIGNORE_RESPECT_TRACKED_FILES) {
+		return sc.newFilteredFiles(folderConfig, logger)
+	}
+
+	return sc.legacyFilteredFiles(folderConfig.FolderPath, logger)
+}
+
+// newFilteredFiles is the new CLI-1721 file filtering, for those with the tracked files rollout flag.
+// Done as a bodge copy of
```
andrewrobinsonhodges-snyk: If this is a bodge, please add a TODO for fixup.

infrastructure/code/code.go
```diff
@@ -297,6 +287,82 @@ func internalScan(ctx context.Context, sc *Scanner, folderPath types.FilePath, l
...
+	conf.Set(configuration.ORGANIZATION,
+		config.FolderOrganizationFromConfig(folderConf, folderConfig.FolderPath, &logger))
+	conf.Set(gafUtils.FF_GITIGNORE_RESPECT_TRACKED_FILES, true)
+	conf.Set(gafUtils.FF_FILE_FILTER_METACHARACTER_FIX,
+		folderConfig.GetFeatureFlag(gafUtils.FF_FILE_FILTER_METACHARACTER_FIX))
+
+	fileFilter := gafUtils.NewFileFilter(string(folderConfig.FolderPath), &logger,
+		gafUtils.WithConfig(conf),
+		gafUtils.WithMetrics(sc.engine.GetAnalytics()),
+		gafUtils.WithThreadNumber(conf.GetInt(configuration.MAX_THREADS)),
```
andrewrobinsonhodges-snyk, line 337: As an aside - is WithThreadNumber the number of threads it will use, or the number of THE thread it will use? Naming could be better here.

infrastructure/code/code.go
```diff
@@ -297,6 +287,82 @@ func internalScan(ctx context.Context, sc *Scanner, folderPath types.FilePath, l
...
+	)
+
+	rules, err := fileFilter.GetRules([]string{".gitignore", ".dcignore", ".snyk"})
```
andrewrobinsonhodges-snyk, line 342: Again, might be worth separating these out into vars. Feels like the kind of magic string that might end up getting duplicated across the codebase.

infrastructure/code/code.go
```diff
@@ -297,6 +287,82 @@ func internalScan(ctx context.Context, sc *Scanner, folderPath types.FilePath, l
...
+
+// legacyFilteredFiles is the pre-CLI-1721 file filtering, for those without the GitIgnoreRespectTrackedFiles FF.
+func (sc *Scanner) legacyFilteredFiles(folderPath types.FilePath, logger zerolog.Logger) (chan string, error) {
+	fileFilter, _ := sc.fileFilters.Load(string(folderPath))
+	if fileFilter == nil {
+		fileFilter = gafUtils.NewFileFilter(string(folderPath), &logger)
+		sc.fileFilters.Store(string(folderPath), fileFilter)
+	}
+
+	rules, err := fileFilter.GetRules([]string{".gitignore", ".dcignore", ".snyk"})
```
andrewrobinsonhodges-snyk, line 360: ...like here, for example.

infrastructure/featureflag/featureflag_test.go
```diff
@@ -368,6 +369,14 @@ func Test_fetch_resolvesConfigKeysAsConfigKeys(t *testing.T) {
 	}
 }
 
+// Test_gafConfigResolvedFlags_coversFileFilterKeys pins the membership the file filtering depends on.
+// A key dropped from the list is not a compile error and not a runtime error: it reads false forever,
+// so the rollout reaches nobody and the metacharacter fix reaches nobody.
```
andrewrobinsonhodges-snyk: Oddly wordy, but fine I guess. "Missing keys evaluate to false" would have sufficed.

andrewrobinsonhodges-snyk (approved):

infrastructure/code/code.go
rrama: Added what they are: version control internals and OS metadata, kept out of every scan whichever filtering path runs. I left them as one list rather than five named constants — they are single-use literals, and `Test_lsDefaultIgnoredRules_excludesEachDocumentedPattern` already has a row per pattern so a regression names the one that broke. Done in 5485ba54.

infrastructure/code/code.go
rrama: Explained — it now names the function instead of saying "the legacy path": *"…work a scan would otherwise repeat per flag, including scans that end up in legacyFilteredFiles, where neither flag can change what the filter does."* Done in 5485ba54.

infrastructure/code/code.go
rrama: Raised IDE-2475 for it and added the TODO:

```go
// TODO(IDE-2475): delete this once the Snyk Code scanner runs the code-client-go workflow, which does
// this filtering itself.
```

The ticket is the real fix: the CLI already runs the CCG Code workflow, which does this filtering itself, and snyk-ls already registers that workflow (`code.Init(engine)` in `application/config/config.go`) — it just never invokes it, building its own `CodeScanner` instead. Adopting it deletes this function outright, so the TODO points at work that lands in this exact code.

I also dropped the word "bodge" and said plainly what it copies and why. Done in 5485ba54.

infrastructure/code/code.go
rrama: It is the number of threads — GAF's field is `max_threads` (`pkg/utils/file_filter.go`), and we pass `configuration.MAX_THREADS`. Agreed the name reads ambiguously, but it is GAF's API rather than ours, so renaming it belongs upstream. No change here.

infrastructure/code/code.go
rrama: Kept them duplicated, but said why in both places rather than leaving it to be guessed.

New path:
```go
// Spelled out rather than shared with legacyFilteredFiles so this stays a line-for-line mirror of
// the code-client-go function above, which is what makes a future diff against that source honest.
```

Legacy path:
```go
// Locked: this is the control arm of a live rollout, so changing which rule files it reads would
// need a feature flag and a rollout of its own.
```

Sharing a var across the two would mean editing the control arm, which is the rollback target for the rollout and has to stay as shipped. The duplication collapses on its own once the flag finishes rolling out and `legacyFilteredFiles` is deleted.

Worth knowing where this actually belongs: that list is hardcoded in five places across four repos — twice here, twice in code-client-go, twice in cli-extension-os-flows — and cli-extension-secrets disagrees with all of them (`.gitignore` only, no `.dcignore` or `.snyk`). GAF, which owns `FileFilter`, does not define it at all; `GetRules` takes it as a parameter. An exported default in GAF would have caught that divergence. Noted on IDE-2475. Done in 5485ba54.

infrastructure/code/code.go
rrama: See the reply above — both call sites now carry a comment explaining why they are not shared. Done in 5485ba54.

infrastructure/featureflag/featureflag_test.go
rrama: Trimmed to roughly your version:

```go
// Test_gafConfigResolvedFlags_coversFileFilterKeys pins the membership file filtering depends on: a key
// dropped from this list is not a compile or runtime error, it just reads false forever.
```

Done in 5485ba54.

infrastructure/code/code.go
rrama: Sorry, AI comment and change... It didn't find the skill so made up the rules. The new AI comment is crap, but I am going to be lazy and leave it.

infrastructure/featureflag/featureflag_test.go
rrama: Again, crap rogue AI comment, but I am going to be lazy and leave it if that's alright.

infrastructure/code/code.go
rrama: Trimmed further on reflection — it now just says what the rules are:

```go
// lsDefaultIgnoredRules keeps version control internals and OS metadata out of every scan.
```

The slice-capacity note that followed it is gone. Done in ddcd4d2a.

infrastructure/code/code.go
rrama: Cut this down again — the doc comment now just says what the function does, with the rationale for reading flags off the folder config left to the commit that introduced it:

```go
// filteredFiles picks the file filtering implementation for the folder being scanned based on a FF:
// newFilteredFiles when it is set for the folder's organization, legacyFilteredFiles when it is not.
```

Done in ddcd4d2a.

PR #1412, opened 2026-08-18
refactor: log delta pre-filter before/after counts in FilterIssues [IDE-2477]

rrama:
### Description

`filterIssuesWithConfig` applies a delta pre-filter (`filterByIsNew`) ahead of its reason-counted filter loop, so an issue set emptied there produced no distinguishing log line — a debug log could show "No issues were filtered out" while diagnostics came out empty, with no way to tell why.

Folds the drop count into the existing `filterReasonCounts` mechanism via a new `FilterReasonNotNetNew` reason, computed only when delta filtering is enabled. `filterByIsNew` now classifies and counts dropped issues in the same pass it filters them (returning `(snyk.IssuesByFile, int)`) rather than diffing two separate before/after counts; nil entries are skipped without being counted as a drop. `GetDelta`, `filterByIsNew`'s other caller, discards the count. Both `filterReasonCounts` log lines now also carry a `deltaEnabled` field.

Also adds direct coverage for `filterByIsNew`'s filtered/dropped return values (including the nil-skip case), and for the previously-untested direction of `FilterIssues` where a non-new issue passes through when delta filtering is disabled.

### Checklist

- [ ] README.md updated, if user-facing (not user-facing)
- [ ] License file updated, if new 3rd-party dependency is introduced (no new dependencies)

### DIFF
--- a/domain/ide/workspace/folder.go
+++ b/domain/ide/workspace/folder.go
@@ -792,7 +792,7 @@ func (f *Folder) filterIssuesWithConfig(
 
 	if len(filterReasonCounts) > 0 {
 		logger.Debug().
-			Interface("filterReasons", filterReasonCounts).
+			Any("filterReasons", filterReasonCounts).
 			Bool("deltaEnabled", deltaEnabled).
 			Msgf("%d issue(s) filtered", lo.Sum(lo.Values(filterReasonCounts)))
 	} else {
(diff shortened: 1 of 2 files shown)


domain/ide/workspace/folder.go
```diff
@@ -777,9 +791,14 @@ func (f *Folder) filterIssuesWithConfig(
 	}
 
 	if len(filterReasonCounts) > 0 {
-		logger.Debug().Interface("filterReasons", filterReasonCounts).Msgf("%d issue(s) filtered", lo.Sum(lo.Values(filterReasonCounts)))
+		logger.Debug().
+			Interface("filterReasons", filterReasonCounts).
```
basti-snyk: nit: I like `Any` better.

basti-snyk (approved):

PR #864, opened 2025-05-07
refactor: move autofix and autofix feedback logic from language server to code-client-go [IDE-727]

andrewrobinsonhodges-snyk:
### Description

Don't merge until the CCG changes are merged (https://github.com/snyk/code-client-go/pull/100)

At that point, update code-client-go to the new release in go.mod on this branch, and _then_ merge.

### Checklist

- [ ] Tests added and all succeed
- [ ] Linted
- [ ] README.md updated, if user-facing
- [ ] License file updated, if new 3rd-party dependency is introduced

PR #1414, opened 2026-08-20
chore: migrate off legacy AI Explain service [IDE-2487]

rrama:
### Description

Autofix responses now carry the AI explanation inline via `code-client-go`, so the separate legacy AI Explain service call path in snyk-ls is redundant.

This PR:
- Removes the legacy `EnrichWithExplain` flow from `AiFixHandler` (explain-endpoint lookup, timeout, and cancellation logic) and its call site in `CodeFixDiffsCommand` — explanations now arrive as a passthrough field from `code-client-go`
- Bumps `code-client-go` from v1.31.3 to v1.31.7 to bring in that passthrough explanation field
- Removes pre–IDE-727 autofix leftovers with no production callers since autofix moved to `code-client-go` (#864): the local `AutofixUnifiedDiffSuggestion`/`AutofixResponse` types, the `gotextdiff`-based diff conversion in `convert.go`, and their tests. **Unified diffs are still computed** — in `code-client-go/llm/convert.go`, which `GetAutofixDiffs` already delegates to
- `GetDetailsHtml` now warns when an autofix suggestion has no explanation
- Adds a unit test asserting the explanation renders in the issue details HTML panel
- Adds an end-to-end smoke-test check that polls the rendered issue description HTML until autofix reaches a terminal state, then asserts a real, non-empty explanation reaches both `AiFixHandler`'s internal state and the client-visible HTML

### Checklist


infrastructure/code/code_html_test.go
```diff
@@ -144,6 +146,44 @@ func Test_Code_Html_getCodeDetailsHtml_withAIfix(t *testing.T) {
 	assert.Contains(t, codePanelHtml, ` folder-path=""`)
 }
 
+func Test_Code_Html_getCodeDetailsHtml_withAIfix_rendersExplanation(t *testing.T) {
```
basti-snyk, line 149: scope creep, but worth considering: AFAIK, we still display 1-5 and navigation buttons for different fixes, even though only one fix is returned nowadays. It would be good to remove them.

domain/ide/command/code_fix_diffs.go
```diff
@@ -93,6 +95,16 @@ func (cmd *codeFixDiffs) handleResponse(ctx context.Context, engine workflow.Eng
...
+	warnOnMissingExplanations(logger, suggestions)
 	aiFixHandler.SetAiFixDiffState(code.AiFixSuccess, suggestions, nil, setStateCallback)
 }
+
+// warnOnMissingExplanations logs a warning for any suggestion missing its passthrough
+// explanation from code-client-go, without treating it as a fix failure.
+func warnOnMissingExplanations(logger zerolog.Logger, suggestions []llm.AutofixUnifiedDiffSuggestion) {
+	for i := range suggestions {
+		if suggestions[i].Explanation == "" {
+			logger.Warn().Msgf("autofix suggestion %s has no explanation", suggestions[i].FixId)
```
basti-snyk: do we need that? I'd say, at most an info level, more likely debug level.

infrastructure/code/convert.go
```diff
@@ -698,47 +695,3 @@ func buildOneLineTextEdit(startLine int, endLine int, text string, lastLineOfOri
 		NewText: text,
 	}, nil
 }
-
-func (s *AutofixResponse) toUnifiedDiffSuggestions(engine workflow.Engine, baseDir types.FilePath, filePath types.FilePath) []AutofixUnifiedDiffSuggestion {
```
basti-snyk, line 702: how comes we don't need that anymore?

infrastructure/code/code_html_test.go
rrama: Good catch — this is already tracked in IDE-2182 (https://snyksec.atlassian.net/browse/IDE-2182), scoped separately since it touches the details panel template/scripts more broadly. Leaving out of this PR to keep it focused on the AI Explain migration.

infrastructure/code/convert.go
rrama: Covered in the PR description: this deletes pre-IDE-727 leftovers with no production callers since autofix moved to code-client-go (#864) — the local `AutofixUnifiedDiffSuggestion` type, `AutofixResponse` JSON shapes, and this `gotextdiff`-based diff conversion (only referenced by unit tests). Unified diffs are still computed, just now in `code-client-go/llm/convert.go`, which `GetAutofixDiffs` already delegates to.

domain/ide/command/code_fix_diffs.go
rrama: Good point — took it further per your suggestion: dropped the whole eager check from here, and moved a debug-level version to `GetDetailsHtml` in `code_html.go` (where the fix result actually gets rendered), so it only logs when we're about to show it. Fixed in 75c4f019: https://github.com/snyk/snyk-ls/commit/75c4f0192a1452c56368fede6821c2d1254550a5

application/server/server_smoke_test.go
```diff
@@ -1044,10 +1045,57 @@ func checkAutofixDiffs(t *testing.T, engine workflow.Engine, issueList []types.S
...
 		break
 	}
 }
 
+// checkAutofixExplanation polls AiFixHandler directly — the same object
+// CodeFixDiffsCommand's handleResponse writes to — rather than round-tripping through
+// rendered HTML. This avoids coupling the assertion to html/template's escaping and to
+// the ambiguous window/showDocument callback, which fires on IN_PROGRESS too, not just
+// on the terminal state.
+func checkAutofixExplanation(t *testing.T, engine workflow.Engine, issueId string) {
```
basti-snyk: A smoke test should check it end-to-end, not unit/integration test style by accessing the objects - it skips the whole notification chain. So what is described in the comment as ambiguous and difficult is exactly what's needed: check the input/output interfaces instead of internals that may or may not be relevant. This should be mostly blackbox, not whitebox.

infrastructure/code/code_html.go
```diff
@@ -173,8 +173,15 @@ func (renderer *HtmlRenderer) GetDetailsHtml(issue types.Issue) string {
 		aiFixErr = aiFixDiffErr.Error()
 	}
 
+	aiFixDiffResult := renderer.AiFixHandler.GetAiFixDiffResult()
+	for i := range aiFixDiffResult {
```
basti-snyk: nit: do we still need a loop? I think only one fix is reported nowadays, right?

basti-snyk (approved):
approved, but please reconsider the smoke test autofix check changes. would very much like them to be black box assertions (on the LSP protocol messages)

domain/ide/command/code_fix_diffs.go
```diff
@@ -93,6 +93,5 @@ func (cmd *codeFixDiffs) handleResponse(ctx context.Context, engine workflow.Eng
 		aiFixHandler.SetAiFixDiffState(code.AiFixError, nil, err, setStateCallback)
 		return
 	}
-	aiFixHandler.EnrichWithExplain(ctx, engine, issue, suggestions)
 	aiFixHandler.SetAiFixDiffState(code.AiFixSuccess, suggestions, nil, setStateCallback)
```
basti-snyk, line 104: Suggestion: Add a one-line comment documenting the contract this now relies on — that code-client-go v1.31.7's `GetAutofixDiffs` is solely responsible for populating `Explanation` (see upstream commit `ef3b7a4`), and that the E2E smoke test (`checkAutofixExplanation`), not a local fallback, is what guards against a future regression here.

infrastructure/code/code_html.go
```diff
@@ -173,8 +173,15 @@ func (renderer *HtmlRenderer) GetDetailsHtml(issue types.Issue) string {
...
+		if aiFixDiffResult[i].Explanation == "" {
+			logger.Debug().Msgf("autofix suggestion %s has no explanation", aiFixDiffResult[i].FixId)
```
basti-snyk: Suggestion: this only logs at DEBUG, which is invisible in production. If code-client-go ever stops populating `Explanation` for a live user (outside of CI, which does catch it via the smoke test), this won't surface anywhere. Consider WARN level or a metric/counter instead.

infrastructure/code/ai_fix_handler_test.go
```diff
@@ -1,149 +0,0 @@
...
-
-	"github.com/snyk/go-application-framework/pkg/configuration/configresolver"
-
-	"github.com/snyk/snyk-ls/application/config"
-	"github.com/snyk/snyk-ls/internal/testutil"
-	"github.com/snyk/snyk-ls/internal/testutil/workspaceutil"
-	"github.com/snyk/snyk-ls/internal/types"
-)
-
-func Test_getExplainEndpoint(t *testing.T) {
```
basti-snyk, line 33: Suggestion: these deleted tests documented org-aware endpoint-selection behavior (`getExplainEndpoint`). That responsibility now lives in code-client-go with only indirect E2E coverage (no direct unit-test replacement). Not blocking — just worth a comment noting the shift, e.g. near `llm.NewDeepcodeLLMBinding` usage.

basti-snyk (commented):
Verification review — AI review

application/server/server_smoke_test.go
rrama: Rewrote checkAutofixExplanation to poll the client-visible rendered issue-description HTML instead of AiFixHandler internals. Kept a secondary check against internal state for the real error message on failure, so it's not purely black-box. Fixed in 627d6dd1: https://github.com/snyk/snyk-ls/commit/627d6dd163f8132f4138a07f3ba97479db2366c5

rrama:
Re: https://github.com/snyk/snyk-ls/pull/1414#pullrequestreview-5016361892

Rewrote checkAutofixExplanation to poll the client-visible rendered issue-description HTML — that's now the pass/fail signal, not AiFixHandler internals. Kept a secondary check against internal state for the real error message on failure (deliberate tradeoff for debuggability, not fully black-box). Fixed in 627d6dd1: https://github.com/snyk/snyk-ls/commit/627d6dd163f8132f4138a07f3ba97479db2366c5

infrastructure/code/code_html.go
rrama: Collapsed to a single index check, with a comment noting only one fix is ever generated. Fixed in 627d6dd1: https://github.com/snyk/snyk-ls/commit/627d6dd163f8132f4138a07f3ba97479db2366c5

domain/ide/command/code_fix_diffs.go
rrama: Added a short comment: 'Explanations come from code-client-go.' Fixed in 627d6dd1: https://github.com/snyk/snyk-ls/commit/627d6dd163f8132f4138a07f3ba97479db2366c5

infrastructure/code/code_html.go
rrama: Changed to Warn. This only fires on a user-triggered autofix path, so it won't be noisy. Fixed in 627d6dd1: https://github.com/snyk/snyk-ls/commit/627d6dd163f8132f4138a07f3ba97479db2366c5

infrastructure/code/ai_fix_handler_test.go
rrama: No comment added. The rewritten smoke test (checkAutofixExplanation, in 627d6dd1: https://github.com/snyk/snyk-ls/commit/627d6dd163f8132f4138a07f3ba97479db2366c5) exercises this end-to-end via the client-visible HTML now, which covers it more directly than a comment would.

PR #1416, opened 2026-08-26
fix: materialize base branch worktree directly [IDE-2473]

basti-snyk:
### Description

Delta scanning reported no new issues for any repository containing a file whose base name is a Windows reserved device (`CON`, `PRN`, `AUX`, `NUL`, `COM1-9`, `LPT1-9`). go-git v5.19.x rejects those names in `worktreeFilesystem.validPath` on **every** operating system, where real git gates the equivalent `is_valid_win32_path` to Windows-native builds. So the base branch clone failed on macOS and Linux with `lstat: invalid path`, no baseline was recorded, and "New Issues" silently showed nothing.

Bumping go-git does not fix it. The reserved-name check on the checkout path is unchanged through v5.19.2, upstream is unfixed on the v5 line, and a fix there would land on `main` (v6) and need a separate backport plus a release. This is deliberately a local fix.

**Approach.** Clone with `NoCheckout` and write the commit tree ourselves, so go-git's worktree filesystem is never involved. On Windows the writes go through a `\\?\` extended-length prefix, which skips Win32 path parsing and lets NTFS hold the file. That covers the case a Windows developer can actually reach, where the reserved name is on the base branch and not in their working tree, since git refuses to check such a file out on Windows at all.

Setting `core.protectNTFS=false` on the clone would be ~5 lines and would fix macOS, but not Windows, and it disables a CVE mitigation wholesale.

### Owning the guards we left behind

Bypassing that layer means taking over the path safety it was applying. Every validator in go-git's `worktree_fs.go` was enumerated and accounted for:

| go-git guard | Here |
|---|---|
| control chars, empty path, `.`/`..`, volume names, `.git` + HFS+/NTFS disguises | still go-git's, via `ValidTreePath` |
| reserved device names | dropped deliberately, this is the bug |
| `validNoLeadingSymlink` | replaced: deferred symlink pass + separator rejection |
| `validSymlinkName` (`.gitmodules`) | reimplemented, plain name only |
| `maxTreeDepth` | reimplemented, 1024 |
| `Chroot`, `TempFile`, `readGitmodulesFile` | unreachable, we never chroot or run submodule ops |

Four of those were missing at some point during development and each has a regression test written red first: symlink traversal via a symlink shadowing a directory, the same via an entry name holding a path separator, `.gitmodules` symlinks, and the depth limit. Colliding entries are tolerated rather than fatal, because two tree entries can differ only in case or Unicode form and fold together on APFS and NTFS.

**Please review `internal/vcs/worktree_materializer.go` with that table in mind.** Worth a ProdSec look given the input is a tree object from a repository we do not control.

### Tests

Outside-in, each layer red before the next. The acceptance scenario drives the real server over LSP only (`initialize` with `WorkspaceFolders`, `initialized`, `didChangeConfiguration`, `executeCommand`) and asserts on `publishDiagnostics`, so it checks the customer-visible outcome: the developer sees the finding they introduced and not the one already on the base branch. It fails on a partial baseline, which an "a baseline exists" assertion structurally cannot.

The Snyk API is substituted at startup via a new test-only `di.WithProductScanners`, leaving the real scanner, folder and registration assembly in place.

### Known gaps

- Three symlink tests are `NotOnWindows`, since symlink creation needs elevated privileges there. So the deferred symlink pass, the privilege-error handling and both escape regressions do not run on Windows.
- `TestClone_MaterializesBackslashInFileName` is also `NotOnWindows`, which means the Windows-only warn-and-skip branch for separator-bearing names has no coverage on any platform. Small and worth a follow-up.
- Four pre-existing IDE-2418 BDD steps reach around the server rather than driving it over LSP. Not touched here; the `WithProductScanners` seam this PR adds would let them be rewired.

### Checklist

- [ ] Regenerated mocks, etc. (`make generate`) — not needed, no interfaces changed
- [ ] README.md updated, if user-facing — not needed
- [ ] License file updated, if new 3rd-party dependency is introduced — none added

### DIFF
--- a/application/server/bdd_steps_test.go
+++ b/application/server/bdd_steps_test.go
@@ -1447,7 +1447,7 @@ func createGitRepoForFix(t *testing.T) (string, error) {
 		dir = canonical
 	}
 	run := func(args ...string) error {
-		cmd := exec.Command("git", args...)
+		cmd := exec.Command("git", testsupport.GitUnsigned(args...)...)
 		cmd.Dir = dir
 		if out, err := cmd.CombinedOutput(); err != nil {
 			return fmt.Errorf("git %v: %w (%s)", args, err, out)
--- a/application/server/bdd_steps_test.go
+++ b/application/server/bdd_steps_test.go
@@ -4,17 +4,22 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io/fs"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"runtime"
 	"runtime/debug"
+	"sort"
 	"strings"
 	"testing"
 	"time"
 
 	"github.com/creachadair/jrpc2/server"
 	"github.com/cucumber/godog"
+	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/go-git/go-git/v5/plumbing/filemode"
+	"github.com/go-git/go-git/v5/plumbing/object"
 	sglsp "github.com/sourcegraph/go-lsp"
 	"github.com/spf13/pflag"
 	"github.com/stretchr/testify/assert"
@@ -59,6 +64,7 @@ type bddSteps struct {
 	stepResult   chan error
 
 	engine              workflow.Engine
+	tokenService        types.TokenService
 	loc                 server.Local
 	jsonRPCRecorder     *testsupport.JsonRPCRecorder
 	deps                di.Dependencies
@@ -75,6 +81,9 @@ type bddSteps struct {
 	deltaOssFilePath types.FilePath
 	lastDiagnostics  []types.Diagnostic
 	lastTreeViewHTML string
+	// expectedNewIssueCode is the diagnostic code editorNotifiedOfOnlyNewIssue
+	// requires; scenarios set it to the finding they deliberately introduced.
+	expectedNewIssueCode string
 
 	savedLlmProvider    string
 	savedLlmModel       string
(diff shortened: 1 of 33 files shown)


application/server/bdd_steps_test.go
```diff
@@ -554,6 +575,157 @@ func (s *bddSteps) theDialogContainsNoLlmApiKeyField() error {
 	return nil
 }
 
+// developerOpensRepositoryWithDeviceFileNameOnBaseBranch builds the customer's
+// repository: master carries the reserved device name plus a file with an
+// already-known finding, and the checked-out feature branch drops the reserved
+// name and adds a file with a new finding. Nothing writes a reserved device
+// name to disk, so this runs on every platform.
+func (s *bddSteps) developerOpensRepositoryWithDeviceFileNameOnBaseBranch() error {
```
basti-snyk: this step is very specific and would be better parameterized with filename and branch.

application/server/bdd_steps_test.go
```diff
@@ -554,6 +575,157 @@ func (s *bddSteps) theDialogContainsNoLlmApiKeyField() error {
...
+	// branch and the working tree is exactly the file the feature branch added.
+	bddKnownFindingFile = "Known.java"
+	bddNewFindingFile   = "New.java"
+)
+
+// bddPerFileScanner reports one finding per .java file it finds under the path
+// it is handed. Only the product scanner is substituted: the real
+// DelegatingConcurrentScanner still clones the base branch and diffs against it,
+// which is the code this scenario exists to exercise.
+type bddPerFileScanner struct{}
```
basti-snyk: this is a circumvention of the requirement to do blackbox tests that execute real code - not a fake scanner.

PR #1423, opened 2026-09-02
chore: block accidental private-repo mentions via pre-commit hooks [IDE-2517]

rrama:
### Description

Adds three pre-commit-framework hooks (`pre-commit`/`pre-merge-commit`, `commit-msg`, `pre-push`) that block a commit/push naming a private `snyk/<repo>` by name, unless it's in the `ALLOWED_REPOS` allowlist in `scripts/lib/repo-privacy-check.sh`. Repo visibility is checked live via `gh api repos/<slug> --jq .private`. Commit-time `gh` errors warn only (offline-friendly); push-time `gh` errors block, since push already requires network.

Adds `default_install_hook_types`, which simplifies `make tools`/`make hooks` to a single `pre-commit install` call (previously a separate `--hook-type pre-push` call was needed). This surfaced that the existing `commitlint` hook was never actually wired up (`commit-msg` was never installed) and, on top of that, had no `commitlint.config.js` despite declaring `@commitlint/config-conventional` as a dependency — it would have failed every `commit-msg` run with an `empty-rules` error. Removed it rather than fix a hook that was never functional; happy to re-add with a real config in a follow-up if wanted.

Same change as snyk-eclipse-plugin ([#453](https://github.com/snyk/snyk-eclipse-plugin/pull/453)), which merged first. No automated test suite exists for the new scripts themselves; the underlying logic was manually verified there. `make test` passes at this commit.

### Checklist

---

> [!NOTE]
> **Low Risk**
> Developer workflow and git hook configuration only; no runtime or production code paths change.
> 
> **Overview**
> Adds **local pre-commit hooks** that scan staged diffs, commit messages, and outgoing pushes for `snyk/<repo>` slugs, then **block** (or warn on commit when `gh` is unavailable) if GitHub reports the repo as private. Allowlisted slugs live in `scripts/lib/repo-privacy-check.sh`; everything else is checked with `gh api repos/<slug>`.
> 
> **Pre-commit wiring** gains `default_install_hook_types` for `pre-commit`, `pre-merge-commit`, `commit-msg`, and `pre-push`, so `make tools` / `make hooks` only need `pre-commit install`. The external **commitlint** `commit-msg` hook is **removed** (it was never installed with the old setup and lacked a working config).
> 
> Push-time scanning covers the push ref range (with a fallback for first-push when pre-commit omits `PRE_COMMIT_TO_REF`) and **blocks** on `gh` failures, unlike commit-time checks which warn only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 20d5adb7a24c469be3ef2d39a2af356b1a6349a7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,13 +38,12 @@ TIMEOUT := "-timeout=90m"
 ## tools: Install required tooling.
 .PHONY: tools
 tools: $(TOOLS_BIN)/go-licenses $(TOOLS_BIN)/golangci-lint $(TOOLS_BIN)/pact/bin/pact
-	@command -v pre-commit >/dev/null 2>&1 && { pre-commit install && pre-commit install --hook-type pre-push; } || \
+	@command -v pre-commit >/dev/null 2>&1 && pre-commit install || \
 		echo "⚠️  pre-commit not found — run 'make hooks' after installing pre-commit to enable git hooks"
 
 .PHONY: hooks
 hooks:
 	@pre-commit install
-	@pre-commit install --hook-type pre-push
 
 
 $(TOOLS_BIN)/go-licenses:
--- a/scripts/lib/repo-privacy-check.sh
+++ b/scripts/lib/repo-privacy-check.sh
@@ -1,0 +1,108 @@
+#!/usr/bin/env bash
+# Shared helpers for the internal-repo-mention hooks (pre-commit / commit-msg / pre-push).
+set -euo pipefail
+
+# Repos it's already fine to mention by name, either because they're public,
+# or because they're private but already named publicly.
+ALLOWED_REPOS=(
+  # Private, but already named publicly in cli's cliv2-private/go.mod dependency list.
+  "snyk/ambient-canary"
+  "snyk/cli-extension-axi"
+  "snyk/cli-extension-cos"
+  "snyk/remy-cli-extension"
+  "snyk/rift-cli-extension"
+
+  # Public: snyk/cli itself, plus its extensions/dependencies (from cli's go.mod).
+  "snyk/cli"
+  "snyk/cli-extension-agent-scan"
+  "snyk/cli-extension-ai-bom"
+  "snyk/cli-extension-dep-graph"
+  "snyk/cli-extension-iac"
+  "snyk/cli-extension-iac-rules"
+  "snyk/cli-extension-os-flows"
+  "snyk/cli-extension-sbom"
+  "snyk/cli-extension-secrets"
+  "snyk/code-client-go"
+  "snyk/container-cli"
+  "snyk/dep-graph"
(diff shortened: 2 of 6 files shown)


scripts/lib/repo-privacy-check.sh
```diff
@@ -0,0 +1,108 @@
...
+    if [[ "$private_status" == "true" ]]; then
+      echo "BLOCKED: $slug is a private repo." >&2
+      return 1
+    fi
+    return 0
+  fi
+
+  if grep -qi '404' "$gh_err"; then
+    rm -f "$gh_err"
+    return 0 # repo doesn't exist -> not a real hit
```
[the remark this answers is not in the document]
rrama: Real gap, but not fixable at this layer: GitHub returns 404 both for "doesn't exist" and "exists but this token can't see it" (deliberate, to avoid leaking private-repo existence to unauthorized callers). `check_slug` can't tell those apart from the API response alone — distinguishing them would need an elevated/org-level check beyond what an individual `gh` token has. Accepting as a known limitation rather than a fixable bug here.

Makefile
```diff
@@ -38,13 +38,12 @@ TIMEOUT := "-timeout=90m"
...
 .PHONY: tools
 tools: $(TOOLS_BIN)/go-licenses $(TOOLS_BIN)/golangci-lint $(TOOLS_BIN)/pact/bin/pact
-	@command -v pre-commit >/dev/null 2>&1 && { pre-commit install && pre-commit install --hook-type pre-push; } || \
+	@command -v pre-commit >/dev/null 2>&1 && pre-commit install || \
 		echo "⚠️  pre-commit not found — run 'make hooks' after installing pre-commit to enable git hooks"
 
 .PHONY: hooks
 hooks:
 	@pre-commit install
-	@pre-commit install --hook-type pre-push
```
basti-snyk, line 47: why? I think the --hook-type pre-push is needed for push hooks to be installed.

Makefile
```diff
@@ -38,13 +38,12 @@ TIMEOUT := "-timeout=90m"
 ## tools: Install required tooling.
 .PHONY: tools
 tools: $(TOOLS_BIN)/go-licenses $(TOOLS_BIN)/golangci-lint $(TOOLS_BIN)/pact/bin/pact
-	@command -v pre-commit >/dev/null 2>&1 && { pre-commit install && pre-commit install --hook-type pre-push; } || \
```
basti-snyk, line 41: why? I think the --hook-type pre-push is needed for push hooks to be installed.

basti-snyk (approved):

PR #1441, opened 2026-09-21
fix: resolve paths to the deepest containing workspace folder [IDE-2584]

danskmt:
### Description

`Workspace.GetFolderContaining` iterates `w.folders`, a map, and returns the first folder that contains the path. With nested workspace folders (e.g. `/repo` and `/repo/src`) a file under `/repo/src` matches both, so the answer alternates between parent and child across calls.

`GetFolderContaining` now returns the deepest containing folder. With nested folders the innermost one owns the path — its `.snyk`, its org, its trust decision — which is what every caller already assumes. No change when folders do not nest.

**What the nondeterminism was causing** (all 14 call sites are affected; these are the concrete ones):

- `didSave` on OSS: a save under `/repo/src` was sometimes scanned as a `/repo` folder scan, which cancels a scan already running on `/repo`; the cancelled scan returns `[]Issue{}, nil` and is reported as clean with zero findings.
- `RemoveFolder(folderPath)`: resolves the folder via `GetFolderContaining` and clears *that* folder's issue cache before deleting the map entry. Removing `/repo/src` could wipe `/repo`'s cache.
- `snyk.workspaceFolder.scan` command: given a nested folder's path, could scan the parent instead.
- `updateFolderOrgIfNeeded` / `updateFolderConfig.clearCacheAndRescan`: a folder-config change on a nested folder could refresh or rescan the parent.
- `FolderConfigForSubPath` / `FolderOrganizationForSubPath`: org and `.snyk` policy for a file under a nested folder were randomly the parent's or the child's.
- `didOpen`, `didChange`, `textDocument/diagnostic`, code actions, code lenses, `IssuesForFile` / `IssuesForRange`, code-fix feedback: a save could be scanned by one folder while the hover looked up issues in the other folder's cache and found nothing.

**Behaviour that becomes consistent rather than random** (worth knowing, not a regression): if a nested folder is untrusted while its parent is trusted, files under it were previously scanned roughly half the time via the parent. They are now never scanned through saves and `textDocument/diagnostic` returns empty for them until the nested folder is trusted, which is the correct reading of the trust model.

**Unchanged:** parent folder scans still include nested folders' files, so both folders cache the same file and the last `publishDiagnostics` wins in the IDE. The pre-existing lock caveat on the function is untouched. No interface change, so mocks need no regeneration.

**Testing:** `Test_GetFolderContaining_ReturnsDeepestNestedFolder` registers `repo` and `repo/src` and asserts 50× that a file under `src` resolves to `src` and a file under `repo/lib` resolves to `repo`. Fails on `main`, passes with the fix. `domain/ide/workspace` and `application/server` suites, `go vet`, and golangci-lint pass.

Relevant ticket: [IDE-2584](https://snyksec.atlassian.net/browse/IDE-2584)

### Checklist

- [ ] Regenerated mocks, etc. (`make generate`) — N/A, no interface change
- [ ] README.md updated, if user-facing — N/A
- [ ] License file updated, if new 3rd-party dependency is introduced — N/A

[IDE-2584]: https://snyksec.atlassian.net/browse/IDE-2584?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Central path-resolution logic used across scans, diagnostics, and folder removal; behavior becomes deterministic but changes trust/scan routing for nested untrusted folders relative to trusted parents.
> 
> **Overview**
> **`GetFolderContaining` now returns the deepest (innermost) workspace folder that contains a path**, instead of whichever match appears first while iterating the folders map. For nested roots like `/repo` and `/repo/src`, files under the child consistently bind to the nested folder for scans, org/config, trust, issue lookup, and cache clearing.
> 
> Tests repeat nested lookups **50×** (`NestedFolderLookupRepetitions`) so map-order flakiness cannot pass. Coverage includes deepest-path resolution, **removing a nested folder without clearing the parent’s scan state**, and **`snyk.workspaceFolder.scan` targeting the nested folder only**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f3172345a239fcd276e1223dc3c40dd77b260db0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


andrewrobinsonhodges-snyk (approved):

danskmt:
Moving it to draft as from our latest discussions, this might not be the right direction

PR #1442, opened 2026-09-21
fix: scope Remy fixes to the client's severity filter

nick-y-snyk:
### Description

Remy runs its own scan inside the fix workflow and fixed every finding it found, ignoring the severity filter the client had already sent. A client showing only high and critical findings still received fix proposals for medium ones it never surfaces. Ambient Canary hit this directly: it defaults to `--severity-threshold high`, yet its fix proposals were dominated by two medium findings.

`buildRemyFixConfig` now reads the severity filter already present in the configuration and passes the lowest enabled level to Remy as `severity-threshold`, the key remy-cli-extension registers as `FlagSeverityThreshold`.

No new setting and no new flag. Every client already sends `severity_filter_critical/high/medium/low`, and `processConfigSettings` already writes them into this configuration on both the initialize and the `didChangeConfiguration` paths. The value was sitting there unread. This also means the four IDE plugins get it for free: Remy starts respecting the severity checkboxes the user has already set.

`--severity-threshold` requires `--auto-approve`, which `buildRemyFixConfig` sets, and rejects `--dry-run`, which it does not set. The flag exists in the pinned remy-cli-extension v1.41.1, so no version bump.

#### Known limit

The threshold is inclusive-and-above, so it cannot express a gapped filter. Critical and medium enabled with high disabled resolves to `medium` and fixes high findings the client hides. Widening is the safe direction for a gap; narrowing would skip a severity the client does show. Remy's `--severity-filter` takes an exact set and is the upgrade once the CLI pins v1.42.0 or later. Marked in the code.

With every severity enabled there is nothing to restrict, so the key stays unset rather than carrying a redundant `low`.

### Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes autonomous fix behavior and workflow invocation guards; mis-wiring could skip fixes or broaden scope beyond the UI filter.
> 
> **Overview**
> Remy’s fix workflow now honors the **same severity checkboxes** the IDE already stores in GAF configuration, instead of fixing findings the client hides.
> 
> **`buildRemyFixConfig`** reads `types.GetFilterSeverityFromConfig` and, when any level is enabled, sets remy-cli-extension’s **`severity-filter`** key to a comma-separated **exact set** (e.g. `critical,high` or `critical,medium` with high omitted). **`gafRunner`** skips **`Invoke`** entirely when every severity is off, because an empty filter would mean “no filter” to Remy and fix everything.
> 
> Unit tests cover forwarding defaults and partial filters, gap preservation, and the skip/invoke paths on **`gafRunner`** via a mock engine.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1f5170adaf4fba5265f616975bed482053ecf93c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


domain/snyk/remediation/remy.go
```diff
@@ -146,15 +146,38 @@ func gafRunner(ctx context.Context, eng workflow.Engine, contentRoot string, _ s
...
+		{sf.Low, "low"},
+	} {
+		if l.enabled {
+			levels = append(levels, l.name)
+		}
+	}
+	if len(levels) == 4 {
+		return ""
+	}
+	return strings.Join(levels, ",")
```
danskmt: @nick-y-snyk here if all levels are enabled, we are returning empty string "", and if no levels are enabled, we are also returning empty string "". Is that expected?

domain/snyk/remediation/remy.go
```diff
@@ -146,15 +146,38 @@ func gafRunner(ctx context.Context, eng workflow.Engine, contentRoot string, _ s
...
+}
```
danskmt, line 187: I would suggest a refactoring (considering the same situation of len == 4 or len == 0 to return empty string, but please ensure if that is the expected behavior):
```go
func remySeverityFilter(sf types.SeverityFilter) string {
	levels := make([]string, 0, 4)

	if sf.Critical {
		levels = append(levels, "critical")
	}
	if sf.High {
		levels = append(levels, "high")
	}
	if sf.Medium {
		levels = append(levels, "medium")
	}
	if sf.Low {
		levels = append(levels, "low")
	}

	// All enabled or none enabled: nothing specific to scope
	if len(levels) == 4 || len(levels) == 0 {
		return ""
	}

	return strings.Join(levels, ",")
}
```

domain/snyk/remediation/remy_config_test.go
```diff
@@ -153,3 +154,55 @@ func TestWithLLMProviderEnvLock_RunsFnUnderExclusiveLock(t *testing.T) {
...
+	sf := types.NewSeverityFilter(true, false, false, false)
+	types.SetSeverityFilterOnConfig(base, &sf, &logger)
+
+	conf := buildRemyFixConfig(base, "/work/repo-root")
+
+	assert.Equal(t, "critical", conf.GetString(remySeverityFilterConfigKey))
+}
+
+// TestBuildRemyFixConfig_SeverityFilterKeepsGaps is what the exact set buys over
+// a floor: high disabled between two enabled levels must survive, not round down.
```
andrewrobinsonhodges-snyk: Comment is a bit confusing

domain/snyk/remediation/remy.go
nick-y-snyk: pushing local commit that it's not here :) current state is not okay

domain/snyk/remediation/remy.go
nick-y-snyk: it's not the behaviour we want. with all sev filters disabled, nothing should happen

andrewrobinsonhodges-snyk (approved):

